Issue #28, opened 2024-02-12
Add PostgreSQL and Redis TLS/mTLS support


mariobehling:
The original version of Pretix has implemented support for PostgreSQL in this PR https://github.com/pretix/pretix/pull/3435/files

Please implement the support here as well.

Shivangi-ch:
I have started working on this issue and will make a Pull request soon.

angel-dot-exe:
@mariobehling I have attempted to resolve the issue, please review the PR.

hongquan:
Where is your PR, @fillingtothemomo ?

angel-dot-exe:
@hongquan I made a PR few days ago it's linked to this issue, https://github.com/fossasia/eventyay-tickets/pull/45

hongquan:
Someone who plan to work on this issue, please remember that you are not allowed to copy code from https://github.com/pretix/pretix/ which are younger than this commit [ce834d7](https://github.com/pretix/pretix/commits/ce834d7028d12711b5ddde729908ee7ce83024a9/).

PR #45, opened 2024-03-02
Adds PostgresSQL and Redis support

fillingtothemomo:
Fixes #28 
Please review the PR and suggest changes , I guess the required changes are a bit more complicated than the referenced PR.

PR #46, opened 2024-03-05
Add TLS/mTLS options in database and cache settings for PostgreSQL and Redis

Shivangi-ch:
This PR fixes: #28 

**Current behavior:-**
With the current settings we can use PostgreSQL as database and Redis as cache, we only have to specify it in the configurations.
- For PostgreSQL we can set the `database: backend:` config to `postgresql` 
- For Redis we can give the server location in `redis: location:` config

By giving these configs our application will start running with postgresql and redis as DB and cache respectively.

However, the current `settings.py` does not specify any TLS/mTLS options for DB and cache.

**New Behavior:-**
So to support TLS connections with the database and cache I have added new options only for postgresql database and redis caches.

The new keys added are:-
- for postresql
  - database: sslmode
  - database: sslrootcert
  - database: sslcert
  - database: sslkey
- for redis
  - redis: ssl_cert_reqs
  - redis: ssl_ca_certs
  - redis: ssl_keyfile
  - redis: ssl_certfile
  
These values should be set in the configurations before running the server if we need to run in TLS/mTLS mode.

I have tested the code manually by running postgresql locally on a different port.

### DIFF
--- a/src/pretix/settings.py
+++ b/src/pretix/settings.py
@@ -83,6 +83,30 @@ if 'mysql' in db_backend:
     db_options['charset'] = 'utf8mb4'
 JSON_FIELD_AVAILABLE = db_backend in ('mysql', 'postgresql')
 
+postgresql_sslmode = config.get('database', 'sslmode', fallback='disable')
+USE_DATABASE_TLS = postgresql_sslmode != 'disable'
+USE_DATABASE_MTLS = USE_DATABASE_TLS and config.has_option('database', 'sslcert')
+
+if USE_DATABASE_TLS or USE_DATABASE_MTLS:
+    tls_config = {}
+    if not USE_DATABASE_MTLS:
+        if 'postgresql' in db_backend:
+            tls_config = {
+                'sslmode': config.get('database', 'sslmode'),
+                'sslrootcert': config.get('database', 'sslrootcert'),
+            }
+    else:
+        if 'postgresql' in db_backend:
+            tls_config = {
+                'sslmode': config.get('database', 'sslmode'),
+                'sslrootcert': config.get('database', 'sslrootcert'),
+                'sslcert': config.get('database', 'sslcert'),
+                'sslkey': config.get('database', 'sslkey'),
+            }
+
+    db_options.update(tls_config)
+
+
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.' + db_backend,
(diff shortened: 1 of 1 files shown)


Shivangi-ch:
@mariobehling, @marcoag could you please review the PR

src/pretix/settings.py
```diff
@@ -83,6 +83,30 @@
...
 JSON_FIELD_AVAILABLE = db_backend in ('mysql', 'postgresql')
 
+postgresql_sslmode = config.get('database', 'sslmode', fallback='disable')
+USE_DATABASE_TLS = postgresql_sslmode != 'disable'
+USE_DATABASE_MTLS = USE_DATABASE_TLS and config.has_option('database', 'sslcert')
+
+if USE_DATABASE_TLS or USE_DATABASE_MTLS:
+    tls_config = {}
+    if not USE_DATABASE_MTLS:
+        if 'postgresql' in db_backend:
```
hongquan, line 93: `db_backend` is a string, like `'postgresql'` or `'mysql'`.
The way you are checking `if 'postgresql' in db_backend` is weird.
Please do just normal, common way with `==`.

hongquan:
I think we should not enable TLS if we are connecting PostgreSQL, Redis on the same machine. It means that we should check for `DATABASES['default']['HOST']` before building and applying config for TLS.

If connecting database on the same machine, the connection is already safe. Adding TLS just add overhead without benefit.

What do you think, Mario, Marco?

hongquan:
By the way, I think that the routine of building TLS config can be extracted to some file and imported to _settings.py_, because the _settings.py_ file is already too long, and has many nested blocks.

hongquan:
Why is this PR almost the same as #45 ? Have you copied from the same source?

hongquan:
You copied the code from upstream https://github.com/pretix/pretix/blob/3b98d87a26d02f3dd72a28452421ce42b6bc6620/src/pretix/settings.py#L126-L143, which is illegal.

You may not understand the scope of the work. `eventyay-tickets` is a fork of `pretix`, but at the time `pretix` changed its license. It means that any code you see in `pretix` which is later than the forking time, is under license which is not compatible with `eventyay-tickets`, and `eventyay-tickets` is not allowed to copy.

Shivangi-ch:
> You may not understand the scope of the work. `eventyay-tickets` is a fork of `pretix`, but at the time `pretix` changed its license. It means that any code you see in `pretix` which is later than the forking time, is under license which is not compatible with `eventyay-tickets`, and `eventyay-tickets` is not allowed to copy.

hey @hongquan, I was not aware about that so I just tried to follow the PR mentioned in the issue... 
I will make a new PR for this with your recommended changes

PR #47, opened 2024-03-06
Add TLS/mTLS settings for postgreSQL and Redis

Shivangi-ch:
This PR fixes: #28 

**Current behavior:-**
With the current settings we can use PostgreSQL as database and Redis as cache, we only have to specify it in the configurations.
However, the current `settings.py` does not specify any TLS/mTLS options for DB and cache.

**New Behavior:-**
So to support TLS connections with the database and cache I have added new options only for postgresql database and redis caches.
I have added new keys  such as sslmode, sslrootcert etc.
These values should be set in the configurations before running the server if we need to run in TLS/mTLS mode.


Shivangi-ch:
Hey @hongquan,
I have taken care of  the recommendations  mentioned by you in PR #46
Also I have moved the TLS config  building logic to a separate file `tls_settings.py` and imported it in the main `settings.py`

Could you please review the PR

hongquan (changes requested):
Please rename _tls_settings.py_ to _settings_helpers.py_. Later on, we can move some other code to it.

Please format your code with  [`Ruff`](https://github.com/astral-sh/ruff). Only format your new file. Don't touch other files (this must be done in a dedicated PR somedays in the future).

Shivangi-ch:
> Please rename _tls_settings.py_ to _settings_helpers.py_. Later on, we can move some other code to it.

@hongquan I have renamed the file as suggested.

> Please format your code with [`Ruff`](https://github.com/astral-sh/ruff). Only format your new file. Don't touch other files (this must be done in a dedicated PR somedays in the future).

and formatted the new file using ruff.

src/pretix/settings_helpers.py
```diff
@@ -0,0 +1,38 @@
...
+    db_ssl_mode = config.get("database", "sslmode", fallback="disable")
+    # add postgresql TLS options
+    if db_ssl_mode != "disable" and db_backend == "postgresql":
+        db_tls_config = {
+            "sslmode": db_ssl_mode,
+            "sslrootcert": config.get("database", "sslrootcert"),
+        }
+        # add postgresql mTLS options
+        if config.has_option("database", "sslcert"):
+            db_tls_config.update(
```
hongquan: I thinks in these lines, writing like this is clearer:

```py
db_tls_config['sslcert'] = config.get('database', 'sslcert')
```

hongquan (changes requested):

src/pretix/settings.py
```diff
@@ -3,7 +3,7 @@
 import os
 import sys
 from urllib.parse import urlparse
-
+from .settings_helpers import *
```
hongquan: Please not use `import *`. Be explicit about what you want to import.
[PEP8 coding style](https://peps.python.org/pep-0008/#imports)

hongquan (changes requested):

src/pretix/settings.py
```diff
@@ -209,22 +214,28 @@
...
-    CACHES['redis'] = {
-        "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": config.get('redis', 'location'),
-        "OPTIONS": {
+    OPTIONS = {
             "CLIENT_CLASS": "django_redis.client.DefaultClient",
             "REDIS_CLIENT_KWARGS": {"health_check_interval": 30}
         }
+    redis_tls_config = build_redis_tls_config(config)
+    if(redis_tls_config is not None):
```
hongquan: You didn't format your code with Ruff?

hongquan (changes requested):

src/pretix/settings.py
hongquan: Ah, this file is an existing file, we don't run Ruff on it. Then you have to format your code manually, based on PEP8 coding style.

src/pretix/settings_helpers.py
Shivangi-ch: I have made the changes.

src/pretix/settings.py
Shivangi-ch: I have made these changes.

src/pretix/settings.py
Shivangi-ch: formatted the code.

src/pretix/settings.py
```diff
@@ -209,22 +214,28 @@
 
 HAS_REDIS = config.has_option('redis', 'location')
 if HAS_REDIS:
+    OPTIONS = {
```
hongquan: Please rename it to `redis_options`. We don't expose it as application settings, no reason to make it all-uppercase.

hongquan (changes requested):

src/pretix/settings.py
Shivangi-ch: Renamed to `redis_options`.

hongquan (approved):

mariobehling:
Due to challenges with the previous branch merges, it would be easiest if you could make this PR again against the current development branch. Appreciate your effort!

PR #49, opened 2024-03-12
Removed MariaDB/MySQL Support

### LINKED ISSUE #29: Remove MariaDB support
The goal of eventyay talk, tickets and video is to run on a common stack. Therefore remove MariaDB support after PostgreSQL support is implemented. 

Compare changes in the PR here https://github.com/pretix/pretix/pull/3381/files

Shivangi-ch:
This PR fixes #29

Removed `mysql/mariadb` from `dockerfile` and `requirements`.
Added a condition to exit with code 1 if we try to start with `mysql` in config

PS: Their is one more PR which tries to solve the same issue but it misses some changes and is quite old without any activity, so I created this PR.

### DIFF
--- a/src/pretix/settings.py
+++ b/src/pretix/settings.py
@@ -71,17 +71,12 @@ PRETIX_AUTH_BACKENDS = config.get('pretix', 'auth_backends', fallback='pretix.ba
 db_backend = config.get('database', 'backend', fallback='sqlite3')
 if db_backend == 'postgresql_psycopg2':
     db_backend = 'postgresql'
-DATABASE_IS_GALERA = config.getboolean('database', 'galera', fallback=False)
-if DATABASE_IS_GALERA and 'mysql' in db_backend:
-    db_options = {
-        'init_command': 'SET SESSION wsrep_sync_wait = 1;'
-    }
-else:
-    db_options = {}
+if db_backend == 'mysql':
+    print("MySQL/MariaDB is not supported")
+    sys.exit(1)
 
-if 'mysql' in db_backend:
-    db_options['charset'] = 'utf8mb4'
-JSON_FIELD_AVAILABLE = db_backend in ('mysql', 'postgresql')
+JSON_FIELD_AVAILABLE = db_backend == 'postgresql'
+db_options = {}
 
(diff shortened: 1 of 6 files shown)


Shivangi-ch:
@hongquan can you please review this PR.

src/pretix/settings.py
```diff
@@ -71,17 +71,12 @@
...
-DATABASE_IS_GALERA = config.getboolean('database', 'galera', fallback=False)
-if DATABASE_IS_GALERA and 'mysql' in db_backend:
-    db_options = {
-        'init_command': 'SET SESSION wsrep_sync_wait = 1;'
-    }
-else:
-    db_options = {}
+if db_backend == 'mysql':
+    print("MySQL/MariaDB is not supported")
+    sys.exit(1)
```
hongquan: Instead of terminating application with `sys.exit`, please raise Django's [`ImproperlyConfigured`](https://docs.djangoproject.com/en/5.0/ref/exceptions/#improperlyconfigured).

Also don't call `print`. Pass the message to `ImproperlyConfigured` instead. By raising exeption, we let library user choose what to do, exiting program or recover with alternative settings.

hongquan (changes requested):

src/pretix/settings.py
Shivangi-ch: I have made the suggested changes.

src/tests/travis_mysql.cfg
```diff
@@ -1,5 +1,5 @@
 [database]
-backend=mysql
+backend=sqlite3
```
hongquan: Why the filename is "my_sql.cfg" and the content is "sqlite3"?

We are removing MySQL support, then this file should be deleted.

hongquan (changes requested):

src/tests/travis_mysql.cfg
Shivangi-ch: I have deleted the file. I don't know how i missed it.

mariobehling:
Due to challenges with the previous branch merges, it would be easiest if you could make this PR again against the current development branch. Appreciate your effort!

PR #52, opened 2024-03-17
update django to 3.2

eMBee:


src/pretix/base/i18n.py
```diff
@@ -73,10 +73,11 @@ def __str__(self):
 def get_babel_locale():
     babel_locale = 'en'
     # Babel, and therefore django-phonenumberfield, do not support our custom locales such das de_Informal
-    if localedata.exists(translation.get_language()):
-        babel_locale = translation.get_language()
-    elif localedata.exists(translation.get_language()[:2]):
-        babel_locale = translation.get_language()[:2]
+    if translation.get_language():
```
hongquan, line 76: How about using [walrus operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions) to write less code:

```py
if lang := translation.get_language():
    if localedata.exists(lang):
        babel_locale = lang
    elif localedata.exists(lang[:2]):
        babel_locale = lang[:2]
```

src/pretix/helpers/templatetags/jsonfield.py
```diff
@@ -0,0 +1,61 @@
+# replacement for jsonfallback functions copied from 
+# https://github.com/raphaelm/django-jsonfallback
+
+import copy
+from django.db import NotSupportedError
+from django.db.models import Expression, JSONField
+
+def mysql_compile_json_path(key_transforms):
```
hongquan, line 8: We are [removing MySQL support](https://github.com/fossasia/eventyay-tickets/issues/29). You can skip MySQL.

eMBee: sure, but it's not removed yet. i like to keep things consistent

mariobehling (approved):
Based on the feedback I think we can merge this. If you find any issues, please open a new issue on the tracker. Thanks

src/pretix/helpers/templatetags/jsonfield.py
eMBee: actually i missed that mysql removal was just merged, sorry.

eMBee: fixed in a new PR

PR #54, opened 2024-03-18
remove duplication of requirements, only use setup.py

eMBee:


eMBee:
i am retracting this PR as it looks like we may as well switch to pyproject.toml directly.

PR #59, opened 2024-03-20
convert build setup to pyproject.toml

eMBee:
i have only done a minimal verification that the build command still works.
please test this forwards, backwards and sideways in every way that you can think of.


pyproject.toml
```diff
@@ -0,0 +1,165 @@
...
+requires = [
+  'setuptools',
+  'setuptools-rust',
+  'wheel',
+  'importlib_metadata',
+
+  # These are runtime dependencies that we unfortunately need to be import in the step that generates
+  # all CSS and JS asset files. We should keep their versions in sync with the definition above.
+  'babel',
+  'Django==3.2.*',
```
hongquan: Could I ask, why do you need Django and other libraries to be present at build time?

eMBee: i can't figure out why it would be needed, so i removed it for now

hongquan (approved):

mariobehling:
Merging this as we have this approved. Thanks!

Issue #33, opened 2024-02-12
Bump project to Python 3.11


mariobehling:
Update the project to Python 3.11. Similar changes were done in Pretix here https://github.com/pretix/pretix/pull/2982/files

weshaan:
still on? i can do this

mariobehling:
Go ahead. No need to wait for anyone. Can just make a PR.

hongquan:
There is another PR https://github.com/fossasia/eventyay-tickets/pull/36

But I marked it as WIP because we haven't tested `eventyay-tickets` with Python 3.10+ yet. With many 6 year old dependencies, I doubt they work with Python 3.10.

hongquan:
In the meantime, someone please remove Python 3.6, 3.7 from GitHub Action script. GitHub seems no longer provide runners with these Python versions.

Shivangi-ch:
> In the meantime, someone please remove Python 3.6, 3.7 from GitHub Action script. GitHub seems no longer provide runners with these Python versions.

@hongquan I have fixed this issue in PR #51

PR #61, opened 2024-03-22
Changed to python 3.11 for issue #33

KanzaAkram:


hongquan:
Hi @KanzaAkram , as explained in https://github.com/fossasia/eventyay-tickets/issues/33#issuecomment-1987577424 , we cannot process this PR.

We need to upgrades the dependencies first, to easily test if _eventyay-tickets_ work with Python 3.10+.

PR #63, opened 2024-03-23
Changed to Django 4.2 version

KanzaAkram:


Kyroyen (changes requested):
pytz is deprecated since django version 4.*, and is removed in 5.* . Please consider upgrading from pyjwt to zone.

hongquan:
Please resolve conflict.

PR #100, opened 2024-05-22
Add initial shot at docker compose

norbusan:
based on https://github.com/ZPascal/pretix-docker-compose


mariobehling:
Also see here, please: https://github.com/fossasia/eventyay-docker/blob/main/pretix.cfg

PR #113, opened 2024-05-24
there should not be any need to check out python and dependencies to …

eMBee:
…test the docker build


norbusan (approved):

norbusan:
Why was that there in the beginning, strange ...

marcoag (approved):

PR #114, opened 2024-05-24
Workflow to build specific PRs for autodeployment

marcoag:
Adding a workflow that helps build PRs from any repo and reference declared on the input and pushes them to `eventyay/eventyay-tickets` with the tag `pr-deployment` so we can auto deploy them in the test server that has been recently created and check the result before merging.


norbusan (approved):

marcoag:
I actually thought this through and I will make this deploy directly to development, no point on having two dev servers.

Issue #137, opened 2024-06-10
Deactivate Perform update checks by default


mariobehling:
"Perform update checks" is activated by default. Please deactivate it by default.

https://tickets-dev.eventyay.com/control/global/update/

![Screenshot from 2024-06-10 20-28-04](https://github.com/fossasia/eventyay-tickets/assets/1583873/011390e8-3212-41f9-bd03-c6d6f2463283)

AviGawande:
hey Mario,raised the PR  #162 fixes this issue and closes it

PR #162, opened 2024-06-14
deactivated the perform update checks by default

AviGawande:
This PR solves and close the issue #137 and here is the end result of PR on my local dev server.

### DIFF
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -127,11 +127,11 @@ DEFAULTS = {
         'serializer_class': serializers.BooleanField,
         'form_kwargs': dict(
             label=_("Ask for email addresses per ticket"),
-            help_text=_("Normally, pretix asks for one email address per order and the order confirmation will be sent "
+            help_text=_("Normally, eventyay asks for one email address per order and the order confirmation will be sent "
                         "only to that email address. If you enable this option, the system will additionally ask for "
                         "individual email addresses for every admission ticket. This might be useful if you want to "
                         "obtain individual addresses for every attendee even in case of group orders. However, "
-                        "pretix will send the order confirmation by default only to the one primary email address, not to "
+                        "eventyay will send the order confirmation by default only to the primary email address, not to "
                         "the per-attendee addresses. You can however enable this in the E-mail settings."),
         )
     },
@@ -1787,7 +1787,7 @@ Your {event} team"""))
         ),
     },
     'theme_color_success': {
-        'default': '#50A167',
+        'default': '#50a167',
         'type': str,
         'form_class': forms.CharField,
         'serializer_class': serializers.CharField,
@@ -1808,7 +1808,7 @@ Your {event} team"""))
         ),
     },
     'theme_color_danger': {
-        'default': '#C44F4F',
+        'default': '#c44f4f',
         'type': str,
         'form_class': forms.CharField,
         'serializer_class': serializers.CharField,
@@ -1829,7 +1829,7 @@ Your {event} team"""))
         ),
     },
     'theme_color_background': {
-        'default': '#FFFFFF',
+        'default': '#f5f5f5',
         'type': str,
         'form_class': forms.CharField,
         'serializer_class': serializers.CharField,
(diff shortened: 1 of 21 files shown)


src/pretix/base/settings.py
```diff
@@ -2146,8 +2146,9 @@ def unserialize(cls, s):
         'default': '',
         'type': str
     },
+    # here is the default setting for the updates check 
     'update_check_perform': {
-        'default': 'True',
+        'default': 'False',
```
norbusan, line 2151: That looks like the correct fix, but what is all the rest for?

AviGawande: that's true rest of the code i had previously commited those change then i figured this above code snippet, that's why

norbusan: THen please remove the changes if they are not necessary nor related.

AviGawande: ok will remove and update the PR

AviGawande (commented):
updated the PR branch with the required changes and solves the issue #137

src/pretix/base/settings.py
norbusan: Thanks, a bit better now. Please see further comments down here.

src/pretix/control/forms/global_settings.py
```diff
@@ -70,9 +70,12 @@ def __init__(self, *args, **kwargs):
         self.fields['banner_message_detail'].widget.attrs['rows'] = '3'
 
 
+
+
 class UpdateSettingsForm(SettingsForm):
     update_check_perform = forms.BooleanField(
         required=False,
+        initial=False,  # Set the initial value to False
```
norbusan: Is this necessary when the above is changed?

src/pretix/control/forms/global_settings.py
```diff
@@ -91,3 +94,4 @@ class UpdateSettingsForm(SettingsForm):
     def __init__(self, *args, **kwargs):
         self.obj = GlobalSettingsObject()
         super().__init__(*args, obj=self.obj, **kwargs)
+
```
norbusan: Avoid adding spurious empty lines, since it makes syncing with other patches more difficult.
So please revert this.

src/pretix/control/templates/pretixcontrol/global_update.html
```diff
@@ -25,9 +25,9 @@
             <div class="alert alert-danger">
                 {% trans "The last update check was not successful." %}
                 {% if gs.settings.update_check_result.error == "http_error" %}
-                    {% trans "The pretix.eu server returned an error code." %}
+                    {% trans "The eventyay server returned an error code." %}
                 {% elif gs.settings.update_check_result.error == "unavailable" %}
-                    {% trans "The pretix.eu server could not be reached." %}
+                    {% trans "The eventyay server could not be reached." %}
```
norbusan, line 30: This is a good change, but should be in a different PR.
But ok, we can leave it hear as extra bonus ;-)

src/pretix/control/views/global_settings.py
```diff
@@ -32,7 +32,7 @@ def form_invalid(self, form):
     def get_success_url(self):
         return reverse('control:global.settings')
 
-
+# for UpdateCheckView  implementation
```
norbusan: Revert please, not needed

src/pretix/control/views/global_settings.py
```diff
@@ -60,7 +60,7 @@ def get_context_data(self, **kwargs):
         return ctx
 
     def get_success_url(self):
-        return reverse('control:global.update')
+        return reverse('control:global.update')
```
norbusan: Revert please, purely whitespace change. Makes merging patches more complicated in the future.

AviGawande: ok will revert those changes and update the branch

AviGawande: but amazed that an extra space can also cause conflicts.

norbusan: patch does compare the lines as they are there and the lines as saved in the patch file.
If there is a difference, even wrt whitespace, patch/git has extra options to ignore whitespace, but that is dangerous, since Python is white space depending language

norbusan:
I pushed changes that fixed the whitespace issue.

FYI, the best way to do this, 
```
git checkout development src/pretix/control/views/global_settings.py
git checkout development src/pretix/control/forms/global_settings.py
git commit -m "Fix whitespace issues"
```

norbusan (approved):

norbusan:
Merged the branch, thanks

src/pretix/control/views/global_settings.py
AviGawande: ok,will always keep next time.

Issue #174, opened 2024-06-18
Yes-No question always required


mariobehling:
The Yes-No question is always required causing issues.

AviGawande:
how can i generate this issue on myside so that i can get detail about it

PR #206, opened 2024-06-24
Improve flow for required field of Yes-No question 

AviGawande:
This PR fixes the issue #174


mariobehling:
Please resolve conflicts.

hongquan:
@AviGawande You mentioned about some `required` attribute but none of your code touch it. Is it because the issue has been fixed by some other PR, or you forgot it?

hongquan (approved):
If the `required` attribute has been fixed by other PR, the changeset here is Ok to merge.

AviGawande:
> @AviGawande You mentioned about some `required` attribute but none of your code touch it. Is it because the issue has been fixed by some other PR, or you forgot it?

here in the PR, i mentioned that the "required" attribute not correctly applied,so i just implemented its logic with the "True" value.

AviGawande:
> If the `required` attribute has been fixed by other PR, the changeset here is Ok to merge.

in this raised PR i had fixed it.

hongquan:
> > @AviGawande You mentioned about some `required` attribute but none of your code touch it. Is it because the issue has been fixed by some other PR, or you forgot it?
> 
> here in the PR, i mentioned that the "required" attribute not correctly applied,so i just implemented its logic with the "True" value.

Could you point out which line? None of the changed lines ever touches `required`.

hongquan (changes requested):
Need to point out which line touches `required`.

src/pretix/base/forms/questions.py
```diff
@@ -488,26 +488,18 @@ def __init__(self, *args, **kwargs):
...
-                    # itself.
                     widget = forms.CheckboxInput(attrs={'required': 'required'})
                 else:
                     widget = forms.CheckboxInput()
 
-                if initial:
-                    initialbool = (initial.answer == "True")
-                else:
-                    initialbool = False
+                initialbool = (initial.answer == "True") if initial else False
```
AviGawande, line 502: i am talking about this line which preserves the logic of "required"

hongquan: But that line touches `initial`, not `required`.

AviGawande: yes,but the logic of implementing required is preserved

hongquan: Does "preserved" mean that you don't change anything related to "required" (the logic of "required" will be the same before and after this PR)? 

Then you should update the PR title and description, to not mislead people into thinking that this PR is about to change the "required".

Your phrase 
> 'required' attribute is only set when 'q.required ' is 'True'

is very confusing, because looking into #174 and this PR, I don't see anywhere mentioning "required attribute" or "`q.required`". Please don't bring something out-of-context here.

hongquan: The PR description can just be "Fixes #174".

hongquan (approved):
Can be merged after rephrasing PR description to avoid confusing.

src/pretix/base/forms/questions.py
AviGawande: > 

> The PR description can just be "Fixes #174".

sure i will rephrase the PR description.

Issue #203, opened 2024-06-23
Implement option to add link to privacy policy


mariobehling:
In the organizer settings area on the tab general add an option to add a link to a privacy policy. Ensure to implement this as a multilanguage setting.

![Screenshot from 2024-06-23 15-35-27](https://github.com/fossasia/eventyay-tickets/assets/1583873/469ed993-e686-47dd-86f8-96c8a5ef486e)

PR #213, opened 2024-06-26
Privacy policy implemented in the organizer settings

AviGawande:
This PR solves and closes the issue #203 and it also support to the selected localization i.e. choice of language selected.
Screenshot (53)


mariobehling:
Although a feature may appear similar across different services, it is crucial to adhere to open source license agreements. This means that code cannot be copied from another source with an incompatible license, nor can it be copied without proper attribution to the original author. The code in this pull request appears to be entirely copied from another source, making it non-compliant and therefore unacceptable.

AviGawande:
oh! @mariobehling I can change the frontend by keeping  the privacy policy to general settting implemented as needed.

Issue #204, opened 2024-06-23
Option to add customizable links to footer


mariobehling:
Add options for footer links for event organizers into display tab.

![Screenshot from 2024-06-23 15-50-13](https://github.com/fossasia/eventyay-tickets/assets/1583873/3cb50950-1e62-484c-ae6f-4cd788ea4c86)

![Screenshot from 2024-06-23 15-16-58](https://github.com/fossasia/eventyay-tickets/assets/1583873/abb54136-bd97-4534-8218-26633d7823c8)

PR #219, opened 2024-06-28
Provide footer-links to the tickets-shop

AviGawande:
This PR fixes and adds the feature #204  of adding customizable links to footer.


mariobehling:
There are commits here that should not be there.

AviGawande:
> There are commits here that should not be there.

yes! only the second last one is from my local branch

PR #239, opened 2024-07-14
Fix issue when create an event from other event configuration

odkhang:


mariobehling (approved):

mariobehling:
Still does not solve the problem entirely though.

PR #243, opened 2024-07-17
Use master branch in docker-pr workflow

norbusan:
SSIA


mariobehling:
The idea was to use "main" everywhere rather. Marco already got rid of "master" in eventyay-talk now. So, it seems we changed in two repos around. Maybe "main" is the standard now?

PR #289, opened 2024-08-09
Add translations of menu schedule, session, tickets, join video online

odkhang:
## Summary by Sourcery

Update translation syntax in event index template and add margin-top to alert classes in SCSS.


mariobehling (changes requested):
Please check what breaks the docker build and resolve it.

Issue #191, opened 2024-06-20
Integrate Check-in and Lead Scanning into a Single QR Code


Sak1012:
Currently, the Badges plugin allows the addition of two separate QR codes on badges: one for Check-in and another for Lead-scanning. To simplify the badge design and improve usability, we propose integrating both functionalities into a single QR code.

<img width="1152" alt="image" src="https://github.com/fossasia/eventyay-tickets/assets/75414859/70b7bcd9-8819-464c-ab57-64820d6f686c">

#### Enhancement:
If possible combine the Check-in and Lead-scanning functionalities into a single QR code on badges. This will streamline the process for attendees and organizers, reducing the need for multiple scans and improving the user experience.

Sak1012:
This might be a possible solution for this issue

Both of the QR codes when scanned gives a single string (From the example below)

**Ticket Code :** 5uhbd49wms2fakafn5nzq7pcqpbjun2v
**Lead Scanning Code :** PZD97WTMFJ


![fossasiatick](https://github.com/user-attachments/assets/e7503085-a6ed-402d-9cb7-e1b5e2140999)

Now a possible solution might be to encode both of these strings in a single QR code that may be in the form of json so that when scanned the value is returned as

```json
{
  "ticket": "5uhbd49wms2fakafn5nzq7pcqpbjun2v",
  "lead": "PZD97WTMFJ"
}
```
while scanning the QR  a logic can be implemented to retrieve the necessary string and proceed with further operations.

PR #304, opened 2024-08-21
Unified QR response

Sak1012:
This PR tries to Solve Issue #191 (Integrate Check-in and Lead Scanning into a Single QR Code)

Since both the codes are generated on a order they are now embedded into a single QR in the following json format

```json
{
  "ticket": "5uhbd49wms2fakafn5nzq7pcqpbjun2v",
  "lead": "PZD97WTMFJ"
}
```

## Summary by Sourcery

Unify the QR code generation by embedding both the ticket and lead information into a single JSON object, allowing for integrated check-in and lead scanning functionality.

New Features:
- Integrate check-in and lead scanning into a single QR code by embedding both codes into a unified JSON format.

### DIFF
--- a/src/pretix/base/pdf.py
+++ b/src/pretix/base/pdf.py
@@ -6,6 +6,7 @@ import os
 import subprocess
 import tempfile
 import uuid
+import json
 from collections import OrderedDict
 from functools import partial
 from io import BytesIO
@@ -552,11 +553,11 @@ class Renderer:
 
     def _draw_barcodearea(self, canvas: Canvas, op: OrderPosition, o: dict):
         content = o.get('content', 'secret')
-        if content == 'secret':
-            content = op.secret
-        elif content == 'pseudonymization_id':
-            content = op.pseudonymization_id
-
+        content_dict = {
+                'ticket': op.secret,
+                'lead': op.pseudonymization_id
+        }
+        content = json.dumps(content_dict)
         level = 'H'
         if len(content) > 32:
             level = 'M'
--- a/src/pretix/base/pdf.py
+++ b/src/pretix/base/pdf.py
@@ -554,6 +554,7 @@ class Renderer:
     def _draw_barcodearea(self, canvas: Canvas, op: OrderPosition, o: dict):
         content = o.get('content', 'secret')
         content_dict = {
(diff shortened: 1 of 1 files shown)


cweitat:
I would recommend to have event id or something for scanner to check for validity locally first before calling for server.

Sak1012:
This PR focuses on unifying two QR codes into one. In Open Event Checkin, the QR code will only be scanned after the device is authenticated and the event is selected for check-in. Once the device is authenticated, the list of events becomes available to the scanner.

Could you please clarify how incorporating the event ID for local validity checks before calling the server might work with this process?

cweitat:
@Sak1012 assuming there's no eventid to reference the selected event to scan, every request, even those qr without the right keys, gets sent to the server which are unwanted. If you have 10,000 devices sending json to server for checks, with and without ticket id field, do you think there will be a bottleneck on the server from returning "ticket key not found" / "ticket is not for this event" kind of error?

Sak1012:
I have modified it so that it now includes the eventname. The response would now look like

```json
{
  "event": "EventName",
  "ticket": "5uhbd49wms2fakafn5nzq7pcqpbjun2v",
  "lead": "PZD97WTMFJ"
}
```

cweitat:
> I have modified it so that it now includes the eventname. The response would now look like
> 
> ```json
> {
>   "event": "EventName",
>   "ticket": "5uhbd49wms2fakafn5nzq7pcqpbjun2v",
>   "lead": "PZD97WTMFJ"
> }
> ```

this is based on the assumption that no event name are equal under the same account. will you have checks on the event creation side that no dup names exist?

mariobehling:
I think it would be better to use a unique event-ID.

Sak1012:
Since there is no particular "event_id" for events in eventyay-tickets can `event-slug` and `organiser` be used instead, as they are unique values and by checking both of them it can be made sure it's unique and accurate for the particular event before sending the request to the server.

So the embedded data would look like
```json
{
  "event": "event_slug",
  "org": "organiser",
  "ticket": "5uhbd49wms2fakafn5nzq7pcqpbjun2v",
  "lead": "PZD97WTMFJ"
}
```

PR #331, opened 2024-09-25
Using URL template instead of url hardcode for join video popup

lcduong:
## Summary by Sourcery

Use URL templates instead of hardcoded URLs in the presale templates to enhance maintainability and flexibility.

Enhancements:
- Replace hardcoded URLs with URL templates in the join video popup and event base templates to improve maintainability and flexibility.


src/pretix/presale/templates/pretixpresale/event/base.html
```diff
@@ -31,7 +31,7 @@
             <div class="container">
                 <span class="fa fa-user-secret" aria-hidden="true"></span>
                 {% trans "This shop is currently only visible to you and your team." %}
-                <a href="{{ site_url }}{% url "control:event.live" event=event.slug organizer=event.organizer.slug %}">
+                <a href="{% url "control:event.live" event=event.slug organizer=event.organizer.slug %}">
```
hongquan: Avoid adding too many code into HTML attribute. You even made it worse by using the same quote style (") when nesting.

Please use `{% url 'some-url-name' arg arg2 as the_url %}` syntax to only insert a variable to HTML attribute.

Example: https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#url

hongquan (changes requested):

hongquan (approved):

PR #405, opened 2024-10-16
Enhance Accounts Feature and Enable Ticket Acquisition for Registered Users

odkhang:
Organizer can view their order at /control area

image

New dropdown for organizer account

image

## Summary by Sourcery

Allow organizers to buy tickets by removing customer account management features and simplifying the order processing and email sending logic. Introduce a new dropdown for organizer accounts to enhance navigation.

New Features:
- Introduce a new dropdown for organizer accounts to enhance navigation.

Enhancements:
- Remove customer account management features, including customer login, registration, and related views and forms.
- Simplify email sending logic by removing organizer-specific SMTP settings and customer-related email handling.
- Refactor the order processing logic to remove customer-specific handling and dependencies.
- Update the UI to remove customer account-related elements and improve the organizer's navigation experience.


src/pretix/control/templates/pretixcontrol/auth/login.html
```diff
@@ -39,7 +39,7 @@
                     </a>
                 {% endif %}
                 {% if can_register %}
-                    <a href="{% url "control:auth.register" %}" class="btn btn-link btn-block">
+                    <a href="{% url "control:auth.register" %}?next={{ request.GET.next|urlencode }}" class="btn btn-link btn-block">
```
hongquan: Better split this to a function to check `next` before appending query string.
- If `next` doesn't exist, don't append query string.
- If `next` is empty string, don't append query string.

src/pretix/static/pretixcontrol/js/ui/popover.js
```diff
@@ -0,0 +1,42 @@
+$(function () {
+    var basePath = JSON.parse(document.getElementById('base_path').textContent);
+    var currentPath = window.location.pathname;
+    var queryString = window.location.search;
+    var logoutPath = `/control/logout?back=${encodeURIComponent(currentPath)}%3F${encodeURIComponent(queryString)}`;
```
hongquan: Please re-implement the part of `encodeURIComponent`, `decodeURIComponent` with [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). It is difficult to review this code, it looks easy to produce bugs here.

src/pretix/static/pretixpresale/js/ui/popover.js
```diff
@@ -29,22 +29,31 @@ $(function () {
 
 $(function () {
     var organizerName = JSON.parse(document.getElementById('organizer_name').textContent);
+    var eventSlug = JSON.parse(document.getElementById('event_slug').textContent);
     var basePath = JSON.parse(document.getElementById('base_path').textContent);
+    var show_organizer_area = JSON.parse(document.getElementById('show_organizer_area').textContent);
     var currentPath = window.location.pathname;
     var queryString = window.location.search;
-    var logoutPath = `/${organizerName}/account/logout?next=${encodeURIComponent(currentPath)}%3F${encodeURIComponent(queryString)}`;
+    var logoutPath = `/control/logout?back=${encodeURIComponent(currentPath)}%3F${encodeURIComponent(queryString)}`;
```
hongquan: Same as above comment (replace with `URLSearchParams`).

src/pretix/control/templates/pretixcontrol/base.html
```diff
@@ -59,6 +59,7 @@
             <script type="text/javascript" src="{% static "fileupload/jquery.fileupload.js" %}"></script>
             <script type="text/javascript" src="{% static "lightbox/js/lightbox.min.js" %}"></script>
             <script type="text/javascript" src="{% static "are-you-sure/jquery.are-you-sure.js" %}"></script>
+            <script type="text/javascript" src="{% static "pretixcontrol/js/ui/popover.js" %}"></script>
```
hongquan, line 62: There is a potential issue here: Later on, if you modify *popover.js* and deploy, end user may still get the old content of the code, due to browser cache.
To force browser to load the latest version of the file, you need to make the file URL vary after each deployment. You can use this lib https://pypi.org/project/dj-static-revision/ to do that.

hongquan (changes requested):

src/pretix/control/templates/pretixcontrol/base.html
odkhang: Hi @hongquan, thanks a lot for your suggestion, as I checked, currently eventyay-tickets system using 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage', The ManifestStaticFilesStorage class in Django already handles the issue of browser caching by appending a hash to the filenames of static files.
And our deployment steps included collectstatic, so the caching issue won't happen.

src/pretix/presale/templates/pretixpresale/event/base.html
```diff
@@ -36,18 +36,18 @@
...
-                    </nav>
-                </div>
-            {% endif %}
-            {% include "pretixpresale/fragment_login_status.html" %}
+        {% if event.settings.theme_color_background|upper != "#FFFFFF" or event.settings.logo_image_large %}
+            <div class="pull-right header-part flip hidden-print header-content">
+                <nav class="locales" aria-label="{% trans "select language" %}">
+                {% if event.settings.locales|length > 1 %}
+                    {% for l in languages %}
+                        <a href="{% url "presale:locale.set" %}?locale={{ l.code }}&next={{ request.path }}{% if request.META.QUERY_STRING %}%3F{{ request.META.QUERY_STRING|urlencode }}{% endif %}" class="{% if l.code == request.LANGUAGE_CODE %}active{% endif %}" rel="nofollow" lang="{{ l.code }}" hreflang="{{ l.code }}">
```
hongquan: As already commented in other places, please split out the code which manipulates URL to some filter, using `urllib` to prevent bugs, and make it easier to review.

src/pretix/presale/templates/pretixpresale/event/index.html
```diff
@@ -109,8 +109,9 @@ <h3 class="panel-title">
                             </form>
                         </div>
                         <div class="col-md-4 col-sm-6 col-md-offset-4 col-xs-12">
-                            <a class="btn btn-block btn-primary btn-lg"
-                                    href="{% eventurl request.event "presale:event.checkout.start" cart_namespace=cart_namespace  %}">
+                            <a class="btn btn-block btn-primary btn-lg
+                                {% if not guest_checkout_allowed and not request.user.is_authenticated %}disabled{% endif %}"
+                                href="{% if guest_checkout_allowed or request.user.is_authenticated %}{% eventurl request.event 'presale:event.checkout.start' cart_namespace=cart_namespace %}{% endif %}">
```
hongquan: This is one example that bug is produced when you put to many logic to template: You forgot to handle `{% else %}` and it is pointless to generate `<a>` element with rubbish `href`.

Please, the template is to render view, don't put too much code here.

src/pretix/presale/templates/pretixpresale/fragment_login_status.html
```diff
@@ -4,6 +4,20 @@
...
+    <nav class="login-hdr" aria-label='{% trans "account" %}'>
+        {% if request.user.is_authenticated %}
+            <div class="navigation-button">
+                <a target="_blank" class="header-nav btn btn-outline-success" data-toggle="popover-profile">
+                    <i class="fa fa-user"></i> {{ request.user.fullname|default:request.user.email }}
+                    <i class="fa fa-solid fa-caret-down"></i>
+                </a>
+            </div>
+        {% else %}
+            <a href='{% url "control:auth.login" %}?next={{ request.path|urlencode }}%3F{{ request.META.QUERY_STRING|urlencode }}'>
```
hongquan: Please manipulate URL in Python (filter) code.

src/pretix/static/pretixcontrol/js/ui/popover.js
```diff
@@ -0,0 +1,51 @@
+$(function () {
+    var basePath = JSON.parse(document.getElementById('base_path').textContent);
+    var currentPath = window.location.pathname;
+    var queryString = window.location.search;
+
+    // Using URLSearchParams to handle query parameters
+    var searchParams = new URLSearchParams(queryString);
```
hongquan: It is not used any where. It is confusing that you define some variable for nothing.

src/pretix/static/pretixcontrol/js/ui/popover.js
```diff
@@ -0,0 +1,51 @@
...
+    var backUrl = `${currentPath}${queryString}`;
+
+    // Constructing logout path using URLSearchParams
+    var logoutParams = new URLSearchParams();
+    logoutParams.set('back', backUrl);
```
hongquan: You can quicly create `logoutParams` as:

```js
let logoutParams = new URLSearchParams({ back: backUrl })
```

src/pretix/static/pretixcontrol/js/ui/popover.js
```diff
@@ -0,0 +1,51 @@
...
+    var logoutPath = `/control/logout?${logoutParams.toString()}`;
```
hongquan: The template string ``${}`` will automatically call `toString`, you can make it shorter:

```js
`/control/logout?${logoutParams}`
```

src/pretix/presale/templates/pretixpresale/event/base.html
mariobehling: @odkhang Please also fix this. Thanks.

mariobehling:
Please also fix isort tests.

src/pretix/control/templatetags/build_locale_url.py
```diff
@@ -0,0 +1,21 @@
...
+    Constructs the locale change URL by appending the locale and
+    next path and query string to the URL.
+    """
+    # Build base URL with locale
+    base_url = f"{request.path}?locale={locale_code}"
+
+    # Append query string if it exists
+    if request.META.get('QUERY_STRING'):
+        query_string = urllib.parse.urlencode(urllib.parse.parse_qsl(request.META['QUERY_STRING']))
+        return f"{base_url}%3F{query_string}"
```
hongquan: The point of using `urllib` is to let it automatically encode the illegal characters and construct query string in safe way, why do you duplicate its functionality by writing "?locale=" and "%3F"?

And why do you do:

```py
urlencode(parse_qsl(request.META['QUERY_STRING']))
```

which is the same as accessing `request.META['QUERY_STRING']` ?

hongquan (changes requested):

src/pretix/control/templatetags/build_locale_url.py
odkhang: Hi @hongquan, thanks for your review, I have updated the code,
using parse_qs to parse the existing query string and add its parameters to the new params then construct the full URL-encoded query string

src/pretix/control/templatetags/build_locale_url.py
```diff
@@ -0,0 +1,20 @@
...
+register = template.Library()
+
+
+@register.filter
+def build_locale_url(request, locale_code):
+    """
+    Constructs the locale change URL by appending the safely encoded locale
+    and next path with the existing query string in a safe and automatic manner.
+    """
+    encoded_locale = urllib.parse.quote(locale_code)
```
hongquan: You don't need to do `quote` here, `urlencode` below will do that.

src/pretix/control/templatetags/build_locale_url.py
```diff
@@ -0,0 +1,20 @@
...
+    params = {'locale': encoded_locale, 'next': request.path}
+    if request.META.get('QUERY_STRING'):
+        existing_params = urllib.parse.parse_qs(request.META['QUERY_STRING'])
```
hongquan: `request.GET` already contains the `dict` you want. But it looks like you want to override the `next` key if doesn't exist. If so, the function can be as simple as:


```py
params = request.GET.copy()  # Do copy because request.GET is immutable
params.setdefault('next', request.path)
params['locale'] = locale_code

return f'{request.path}?{params.urlencode()}'
```

Though I don't understand why you have to set default `next` with the same `request.path`? Will it cause circular redirection?

hongquan (changes requested):

src/pretix/control/templatetags/build_locale_url.py
odkhang: Hi @hongquan,
Thanks for your suggestion.
I keep next = request.path because it's the old implementation from pretix, so I don't want to have any change which not related in this pull request.
Actually from the start this PR not related to locale feature, I just add a login status button next to locale in ticket shop

src/pretix/control/templatetags/append_next_url.py
```diff
@@ -0,0 +1,12 @@
...
+
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag
+def append_next(next_url=None):
+    if next_url and next_url.strip():
+        return f'?next={urlencode({"next": next_url})}'
```
hongquan: Please provide a screenshot to prove that you already test this code.

odkhang: Hi @hongquan, thanks for the review, 
I tested the case when manipulate URL in template 
`<a href="{% url "control:auth.register" %}?next={{ request.GET.next|urlencode }}" class="btn btn-link btn-block">`
But after rework, I missed to test it with template tag, I have removed the 'next=' which will cause the 404 error.

Here is the testing result:

image

hongquan (approved):

Issue #383, opened 2024-10-10
Implement an organizer customer account and fee system


mariobehling:
Currently the system does not have an account system stating the current account debt or surplus, invoices and fee overview in eventyay-common.

Implement a system based on the following FAQ.

---

**How do I pay my fees?**  
At the end of each month, we’ll send you an invoice. You can pay it using a credit card or any other supported Stripe gateway in your region.

**What is the ticket fee?**  
We charge a fee of 2.5% per ticket.

**What currencies can I use?**  
You can use any currency supported by your preferred payment provider. Typically, we convert the fees to USD using a standard conversion rate before invoicing you.

**What are payment provider fees?**  
You can accept payments directly into your bank account or through one of our integrated payment providers, like PayPal or Stripe. These providers may charge their own fees, which can range from a few cents to a few percent, depending on the payment method.

**Is there a minimum contract term?**  
No, there’s no minimum term! If you decide to stop using our service, simply stop selling tickets, and we’ll stop invoicing you. It’s as simple as that.

**Is there a free option?**  
Yes! Our hosted service is free if you only distribute free tickets, up to 500 tickets per year.  
*Note:* Tickets are considered "free" only if attendees don't pay for access in any form. If payments are made online or in cash at the door, we consider it a paid ticket, and standard fees will apply.

**How is the fee calculated?**  
We only charge for successfully paid orders. The fee is 2.5% of each sold item’s price (including non-ticket items), rounded to two decimal places. We then subtract any payments you’ve already made to us and invoice the remaining balance each month.

**How are taxes handled?**  
We calculate the 2.5% fee based on the net price of your tickets. If your ticket prices include VAT, we’ll deduct the VAT before calculating our fee.

**What happens if I cancel my event or some orders?**  
If you cancel orders, we cannot refund the ticket fees you’ve already paid, but we’ll credit the amount to your account. This credit will be applied to your next invoice, reducing the cost of your next event.

odkhang:
Hi @mariobehling,
Please help me with following questions:
1. About how Event can be public, it will be handled by this issue https://github.com/fossasia/eventyay-tickets/issues/379
2. About FAQ, we will implement a FAQ text create/update in eventyay-common which allow Admin edit FAQ there. And where should we show it to User? in the Header of register page, along with /privacy and /terms?
3. Fee system:
[...]

PR #412, opened 2024-10-25
Implement feature to allow Admin to create/update FAQ content

lcduong:
This issue part of #383 
- In common component, add another setting page (only visible for admin)

image

- Admin can create/update FAQ title, content

image


mariobehling:
We do not need an implementation of an FAQ at the moment. We need an implementation of the actual logic of the business processes. Therefore no need to work on this PR at the moment.

mariobehling:
Duplicate with https://github.com/fossasia/eventyay-tickets/issues/379

PR #417, opened 2024-10-29
Implement schedule task to collect billing invoice

lcduong:
This PR is part of issue #383 
Implement:
1. Use celery beat for schedule task
2. Implement task billing collect to collect all billing for all events on monthly basic.
Tasks implemented
- Billing invoice collect: Collect billing on a monthly basis for all events which have order in last month. Running at: 0:00 AM 1st of every month
- Billing invoice notification: Get all billing invoices which created in last month. Trigger an email to Organizer' Primary Contact Email (in billing settings). attach invoice pdf. Running at: 0:10 AM 1st of every month.
- Billing charge on invoice: Get all pending invoice created last month and its status is PENDING, trigger stripe to charge for that event. Running at 00:20 1st of every month.
- Billing invoice payment retry: Get all pending invoice created last month, check if today is schedule date for payment retry, if yes, then trigger Stripe to charge pending invoice again. Run every day at 00:30 AM
- Billing invoice warning: Get all billing invoices which status is PENDING and 'reminder_enabled' flag is True, Check if today is the date which configured in 'reminder_schedule' OR the warning date is before but no warning send yet, then send warning to organizer. Run every day at 00:40 AM

For testing, I have created 5 endpoints to trigger the schedule task:

- Billing invoice collect: /api/v1/billing-testing/invoice-collect
- Billing invoice notification: /api/v1/billing-testing/invoice-notification
- Billing charge on invoice: /api/v1/billing-testing/invoice-charge
- Billing invoice payment retry: /api/v1/billing-testing/invoice-retry
- Billing invoice warning: /api/v1/billing-testing/invoice-warning

## Summary by Sourcery

Implement a scheduled task to collect billing invoices for events monthly using Celery Beat, and introduce a new 'BillingInvoice' model to manage billing data. Update the deployment configuration to support the new task scheduling.

New Features:
- Introduce a scheduled task using Celery Beat to collect billing invoices for all events on a monthly basis.

Enhancements:
- Add a new model 'BillingInvoice' to manage billing data, including fields for status, amount, currency, and reminder schedules.

Build:
- Update the Celery Beat schedule to include a new task for monthly billing collection.

Deployment:
- Add a new supervisor configuration for running the Celery Beat process in the deployment setup.


src/pretix/base/migrations/0004_billinginvoice_stripe_payment_intent_id_and_more.py
hongquan, line 1: Let introduce only 1 DB migration file, to not inflate the number.

src/pretix/settings.py
```diff
@@ -735,6 +736,31 @@
     ('pretix.plugins.banktransfer.*', {'queue': 'background'}),
 ],)
 
+CELERY_BEAT_SCHEDULE = {
```
hongquan, line 739: Tasks which run periodically, better implemented with [systemd timer](https://wiki.archlinux.org/title/Systemd/Timers). Benefit:

- Beside the time point which you scheduled, you can run the task any time, (when you need to test, or there is a need to run task one more time) with the command `sudo systemctl start task-name.service`.

- Easily to temporary disable the schedule, for example when you found that there is bug and you want to disable the task until you fix, by `sudo systemctl stop task-name.timer`.

- Logs are collected by `journald` (part of systemd), with rich feature to filter and view logs.

- The program only runs when needed. If implemented as Celery Beat, there will be a process always running with the whole code of our application, which is wasteful.

hongquan (changes requested):

src/pretix/base/migrations/0004_billinginvoice_stripe_payment_intent_id_and_more.py
lcduong: I solved it, only 1 migration file now for new table creation.

mariobehling:
As discussed elsewhere: 

> Thanks for your suggestion, so instead of using systemd, can we apply django-celery-beat (https://django-celery-beat.readthedocs.io/en/latest/) which will store periodic tasks into database, and we can implement an API (or setting page) to enable/disable it.

src/pretix/eventyay_common/tasks.py
```diff
@@ -180,3 +198,417 @@ def get_header_token(user_id):
...
+                    continue
+        logger.info("End - completed task to collect billing on a monthly basis.")
+    except DatabaseError as e:
+        logger.error("Database error when trying to collect billing: %s", e)
+        # Retry the task if an exception occurs (with exponential backoff by default)
+        try:
+            self.retry(exc=e)
+        except self.MaxRetriesExceededError:
+            logger.error("Max retries exceeded for billing collect.")
+    except Exception as e:
```
hongquan, line 287: Don't catch general `Exception` class. Catch concrete ones.

PR #508, opened 2025-01-15
Create speaker account with SSO as part of the answer to Call for Proposals

HungNgien:
Make social login to redirect to previous page when Login with SSO in other components

## Summary by Sourcery

New Features:
- Added support for Single Sign-On (SSO) for speaker accounts during the Call for Proposals process.


src/pretix/plugins/socialauth/views.py
```diff
@@ -23,38 +25,108 @@
...
-        return process_login_and_set_cookie(request, user, False)
-    except AttributeError:
-        messages.error(
-            request, _('Error while authorizing: no email address available.')
-        )
-        logger.error('Error while authorizing: user has no email address.')
-        return redirect('control:auth.login')
+        )[0]
+
+    def prepare_oauth2_params(self, oauth2_params: dict) -> QueryDict:
```
hongquan: Please simplify the implementation, because you don't need to create a `QueryDict` in order to build some URL query string, a dict is enough.
Here, you just need to define a dict with default values, then merge with the passed-in `oauth2_params`.

Btw, this method doesn't use anything from `self`, it should be moved to a function, to make code less nested.

src/pretix/plugins/socialauth/views.py
```diff
@@ -23,38 +25,108 @@
...
+
+        request.session["oauth2_params"] = sanitized_params
+
+
+class OAuthReturnView(View):
+    def get(self, request: HttpRequest) -> HttpResponse:
+        try:
+            user = self.get_or_create_user(request)
+            response = process_login_and_set_cookie(request, user, False)
+            oauth2_params = request.session.get("oauth2_params", {})
```
hongquan: You should use `.pop()` here, so that you don't have to do `del` below.

hongquan (changes requested):

src/pretix/plugins/socialauth/views.py
HungNguyen-tma: I changed this function to a @staticmethod since it’s only used within this class. 
Would it be better to move it to a utils.py file for better modularity?

src/pretix/plugins/socialauth/schemas/oauth2_params.py
```diff
@@ -0,0 +1,9 @@
+from pydantic import BaseModel, Field
+
+
+class OAuth2Params(BaseModel):
+    response_type: str = Field(default="code")
```
hongquan: To make code clean, please don't use `Field` when you just need to define default value. This is enough:

```py
response_type: str = 'code'
```

src/pretix/plugins/socialauth/views.py
```diff
@@ -18,43 +20,107 @@
...
+            logger.error("Error while validating OAuth2 parameters: %s", e)
+
+
+class OAuthReturnView(View):
+    def get(self, request: HttpRequest) -> HttpResponse:
+        try:
+            user = self.get_or_create_user(request)
+            response = process_login_and_set_cookie(request, user, False)
+            oauth2_params = request.session.pop("oauth2_params", {})
+            if oauth2_params and self.validate_oauth2_params(oauth2_params):
```
hongquan: Now, you already adopt Pydantic model, you should use its validation result:

```py
try:
    oauth2_params = OAuth2Params.model_validate(oauth2_params_raw)
    auth_url = reverse("control:oauth2_provider.authorize")
    query_string = urlencode(oauth2_params.model_dump())
    return redirect(f"{auth_url}?{query_string}")
except ValidationError as e:
    logger.warning("Ignore invalid OAuth2 parameters: %s.", e)
return response
```

Here I change log level to `warning` because this issue of invalid OAuth params is not severe, we can continue to serve user without it.

src/pretix/plugins/socialauth/views.py
```diff
@@ -18,43 +20,107 @@
...
+            return
+
+        parsed = urlparse(next_url)
+
+        # Only allow relative URLs
+        if parsed.netloc or parsed.scheme:
+            return
+
+        params = parse_qs(parsed.query)
+        sanitized_params = {
```
hongquan, line 66: You don't need this step if you have Pydantic model. You can tell Pydantic to automatically strip whitespace.

```py
from typing import Annotated

from pydantic import BaseModel, StringConstraints


class OAuth2Params(BaseModel):
     response_type: Annotated[str, StringConstraints(strip_whitespace=True)] = 'code'
```

Learn more [here](https://docs.pydantic.dev/latest/api/types/#pydantic.types.conbytes).

hongquan (changes requested):

PR #576, opened 2025-03-12
Update pyproject.toml

yaxit24:
production dependencies remain unchanged, dev tools are split into separate docs, test, and lint groups, and Docker/CI scripts are updated to use uv for installing Python libraries.

Moved development libraries out of the single "dev" feature and split them into three groups:
docs: For Sphinx and other libraries needed to build the documentation.
test: For pytest and related testing tools (we'll add mypy and stubs later).
lint: For linting tools like Ruff.
Switched our dependency management to uv for faster and more reliable version resolution.
Updated our Docker build scripts and GitHub Actions workflows to use uv commands (e.g., uv sync) instead of pip, since pip can’t handle uv’s lock file.

## Summary by Sourcery

Refactor development dependencies into separate groups for documentation, testing, and linting. Migrate to uv for dependency management, updating Docker build scripts and CI workflows accordingly.

Enhancements:
- Split development libraries into docs, test, and lint groups.
- Migrate to uv for dependency management for faster and more reliable version resolution.
- Update Docker build scripts and GitHub Actions workflows to use uv commands instead of pip, since pip can’t handle uv’s lock file.
- Remove pydantic dependency


hongquan:
This PR is broken (file is not complete).

PR #579, opened 2025-03-14
Different events can now have the same ticket ID

### LINKED ISSUE #575: Allow tickets with the same ID in different events.

suhailnadaf509:
I've modified the Secret class to only check for duplicate ticket IDs within the same event by changing the query from order__event__organizer=self.event.organizer to order__event=self.event 
Fixes #575

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where ticket IDs could not be reused across different events by modifying the Secret class to only check for duplicate ticket IDs within the same event.

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/base.html
+++ b/src/pretix/control/templates/pretixcontrol/base.html
@@ -107,6 +107,7 @@
                         {% if has_domain and not request.event.live %}
                             <form action="{% eventurl request.event "presale:event.auth" %}" method="post"
                                     target="_blank" class="mobile-navbar-view-form visible-xs-block">
+                                {% csrf_token %}
                                 <input type="hidden" value="{{ new_session }}" name="session">
                                 <button type="submit" class="btn btn-link navbar-toggle">
                                     <i class="fa fa-eye"></i>
@@ -124,7 +125,7 @@
                             <i class="fa fa-eye"></i>
                         </a>
                     {% endif %}
-                    <a class="navbar-brand" href="{% url "control:index" %}">
+                    <a class="navbar-brand" href="{% url 'eventyay_common:dashboard' %}">
                         <img src="{% static "pretixbase/img/eventyay-icon.svg" %}" />
                         {{ django_settings.INSTANCE_NAME }}
                     </a>
--- a/src/pretix/eventyay_common/templates/eventyay_common/base.html
+++ b/src/pretix/eventyay_common/templates/eventyay_common/base.html
@@ -8,7 +8,7 @@
 <html{% if rtl %} dir="rtl" class="rtl"{% endif %}>
 	<head>
 		<title>{% block title %}{% endblock %}{% if url_name != "index" %} :: {% endif %}
-            {{ settings.INSTANCE_NAME_COMMON }}</title>
+            {{ settings.INSTANCE_NAME }}</title>
 		{% compress css %}
     		<link rel="stylesheet" type="text/x-scss" href="{% static "pretixcontrol/scss/main.scss" %}" />
             <link rel="stylesheet" type="text/x-scss" href="{% static "lightbox/css/lightbox.scss" %}" />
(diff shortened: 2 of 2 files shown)


suhailnadaf509:
The fix is necessary because the current implementation incorrectly restricts ticket IDs (secrets) to be unique across all events managed by the same organizer, rather than just within a single event.

hongquan:
Please don't write "Fix #..." in the title. It should be in PR description.

Sak1012 (commented):
## **This will not work as expected**

While the orders will be imported successfully, the original `Ticket ID` (used as the secret) will not be preserved if it already exists in another event.

For example, consider the following CSV row being imported:

```csv
Order#,Order Date,Status,Payment Type,Payment Mode,Ticket ID,Ticket Name,Ticket Price,Ticket Type,Tax ID,Address,Company,Country,State,City,Zipcode,Email,First Name,Last Name,Job Title,Organisation,Partner contact consent,LinkedIn,GitHub,virtual_event_checkin_times
O1742001779-46317,"March 15, 2025 01:22 +0000",completed,,free,f3cdd55c-6bea-4800-a643-5e3dce22ac2b-628a92,Free Virtual Ticket,0,free,,,,Thailand,,,,jason@gmail.com,Jason,White,Full stack developer,Other,Yes,,,[]
```

When imported into the first event (`test1`), the ticket is created with the correct `secret` (matching the provided Ticket ID):

```json
{
    "id": 26326,
    "order": "K0WXU",
    "positionid": 1,
    "item": 23,
    "variation": null,
    "price": "0.00",
    "attendee_name": "Jason White",
    "attendee_name_parts": {
        "_scheme": "given_family",
        "given_name": "Jason",
        "family_name": "White"
    },
    "attendee_email": "jason@gmail.com",
    "secret": "f3cdd55c-6bea-4800-a643-5e3dce22ac2b-628a92", #expected_secret
    "pseudonymization_id": "BLTLY37X3N",
    "seat": null,
    "canceled": false
},
```

However, when importing the same data into a second event (`test2`), since the secret already exists globally, it is replaced with a new one without a error:

```json
{
    "id": 29209,
    "order": "CH0ML",
    "positionid": 1,
    "item": 24,
    "variation": null,
    "price": "0.00",
    "attendee_name": "Jason White",
    "attendee_name_parts": {
        "_scheme": "given_family",
        "given_name": "Jason",
        "family_name": "White"
    },
    "attendee_email": "jason@gmail.com",
    "secret": "krp6hdh7rp2pm6jj9szge37gbufu7ubc", #new_secret
    "pseudonymization_id": "BLTLY37X3N",
    "seat": null,
    "canceled": false
},
```

mariobehling (changes requested):
@suhailnadaf509 Would be great if you follow up on the comments of @Sak1012 . Thanks!

suhailnadaf509:
@Sak1012 Your are right! , well i think we should change the order model to contain a event-scoped constraint 
something like this to the OrderPosition model in `src/pretix/base/models/orders.py`
```
class Meta:
    verbose_name = _("Order position")
    verbose_name_plural = _("Order positions")
    ordering = ("positionid", "id")
    constraints = [
        # Make secrets unique only within the same event
        models.UniqueConstraint(
            fields=['order__event', 'secret'], 
            name='unique_secret_per_event'
        )
    ]
```
and then create a migration file for the above made changes 
```
from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('pretixbase', 'previous_migration'),  # I will Replace these with actual dependencies just a placeholder for now
    ]
        migrations.AddConstraint(
            model_name='orderposition',
            constraint=models.UniqueConstraint(fields=['order__event', 'secret'], name='unique_secret_per_event'),
        ),
    ]
```

then i decided that we should update the the `assign_ticket_secret` function to make it event-aware i was hoping this updation using random generator might do it
```
def assign_ticket_secret(event, position, force_invalidate_if_revokation_list_used=False, force_invalidate=False, save=True):

    from pretix.base.models.orders import generate_position_secret
    gen = event.ticket_secret_generator

    if gen.use_revocation_list and force_invalidate_if_revokation_list_used:
        force_invalidate = True

    original_secret = position.secret

    # For the RandomTicketSecretGenerator, we want to use our custom event-aware function
    if gen.identifier == 'random':
        if force_invalidate or not position.secret:
            # Use the event-aware generation
            position.secret = generate_position_secret(event=event)
            changed = original_secret != position.secret
        else:
            # Keep existing secret
            changed = False
    else:
        # For specialized generators like Sig1, use their own methods
        kwargs = {}
        if 'attendee_name' in inspect.signature(gen.generate_secret).parameters:
            kwargs['attendee_name'] = position.attendee_name

        secret = gen.generate_secret(
            item=position.item,
            variation=position.variation,
            subevent=position.subevent,
            current_secret=position.secret,
            force_invalidate=force_invalidate,
            **kwargs
        )
        changed = position.secret != secret
        if position.secret and changed and gen.use_revocation_list:
            position.revoked_secrets.create(event=event, secret=position.secret)
        position.secret = secret

    if save and changed:
        position.save()
 ```

Sak1012:
> ```
> class Meta:
>     verbose_name = _("Order position")
>     verbose_name_plural = _("Order positions")
>     ordering = ("positionid", "id")
>     constraints = [
>         # Make secrets unique only within the same event
>         models.UniqueConstraint(
>             fields=['order__event', 'secret'], 
>             name='unique_secret_per_event'
>         )
>     ]
> ```


This does not add a global scope. This makes sure that for each `Event`, no two `OrderPosition` objects have the same `secret`, which is already done in a separate function `save()`. 

In the changes for `assign_ticket_secret` some of the imports like

>`from pretix.base.models.orders import generate_position_secret`

will not work as it does not exist.

hpdang:
@suhailnadaf509 can you finish the work you started? Thanks

Issue #558, opened 2025-02-19
Add Nick Name field into account settings


mariobehling:
1. In account settings of the user add a field "Nick name". When the user logs into with Wikimedia add the Wikimedia username into this field automatically.

![Image](https://github.com/user-attachments/assets/a9cc0dc4-71dd-421a-8665-fa817535927a)

2. If the user has the full name field or the username field entered show this information on the top right in the drop down.

![Image](https://github.com/user-attachments/assets/77d1f55c-718c-4971-8bf4-d11cb1db6bf1)

Jaishree2310:
@mariobehling I would like to work on this issue.
can you please assign this to me?

Gagan-Ram:
I am working on this issue

Gagan-Ram:
@mariobehling  While working on this issue, I realized that I need to access https://wikimedia.eventyay.com locally, as the "LOGIN WITH WIKIMEDIA" button only appears on wikimedia.eventyay.com and app-test.eventyay.com.

I have set up Eventyay using the Docker-based setup, but I'm unable to access either https://wikimedia.eventyay.com or https://app-test.eventyay.com locally. Whenever I try, I get an "Internal Server Error" on these pages. (Unfortunately, I couldn’t attach screenshots here—I’m not sure why.)

@hongquan  Could you please join the discussion and help troubleshoot this?

Gagan-Ram:
@mariobehling While working on this issue, I realized that I need to access https://wikimedia.eventyay.com locally, as the "LOGIN WITH WIKIMEDIA" button only appears on wikimedia.eventyay.com and app-test.eventyay.com.

I have set up Eventyay using the Docker-based setup, but I'm unable to access either https://wikimedia.eventyay.com or https://app-test.eventyay.com locally. Whenever I try, I get an "Internal Server Error" on these pages. (Unfortunately, I couldn’t attach screenshots here—I’m not sure why.)

I even tried modifying /etc/hosts file and also nginx files, but nothing really worked! 

@hongquan Could you please join the discussion and help troubleshoot this?

hongquan:
@Gagan-Ram About attaching screenshots, you can upload the screenshot to Imgur and embed it here.

To enable WikiMedia login, please open this page in your local site to see the guide and the link to register Wikimedia app:

![Image](https://github.com/user-attachments/assets/57c4e300-4594-4aab-8476-43987ec66883)

Gagan-Ram:
[...]

PR #582, opened 2025-03-15
Add Wikimedia Nick Name field

Gagan-Ram:
This PR updates the user model to include the wikimedia_username field.
Fixes: #558 

###  Changes Made
 - Added a `wikimedia_username` field to the user model.
 - Updated the account settings UI to include the nickname field and set it to read-only.

### Potential Issues / Considerations
 - Since Wikimedia uses the username as the primary key, a single user can create multiple Wikimedia accounts with the same email ID.
 - If a user signs in to our platform using multiple Wikimedia accounts linked to the same email, their nickname will be updated to the latest Wikimedia account they used.
 - However, they won’t be able to sign in as multiple separate users on our platform using those duplicate Wikimedia accounts.

### Screenshot after the update
Account settings eventyay-common

## Summary by Sourcery

Adds a `wikimedia_username` field to the user model and displays it as a read-only "Nick name" field in the account settings UI. The field is automatically populated when a user logs in via Wikimedia.

Enhancements:
- Populate the wikimedia_username field upon user creation or update if the user logs in via Wikimedia.
- Display the Wikimedia username as a read-only "Nick name" field in the account settings UI.


src/pretix/base/migrations/0009_user_wikimedia_username_and_more.py
```diff
@@ -0,0 +1,313 @@
...
+        ('pretixbase', '0008_cartposition_job_title_orderposition_job_title'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='wikimedia_username',
+            field=models.CharField(max_length=255, null=True),
+        ),
+        migrations.AlterField(
```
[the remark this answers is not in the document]
Gagan-Ram: 1. We don't let user to modify the wikimedia_username so not adding unique constraint should be okay.
2. The migration files are autogenerated when I ran python manage.py makemigrations command

Gagan-Ram:
@hongquan could you please look into this

hongquan:
@Gagan-Ram Please delete all DB migration `AlterField` lines which are not related to your changes of model.

You said that they are generated when you ran `python manage.py makemigrations`. There are two possible causes:
- Your localhost environment was damaged.
- The existing DB migration files are broken.

We need time to check and determine which is the case. In the meantime, to prevent the codebase from being broken if the cause is the 1st case, you need to delete those lines from this PR.

Gagan-Ram:
@hongquan   I noticed that Django is automatically applying BigAutoField because of DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField' in settings.py.
I also identified the models related to the User model:
EmailAddress, LogEntry, NotificationSetting, OAuthAccessToken, OAuthApplication, OAuthGrant, OAuthIDToken, OAuthRefreshToken, RequiredAction, SocialAccount, StaffSession, StaffSessionAuditLog, StaticDevice, TOTPDevice, Team, U2FDevice, WebAuthnDevice.
However, none of these models are affected when the User model is modified.
So I am deleting all migrations which is altering the field 'id' into BigAutoField.

hongquan:
Of course we know that Django applies `BigAutoField` because of the settings. The problem is that it shouldn't generate those `AlterField` lines, because those fields should  "already" be `BigAutoField` before your PR.

src/pretix/base/models/auth.py
```diff
@@ -162,10 +164,13 @@ def get_full_name(self) -> str:
         Returns the first of the following user properties that is found to exist:
 
         * Full name
+        * wikimedia_username
```
hongquan: Please look at the lines above and below and follow the same style.

Gagan-Ram:
It seems like we still need to figure out the problem.
By the way, does the PR look good now? I've deleted the AlterField lines.

hongquan (changes requested):

hongquan:
Hi, I added a comment yesterday but forgot to submit. Please check.

src/pretix/base/models/auth.py
Gagan-Ram: Done, check once

hongquan (approved):

PR #585, opened 2025-03-18
Update test_devices.py

td15:
This Django test suite verifies API functionality related to Device Management for organizers. It includes:

Device Listing (test_device_list) – Ensures the API returns a list of devices associated with an organizer.
Device Detail (test_device_detail) – Retrieves details of a specific device and validates the response.
Device Creation (test_device_create) – Tests device creation with required fields, checks event association, and ensures initialization tokens are generated.
Device Update (test_device_update) – Confirms that updating fields like name is allowed while certain attributes (hardware_brand) remain unchanged.
Device Deletion (test_device_delete) – Ensures the API correctly rejects device deletion requests (405 Method Not Allowed).
The test suite leverages pytest with Django’s ORM, using fixtures to create test data and scopes_disabled() to access database records.

## Summary by Sourcery

Refactor device API tests to use JSON schema validation for responses and improve test coverage. Update device update test to allow changing the hardware brand. Rename device delete test to device soft delete and assert that the device is not deleted but revoked.

Tests:
- Add JSON schema validation to device list and detail API tests.
- Update device update test to allow changing the hardware brand.
- Rename device delete test to device soft delete and assert that the device is not deleted but revoked.
- Add test case for device creation with missing fields.

### DIFF
--- a/src/tests/api/test_devices.py
+++ b/src/tests/api/test_devices.py
@@ -1,9 +1,9 @@
 import pytest
 from django.utils.timezone import now
 from django_scopes import scopes_disabled
+from jsonschema import validate
 
 from pretix.base.models import Device
-
 
 @pytest.fixture
 def device(organizer, event):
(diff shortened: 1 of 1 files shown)


hongquan:
- Validation is needed in application code, but not necessary in test cases, because in application, if the received data is not as expected, we want application to still run, and have different behavior, but in test cases, as long as the data is not valid, we want the test case to fail right away.
- You are bringing jsonschema from JS world, it is far inferior than what Python world has.

PR #588, opened 2025-03-18
Update test_devices.py

td15:
## Summary by Sourcery

Tests:
- Use reverse to construct URLs in device tests.

### DIFF
--- a/src/tests/api/test_devices.py
+++ b/src/tests/api/test_devices.py
@@ -1,6 +1,7 @@
 import pytest
 from django.utils.timezone import now
 from django_scopes import scopes_disabled
+from django.urls import reverse
 
 from pretix.base.models import Device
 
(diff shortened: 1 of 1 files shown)


td15:
Summary of Changes:
Replaced format() with reverse():
Makes API URL construction dynamic and avoids hardcoded paths.
Uses Django’s reverse() function to generate URLs based on named URL patterns.
Updated All API Calls:
test_device_list, test_device_detail, test_device_create, test_device_update, and test_device_delete now use reverse() for API endpoints.

hongquan:
The style of URL `'/api/v1/organizers/{}/devices/'` in test cases is intentional because the author also wants to make sure the final URL appearing to users will be like that.
For example, the author wants to make sure that users see `/api/v1/organizers/`, not `/api/orgranizers/`.

td15:
Thank you for the clarification, @hongquan! I now understand the reasoning
behind keeping the URL format as it is to ensure consistency in what users
see.

I’ll revert that change and review the PR to see if any other improvements
(like JSON schema validation and test case refinements) are still relevant.
If so, I’ll submit them separately.

Also kindly prompt if any other additions or changes are required in any
other file or codepiece may it be unexplored by me or already edited by me
previously, since I have been trying to understand the project and its
design and development process thoroughly to the best of my knowledge.

td15:
Although, would it be useful if I submit a separate PR with JSON schema
validation and test refinements, excluding the URL formatting change?

PR #596, opened 2025-03-26
Fix redundant slash in "/talk//orga" URL

hongquan:
From the dropdown menu on the top nativation bar, when user clicks to the "Talks" link, he/she will be brought to the "/talk//orga/" address. This URL has a redundant slash ("/").
This PR is to fix it.


hongquan:
@sourcery-ai: The JS `URL` object doesn't have `join` functionality and I hesitate to add a library for this small issue.

We may make the frontend app more robust at some point in the future (after getting rid of jQuery).

Issue #595, opened 2025-03-24
Event Creation Allows Past Dates Without Warning


Gagan-Ram:
**Bug Description:**
Currently, users can create an event with a date set in the past without receiving any warnings or errors.

**Expected Behavior:**
The system should prevent event creation with a past date.

**Proposed Solution:**
- On the frontend, display a warning and prevent users from selecting past dates using JavaScript.
- Validate the date on the backend, and return an error if a past date is submitted.

https://github.com/user-attachments/assets/dbe61f78-24c5-47d5-9fc7-ddf87f608206

Gagan-Ram:
@hongquan is this something eventyay is meant to allow, or is it actually a bug? I'm a bit confused.

hongquan:
I think it is a bug. Please go ahead with fixing it.

Gagan-Ram:
Ok, I'll look into it.

mariobehling:
Hi @Gagan-Ram, thanks for taking the time to report this.

We’ve discussed the behaviour internally and discovered several real-world scenarios where organisers **deliberately need to create or edit events in the past**—for example:

* importing legacy events during a migration,
* back-filling ticket data after an on-site sale, or
* publishing slide decks or recordings for a conference that has already taken place.

In these cases, a hard block on past dates would interrupt legitimate workflows.
Displaying an error when someone merely updates an already-finished event could be especially confusing.

Because of that, we see only limited benefit in enforcing a strict date check right now:

* A past date automatically disables ticket sales, so there is no risk of accidental income loss.
* Admins can spot past-dated events easily in the dashboard.
* Implementing and testing all edge cases (cloning, bulk import, multi-day events, timezone shifts, etc.) would consume development bandwidth we currently need for higher-impact issues.

**Next step:** we’ll park this as a *low-priority enhancement*. A possible compromise for the future is a **soft warning** (“You are choosing a past date—continue?”) rather than a hard block, but that requires further UX discussion.

Thanks again for raising the point; we will revisit once the higher-priority items on the roadmap are cleared.

PR #599, opened 2025-03-30
Fix: Prevent past date selection in event creation (#595)

Gagan-Ram:
This PR ensure that the date_from field is correctly restricted to prevent past date selection based on the user's timezone.
Fixes: #595

### Changes Made
- Updated EventWizardBasicsForm and EventUpdateForm to enforce the correct date restriction.
- Ensured backend validation aligns with the frontend restriction.

### How to Test
    1. Go to the event creation page.
    2. Try selecting a past date—it should now be restricted.
    3. Check the behavior for different timezones (e.g., Asia/Kolkata, UTC, etc.)
    4. Ensure the backend also prevents past dates when saving.

#### Question: 
What was the need for converting the date_from value to the user's specific timezone? Why can't we use a simple now()?
#### Answer: 
The conversion is necessary because now() defaults to UTC. If we were to use now() directly, users in different timezones (e.g., Asia/Kolkata, which is UTC+5:30) might see incorrect behavior.

For example, a user in India could still select a past date until 5:30 AM UTC, which is unintended. This happens because, without timezone conversion, the system might consider a new day in UTC while it's still the previous day for the user.

By explicitly converting now() to the user’s timezone before setting min_date, we ensure:
   1. Correct Date Display: The frontend widget (SplitDateTimePickerWidget) correctly restricts date selection based on the user's local time.
   2. Consistent Backend Validation: The backend enforces the same restriction, preventing inconsistencies between the UI and server-side validation.

This approach ensures that users always see the expected minimum date, avoiding cases where they could unintentionally select past dates due to timezone mismatches.

### DIFF
--- a/src/pretix/control/forms/event.py
+++ b/src/pretix/control/forms/event.py
@@ -9,7 +9,7 @@ from django.forms import CheckboxSelectMultiple, formset_factory
 from django.urls import reverse
 from django.utils.html import escape
 from django.utils.safestring import mark_safe
-from django.utils.timezone import get_current_timezone_name
+from django.utils.timezone import get_current_timezone, get_current_timezone_name, now
 from django.utils.translation import gettext, gettext_lazy as _, pgettext_lazy
 from django_countries.fields import LazyTypedChoiceField
 from i18nfield.forms import (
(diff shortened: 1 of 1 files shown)


Sak1012:
image

These Changes work

mariobehling:
I am closing this as it could cause other issues, e.g. when an organiser edits a past event. There are a number of edge cases that would need a longer testing and it is therefore out of scope currently.

PR #611, opened 2025-04-21
Rename Stripe settings keys from stripe_connect_ to stripe_

### LINKED ISSUE #609: Stripe keys do not get stored
1) If the Stripe keys are updated they do not get stored in the system. A user who tries to buy a ticket will receive an error message.

Expired API Key provided: pk_live_*********************************************************************************************FhGvDD. Platform access may have been revoked.

The new keys would be entered in e.g.
https://app-test.eventyay.com/tickets/control/admin/global/settings/

After entereing the new keys here, the system still shows the same error message.

Image

2) The second part of this issue is that the user needs to update all keys in the admin area at https://app-test.eventyay.com/tickets/control/admin/global/settings/. As some previously entered keys show ***** when entering some keys and hitting "Save" there will be  an error message that the keys are wrong. Expected: Just remove the **** hiding option and always show the previously entered keys here as well.

hongquan:
Resolve part of #609

## Summary by Sourcery

Enhancements:
- Update Stripe-related settings key names to use a more concise naming convention

### DIFF
--- a/src/pretix/control/forms/event.py
+++ b/src/pretix/control/forms/event.py
@@ -1528,7 +1528,7 @@ class QuickSetupForm(I18nForm):
         kwargs["locales"] = self.locales
         super().__init__(*args, **kwargs)
         plugins_active = self.obj.get_plugins()
-        if ('eventyay_stripe' not in plugins_active) or (not self.obj.settings.payment_stripe_connect_client_id):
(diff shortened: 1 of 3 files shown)


hongquan:
@sourcery-ai
The DB table for saving these settings is designed with two columns, `key` and `value`. Renaming these settings doesn't make change in DB columns, and we have only one user for prod, we don't need DB migration.
The new values should be updated in Admin.

norbusan (approved):
LGTM

PR #660, opened 2025-05-21
Enable video module by default during event creation

### LINKED ISSUE #647: By default enable video component during event creation
The video module is currently disabled during event creation by default. Change it to be enabled by default.

suhailnadaf509:
Updated `EventCreateView.done()` method to default `is_video_creation` to True
Fixes #647

### DIFF
--- a/src/pretix/eventyay_common/views/event.py
+++ b/src/pretix/eventyay_common/views/event.py
@@ -226,7 +226,7 @@ class EventCreateView(SafeSessionWizardView):
                 "timezone": str(basics_data.get("timezone")),
                 "locale": basics_data.get("locale"),
                 "locales": foundation_data.get("locales"),
-                "is_video_creation": foundation_data.get("is_video_creation"),
+                "is_video_creation": foundation_data.get("is_video_creation", True),
             }
             send_event_webhook.delay(
                 user_id=self.request.user.id, event=event_dict, action="create"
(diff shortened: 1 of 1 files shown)


suhailnadaf509:
> Hey @suhailnadaf509 - I've reviewed your changes - here's some feedback:
> 
> * You’re defaulting `is_video_creation` in multiple places but still sending the raw or default value to the webhook even when you force it to `False` on permission failure—this can cause mismatches; compute the final value once (factoring in permissions) and reuse it consistently.
> * There’s a lot of repeated `foundation_data.get('is_video_creation', True)` logic across branches—consider extracting it into a local variable (or setting the form field’s `initial=True`) to reduce duplication and make the default behavior clearer.
> 
> Here's what I looked at during the review
> Sourcery is free for open source - if you like our reviews please consider sharing them ✨
> Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.

I've updated the logic , so that i don't send raw data to the webhooks  along with factoring in permissions

hongquan (approved):

Gagan-Ram:
@suhailnadaf509 can you share some resources (like screenshots or video) on what this PR does?

mariobehling:
Screenshot from 2025-05-23 23-45-29

After merging the video module is still not enabled by default.

PR #665, opened 2025-05-23
enhancement: Use event creation method in common module only

### LINKED ISSUE #646: Use event creation method in common module only
Currently it is still possible to create events directly within the ticket module, e.g. from the links in the form of 
- [domain]/tickets/control/organizer/[organisername]/
- [domain]/tickets/control/
- [domain]/tickets/control/events/

Instead: Expected is that the button "Create a new event" uses the event creation process of the "common" module as in [domain]/common/.

Image

Saksham-Sirohi:
Changed links on organizers and default ticket control views for creating a new event it uses the common link now to create an event

Fixes #646

## Summary by Sourcery

Use the common module’s event creation URL in dashboard and organizer templates instead of the local control route

Enhancements:
- Replace direct control:events.add links with eventyay_common:events.add in the dashboard view
- Replace direct control:events.add links with eventyay_common:events.add in the organizer detail view

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/dashboard.html
+++ b/src/pretix/control/templates/pretixcontrol/dashboard.html
@@ -16,7 +16,7 @@
     <div class="dashboard">
         {% if can_create_event %}
             <div class="widget-small widget-container">
-                <a href="{% url "control:events.add" %}" class="widget">
+                <a href="{% url "eventyay_common:events.add" %}" class="widget">
                     <div class="newevent"><span class="fa fa-plus-circle"></span>{% trans "Create a new event" %}</div>
                 </a>
             </div>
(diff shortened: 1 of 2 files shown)


norbusan (approved):

mariobehling (approved):

mariobehling:
Ok, further testing shows that this change creates a new bug.
1. Go to https://app-test.eventyay.com/tickets/common/events/add
2. Choose "Create this events for tickets system only"
3. Check the event creation - it is created for both tickets and talk component.

Honestly I am not even sure if we need this feature. Therefore a quick fix would be to just remove the option to enable the separate event modules.

Currently like this:
Screenshot from 2025-05-25 16-31-29

Please remove the option and change it to this:
Screenshot from 2025-05-25 16-31-29a

* Enable module
* Create video platform for this event.
* This will create the video module for live online events.

PR #723, opened 2025-06-05
Make badge page only accessible after plugin is enabled

yaxit24:
Fixes#[687](https://github.com/fossasia/eventyay-tickets/issues/687) I have added the 'BadgePluginEnabledCheck' class in the views.py in the badge in plugins, which checks the plugin in  the enabled plugin list. if the it not enabled then it will return to the 404 page.

## Summary by Sourcery

Enforce badge plugin activation check on all badge views

Enhancements:
- Introduce BadgePluginEnabledCheck mixin to return a 404 when the badges plugin is not enabled
- Apply the BadgePluginEnabledCheck mixin to all existing badge-related views to centralize plugin-enabled validation

### DIFF
--- a/src/pretix/plugins/badges/views.py
+++ b/src/pretix/plugins/badges/views.py
@@ -30,7 +30,7 @@ from pretix.plugins.badges.tasks import badges_create_pdf
 
 from .models import BadgeLayout
 
-class BadgePluginEnabledCheck:
+class BadgePluginEnabledMixin:
 
     def dispatch(self, request, *args, **kwargs):
         if 'pretix.plugins.badges' not in request.event.get_plugins():
@@ -38,7 +38,7 @@ class BadgePluginEnabledCheck:
         return super().dispatch(request, *args, **kwargs)
 
 
-class LayoutListView(BadgePluginEnabledCheck, EventPermissionRequiredMixin, ListView):
+class LayoutListView(BadgePluginEnabledMixin, EventPermissionRequiredMixin, ListView):
     model = BadgeLayout
     permission = ('can_change_event_settings', 'can_view_orders')
     template_name = 'pretixplugins/badges/index.html'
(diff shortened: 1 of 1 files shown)


mariobehling (changes requested):
1. Please check sourcery review.
2. Please follow best practices, e.g. include issue in the PR description "Fixes #[issuenumber]".

A reference to FOSSASIA best practices is here: https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia/

yaxit24:
Thank you for the reminder.
I've reviewed the Sourcery suggestions, incorporated the relevant changes.
I’ll also update the PR description to include the related issue using the "Fixes #[issue_number]" format, and will continue to follow the best practices shared in the FOSSASIA guide.

src/pretix/plugins/badges/views.py
```diff
@@ -30,8 +30,15 @@
 
 from .models import BadgeLayout
 
+class BadgePluginEnabledMixin:
 
-class LayoutListView(EventPermissionRequiredMixin, ListView):
+    def dispatch(self, request, *args, **kwargs):
+        if 'pretix.plugins.badges' not in request.event.get_plugins():
+            raise Http404(_("The requested resource does not exist."))
```
Sak1012: image

Hey @yaxit24 , instead of showing a page not found error(404) i think we can redirect them to the plugins page if the plugin is not enabled, as this might be confusing

Sak1012 (changes requested):

src/pretix/plugins/badges/views.py
yaxit24: Nice idea but is would may be confusing too.

yaxit24: I have updated @Sak1012

Sak1012 (approved):

hongquan (approved):

Issue #657, opened 2025-05-21
Fix naming of email tab and enhance UI logic


mariobehling:
Fix naming of email tab and follow the implementation of eventyay-talk.
1. Rename "Send out emails" to "Mails"
2. Rename "Send email" to "Compose emails"
3. Rename Email history to "Sent emails"
4. Move E-mail content from Settings > E-mail > E-mail content below the "Mails" tab (compare 1.) and name the tab "Templates"

![Image](https://github.com/user-attachments/assets/03465a01-15e7-4763-9383-6920a2336060)

Expected: Make it look more closely to eventyay-talk.
![Image](https://github.com/user-attachments/assets/80c32437-4ac6-497a-a303-a909769f9704)

mariobehling:
Also rename the page heading to match the sidebar menu item, please @Gagan-Ram 

![Image](https://github.com/user-attachments/assets/47a7ca8f-e9f2-4c44-860c-f7c5d380ba23)

PR #743, opened 2025-06-14
Moved email-content to Templates tab under Mails

Gagan-Ram:
fixes: #657

## Summary by Sourcery

Move email content editing out of the main email settings and into a new "Templates" tab under the Sendmail plugin UI, providing a dedicated form and view for configuring all email template texts and placeholders.

New Features:
- Introduce MailContentSettingsForm to manage all event email templates with placeholder support.
- Add MailTemplatesView and URL route to handle the new "Templates" tab for email content.
- Exclude email content fields from the existing email settings form and relocate them to the new Templates interface.
- Update plugin navigation: rename menu items to "Mails", "Compose emails", "Sent emails" and add a "Templates" entry.

Enhancements:
- Enforce selection of the "web" sales channel for download reminder emails via a custom form validator.

### DIFF
--- a/src/pretix/plugins/sendmail/urls.py
+++ b/src/pretix/plugins/sendmail/urls.py
@@ -13,4 +13,9 @@ urlpatterns = [
         views.EmailHistoryView.as_view(),
         name='history',
     ),
+    url(
+        r'^control/event/(?P<organizer>[^/]+)/(?P<event>[^/]+)/sendmail/templates/',
+        views.MailTemplatesView.as_view(),
+        name='templates',
+    ),
 ]
--- a/src/pretix/plugins/sendmail/urls.py
+++ b/src/pretix/plugins/sendmail/urls.py
@@ -1,20 +1,20 @@
-from django.urls import re_path as url
+from django.urls import path
 
 from . import views
 
 urlpatterns = [
-    url(
-        r'^control/event/(?P<organizer>[^/]+)/(?P<event>[^/]+)/sendmail/$',
+    path(
+        'control/event/<str:organizer>/<str:event>/sendmail/',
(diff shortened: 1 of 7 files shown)


mariobehling:
@Gagan-Ram Please add screenshots.

Gagan-Ram:
templates
template_open

norbusan (approved):
Looks reasonable

src/pretix/plugins/sendmail/urls.py
```diff
@@ -13,4 +13,9 @@
         views.EmailHistoryView.as_view(),
         name='history',
     ),
+    url(
```
hongquan: Please use new way of defining URL (`path`). Other use of `url` should be updated some days.

src/pretix/plugins/sendmail/views.py
```diff
@@ -244,3 +248,41 @@ def get_context_data(self, **kwargs):
...
+        )
+
+    def form_invalid(self, form):
+        messages.error(
+            self.request,
+            _('We could not save your changes. See below for details.'),
+        )
+        return super().form_invalid(form)
+
+    @transaction.atomic
```
hongquan, line 274: Any reason to use `transaction.atomic` here?

Gagan-Ram: That's because 'src/pretix/control/views/event.py > MailSettings' was using @transaction.atomic.
May be because of self.request.event.log_action(...)

hongquan: Ok, so it is just part of the old code that you are moving.
I will approve it. But the use of `@transaction.atomic` will be reviewed later (they are redundant if looking into what the function does).

Issue #753, opened 2025-06-18
Migrate eventyay video component to unified system


suhailnadaf509:
Add/Modify Models into ENext from eventyay-video along with the necessary dependencies/files
- [ ] announcement.py
- [ ] audit.py
- [ ] bbb.py
- [ ] cache.py
- [ ] chat.py
- [ ] exhibitor.py
- [ ] feedback.py
- [ ] janus.py
- [ ] poll.py
- [ ] poster.py
- [ ] question.py
- [ ] room.py
- [ ] roulette.py
- [ ] streaming.py
- [ ] turn.py
- [ ] world.py
- [ ] control/models.py

Riteshk1314:
Hey @suhailnadaf509 I am new the org, could you please elaborate the issue. I have been successful in setting up the project locally, If you feel this wouldnt be a "good first issue" then can you please assign any other issue to me. I am comfortable in python and docker.

mariobehling:
Base has been merged, but other areas are still remaining, e.g. here https://github.com/fossasia/eventyay-tickets/pull/915

PR #756, opened 2025-06-19
Eventyay-Video model migration to enext

suhailnadaf509:
I've migrated the models of eventyay-video and unified the similar models/files to avoid conflicts 
solves #753

## Summary by Sourcery

Refactor and unify application configuration, consolidate and expose new core models, introduce secure template filters, include jQuery, and implement a live PollModule with WebSocket commands and events

New Features:
- Unify and enhance Django settings to support environment variables, TLS, Redis, Channels, ASGI, Celery, Sentry, multi‐factor authentication, CORS, and structured directories for data, media, and logs
- Introduce a real‐time PollModule over WebSockets for creating, updating, deleting, voting, pinning, and unpinning polls with channel events
- Bundle jQuery 3.6.0 into the video‐control static assets

Enhancements:
- Consolidate base model imports to expose new models (Poll, Poster, Room, StreamingServer, TurnServer, etc.) in the core package
- Add `escapejson` and `escapejson_attr` template filters and a safe link token generator for secure JSON and attribute rendering
- Update LogEntry to reference `AUTH_USER_MODEL` instead of a hardcoded user model

Documentation:
- Add a default password change form template under video‐control for user registration


Sak1012 (changes requested):
The goal is to migrate the models alone first later we need to migrate the other files, this Migration will not work in many aspects, some of these files should not be in eventyay/base and most of the migrated files still depends on imports from `venueless` that will fail

PR #765, opened 2025-06-26
Enext Update Talk Import

Sak1012:
This PR includes all the upstream changes since Dec 2024 in eventyay-tickets migrated to enext


mariobehling:
There is an issue with this PR (ref. https://github.com/fossasia/eventyay-talk/issues/365) which we need to resolve first before merging this PR.

mariobehling (approved):

Issue #764, opened 2025-06-26
Enhance Event Table by Adding Status Field


Sana0523:
### Description

In the `eventyay/tickets/common/events` page, there is an event table that currently does not display any status field. Enhancing this table by including a **status field** would improve event visibility and management.

This could help:
- Quickly identify the lifecycle stage of events
- Improve UX for organizers
- Enable filtering/sorting by status in future updates

### Suggested Implementation

- Add a `status` column to the event model
- Display the field in the event table view

### Screenshot

![Screenshot](https://github.com/user-attachments/assets/1e29d574-7df1-4321-839b-77ad90162843)

### Labels
`enhancement`

Sana0523:
Hi @mariobehling,
I’ve created a PR for the enhancement discussed in issue #766 , which adds a status field to the event table for better event visibility and management.

Could you please take a moment to review it? Let me know if any changes are needed.

Thank you!

mariobehling:
@Sana0523 I am adding priority low for the moment as we need to focus on other topics first. Will come back to this a bit later. Thank you

chandaAvishek:
I will be fixing this :)

PR #766, opened 2025-06-26
Added status field in events table

Sana0523:
Added a status field in events table in eventyay-common/events
image

Issue-fix: #764

## Summary by Sourcery

Add dynamic status field to events listing

Enhancements:
- Compute event status (upcoming, live, ended) based on current date in the view context
- Display the computed status in a new "Status" column in the events table UI


mariobehling:
@Sana0523 Thanks again. As we are moving forward with the project we need all PRs made against the enext branch. It would be great if you engage and follow up on making your contributions against the enext branch.

Issue #658, opened 2025-05-21
Implement an Outbox for Emails


mariobehling:
Eventyay-talk has an "Outbox" for emails where organisers can check again before an email gets send. Implement an outbox for eventyay-tickets with the same functionality.

Add the outbox menu tab below "Mails" (prerequisite https://github.com/fossasia/eventyay-tickets/issues/657)

Expected: It should look and work like eventyay-talk.

![Image](https://github.com/user-attachments/assets/4b5ee51f-2b74-4057-83aa-db004da088fc)

PR #780, opened 2025-07-04
Implement an email outbox system supporting multiple recipients

Gagan-Ram:
Fixes: #658


Gagan-Ram:
> Please also test this with events with different languages, e.g. an event that has Hindi and English enabled has the option to send registered attendees the language email that they used to register. This behavior should continue to work.

I tested it, it works fine.

mariobehling:
Please address the sourcery reviews.

src/pretix/plugins/sendmail/models.py
```diff
@@ -0,0 +1,400 @@
...
+    raw_message = I18nTextField(null=True, blank=True)
+
+    filters = models.JSONField(null=True, blank=True)
+
+    reply_to = models.CharField(max_length=100, null=True, blank=True)
+    bcc = models.TextField(null=True, blank=True)  # comma-separated
+    locale = models.CharField(max_length=16, null=True, blank=True)
+    attachments = models.JSONField(null=True, blank=True)
+
+    created = models.DateTimeField(auto_now_add=True)
```
hongquan: Please name the fields `created_at`, `updated_at`. Without `_at`, it sounds like Boolean field.

src/pretix/plugins/sendmail/models.py
```diff
@@ -0,0 +1,400 @@
...
+    updated = models.DateTimeField(auto_now=True)
+    sent = models.BooleanField(default=False)
```
hongquan: This field is redundant if you already have `sent_at`, because if `sent_at` has data, it means that the email has been sent. If `sent_at` is null, it means that the email has not been sent.

src/pretix/plugins/sendmail/models.py
```diff
@@ -0,0 +1,400 @@
...
+            return data.get('en', '')
+
+        except (TypeError, ValueError, json.JSONDecodeError):
+            try:
+                # Fallback: Python literal eval for single-quoted dict
+                data = ast.literal_eval(self.raw_subject)
+                if isinstance(data, dict):
+                    return data.get('en', '')
+            except (ValueError, SyntaxError):
+                pass
```
hongquan: The app should not be silent if there are some errors. It should write logs for the developers to know and fix.

src/pretix/plugins/sendmail/models.py
```diff
@@ -0,0 +1,400 @@
...
+
+    raw_subject = I18nTextField(null=True, blank=True)
+    raw_message = I18nTextField(null=True, blank=True)
+
+    filters = models.JSONField(null=True, blank=True)
+
+    reply_to = models.CharField(max_length=100, null=True, blank=True)
+    bcc = models.TextField(null=True, blank=True)  # comma-separated
+    locale = models.CharField(max_length=16, null=True, blank=True)
+    attachments = models.JSONField(null=True, blank=True)
```
hongquan: Do you plan to store file content here? If yes, please don't. Database is not a place to store files.

src/pretix/base/email.py
```diff
@@ -356,8 +356,11 @@ def get_email_context(**kwargs):
         if not isinstance(val, (list, tuple)):
             val = [val]
         for v in val:
-            if all(rp in kwargs for rp in v.required_context):
-                ctx[v.identifier] = v.render(kwargs)
+            try:
+                if all(rp in kwargs for rp in v.required_context):
+                    ctx[v.identifier] = v.render(kwargs)
+            except Exception as e:
```
hongquan: Don't catch the general `Exception`. Use concrete one. There is rare case that you can use it, but need to accompany with `logger.exception`.

src/pretix/base/email.py
```diff
@@ -356,8 +356,11 @@ def get_email_context(**kwargs):
...
+                logger.warning(f"Skipping placeholder {v.identifier} due to error: {e}")
```
hongquan: Don't build string before passing to logger. Use format string to delay building string until it is needed.

```py
logger.warning("Skipping placeholder %s due to error: %s", v.identifier, e)
```

src/pretix/plugins/sendmail/tasks.py
```diff
@@ -110,3 +114,30 @@ def send_mails(
...
+    try:
+        qm = QueuedMail.objects.get(pk=queued_mail_id, event=event)
+        result = qm.send(async_send=True)
+
+        if not result:
+            logger.warning(f"[SendMail] QueuedMail ID {queued_mail_id}: no recipients to send to.")
+        elif not qm.sent:
+            logger.warning(f"[SendMail] QueuedMail ID {queued_mail_id}: partially sent, some recipients failed.")
+        else:
+            logger.info(f"[SendMail] QueuedMail ID {queued_mail_id}: all emails sent successfully.")
```
hongquan: Fix all these logger usage too.

src/pretix/plugins/sendmail/tasks.py
```diff
@@ -110,3 +114,30 @@ def send_mails(
                     )
             except SendMailException:
                 failures.append(o.email)
+
+
+logger = logging.getLogger(__name__)
+
+@app.task(base=ProfiledEventTask, acks_late=True, bind=True, max_retries=3, default_retry_delay=60)
+def send_queued_mail(self, event: Event, queued_mail_id: int):
```
hongquan: You should pass the event_id to the task, to avoid serialization of event object.

src/pretix/plugins/sendmail/models.py
Gagan-Ram: No, this is to store list of file UUIDs

Gagan-Ram:
@hongquan The QueuedMail model is newly created. Any Idea why src/pretix/base/migrations/0010_alter_cachedcombinedticket_id_alter_cachedticket_id_and_more.py got created when I ran:
python manage.py makemigrations sendmail
python manage.py migrate sendmail

hongquan:
@Gagan-Ram The *0010_alter_cachedcombinedticket_id_alter_cachedticket_id_and_more.py* file is created because at some point in the past, when we upgraded Django, we didn't generate the DB migration script. There was a version where Django changed the primary key field from `Integer` to `BigInteger` and we should generate DB migration script for that.

Btw, in your `QueuedMail` model, some fields are better defined as [`ArrayField`](https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/fields/#arrayfield) than `JSONField`. For example, `attachments` is intended to store list of UUID, then it should be defined as:

```py
from django.contrib.postgres.fields import ArrayField

attachements = ArrayField(UUIDField())
```

Because `JSONField` accepts a single scalar value, a dict, a list of string, list of integer, we can save wrong data by accident. Using stricter data type helps prevent that. But it means that we will have to stop supporting SQLite, because SQLite doesn't have "array" type. SQLite also doesn't have native UUID type (Django has to save the string representation of UUID to SQLite and it is not safe, because people can write artbitrary string by accident).

@mariobehling I think that our project has expand to a level that SQLite is no longer suitable.

Gagan-Ram:
> @Gagan-Ram The _0010_alter_cachedcombinedticket_id_alter_cachedticket_id_and_more.py_ file is created because at some point in the past, when we upgraded Django, we didn't generate the DB migration script. There was a version where Django changed the primary key field from `Integer` to `BigInteger` and we should generate DB migration script for that.

Shall I delete the file then? or what you want me to do? if this file is not intended, then won't it affect migrations of other models in future. Also we can see some plugin related migrations are not applied properly when we do 'docker compose up"
See below logs

eventyay-ticket        |   Your models in app(s): 'badges', 'banktransfer', 'eventyay_paypal', 'eventyay_stripe', 'exhibitors', 'pretix_pages', 'pretixapi', 'pretixbase', 'pretixhelpers', 'pretixmultidomain', 'ticketoutputpdf' have changes that are not yet reflected in a migration, and so won't be applied.

hongquan:
Keep it. As I said, it should be created, but we didn't, we owed Django and now we pay the debt.

> 'badges', 'banktransfer', 'eventyay_paypal', 'eventyay_stripe', 'exhibitors', 'pretix_pages', 'pretixapi', 'pretixbase', 'pretixhelpers', 'pretixmultidomain', 'ticketoutputpdf' have changes that are not yet reflected in a migration, and so won't be applied.

These are also the debts.

Gagan-Ram:
I specifically ran python manage.py makemigrations sendmail to avoid applying all pending migrations at once. If I had run python manage.py makemigrations, it would have applied all the unapplied migrations. You can check this by running the command yourself. My concern is whether this new file might cause any issues the next time we run migrations for unapplied ones.

hongquan:
`./manage.py makemigrations` command is to _generate_ (not _apply_) migration scripts.
Your command triggered that file generation because your `QueuedMail` model depends on the models in "pretix/base" (foreign key). Django saw that the one declared in _models.py_ (`BigInteger`) does not match the one in database (`Integer`), so it attempts to generate that script to make them in sync and correct. Nothing wrong with it.

image

hongquan:
@Gagan-Ram What you need to do is to replace those `JSONField` with some stricter field types. Only introduce one DB migration script for your _plugins/sendmail_ app.

hongquan:
What is this field for?

```py
filters = models.JSONField(null=True, blank=True)
```

Gagan-Ram:
> ```python
> filters = models.JSONField(null=True, blank=True)
> ```
I previously used it to store all the filter fields selected by the organizer while composing an email. However, I've now switched to using a separate model, QueuedMailFilter, to store the filters. 

I think I forgot to delete it from QueuedMail model.
Thank you for pointing out. I will remove it.

src/pretix/plugins/sendmail/models.py
```diff
@@ -0,0 +1,400 @@
...
+        except SendMailException as se:
+            recipient.sent = False
+            recipient.error = str(se)
+            recipient.save(update_fields=["sent", "error"])
+            logger.exception("SendMailException error while sending to %s: %s", email, str(e))
+        except Exception as e:
+            recipient.sent = False
+            recipient.error = f"Internal error: {str(e)}"
+            recipient.save(update_fields=["sent", "error"])
+            logger.exception("Unexpected error while sending to %s: %s", email, str(e))
```
hongquan: When using `log.exception`, you don't need to embed error to the logging message.
The error will be run with traceback separately.

src/pretix/plugins/sendmail/models.py
```diff
@@ -0,0 +1,400 @@
...
+    :type sent: bool
+    
+    :param error: Error message if sending failed.
+    :type error: str or None
+    """
+    mail = models.ForeignKey(QueuedMail, on_delete=models.CASCADE, related_name="recipients")
+    email = models.EmailField()
+    orders = ArrayField(models.CharField(max_length=64), blank=True, default=list)
+    positions = ArrayField(models.CharField(max_length=64), blank=True, default=list)
+    items = ArrayField(models.CharField(max_length=64), blank=True, default=list)
```
hongquan, line 317: What is `item`?

src/pretix/plugins/sendmail/models.py
```diff
@@ -0,0 +1,400 @@
...
+            )
+            for email, data in recipients.items()
+        ]
+
+        QueuedMailToUsers.objects.bulk_create(objs)
+        if save:
+            self.save()
+
+
+class QueuedMailToUsers(models.Model):
```
hongquan: Why name the model with plural (and the comment below is "epresents a single recipient"?

hongquan (changes requested):

src/pretix/plugins/sendmail/models.py
Gagan-Ram: Products are represented as items in pretix.
"Items are often also called 'products' but are named 'items' internally due to historic reasons." -> I found this line on class Item(LoggedModel):

hongquan: First, pretix's comment already told that the `item` name is obsolete. They use that name just because in the early days the use case was different and now it is difficult to rename. Your model is newly-created, you don't have to incur debt of the past, why you repeat the same mistake?

Second, what is "product" conceretly, in the use case of this model?

hongquan (changes requested):

Gagan-Ram:
In compose email page, the products are rendered as items in form field - 
>    items = forms.ModelMultipleChoiceField(
        widget=forms.CheckboxSelectMultiple(attrs={'class': 'scrolling-multiple-choice'}),
        label=_('Only send to people who bought'),
        required=True,
        queryset=Item.objects.none(),
    )

src/pretix/base/services/mail.py expects us to pass list of item IDs associated with the user.
To get better understanding of Products take a look at the screenshot
Screenshot from 2025-07-23 10-20-52

hongquan:
Based on the screenshot, please use a clearer name of the `items` field, like `bought_ticket_kinds`.
`items` is a very general name, it will cause question everytime someones see this code.

mariobehling (changes requested):
Please make this PR against the enext branch.

Gagan-Ram:
Closing this PR, the new PR is raised against enext branch: #1190

PR #785, opened 2025-07-10
Fix form initialization crash in global settings admin panel

Gagan-Ram:
When try to access https://app.eventyay.com/tickets/control/admin/global/settings/, it shows runtime error (Internal error)

This PR fixes it.
gs_form_error


Gagan-Ram:
Closing the PR as this seems to be a duplicate of PR #784

Issue #807, opened 2025-07-28
On the tickets page add menu items and move join online event to right


mariobehling:
Add the links to 
* schedule
* sessions
* speakers
* tickets

and move "Join online event" to the right.

<img width="2312" height="102" alt="Image" src="https://github.com/user-attachments/assets/0bfb6f46-5ded-47fe-94e4-1d7245851a50" />

<img width="2426" height="1226" alt="Image" src="https://github.com/user-attachments/assets/96a7e0e5-2b4f-4b6a-86ed-fdf3ea7b8554" />

Gagan-Ram:
@mariobehling, If you'd like the submenu to appear inside the "Join online event" top bar, we’ll need to modify the following template file from the third-party package: https://github.com/pretix/pretix-venueless/blob/master/pretix_venueless/templates/pretix_venueless/order_info.html.
This would also mean discontinuing the use of the pretix_venueless package and instead creating our own in-house package, eventyay_venueless.

If you are okay with adding those menu items below 'Your order: code' or above "Join online event" top bar, then we can implement that directly within the ey-ticket component itself.
I have attached the final output for the same below:

<img width="1542" height="215" alt="Image" src="https://github.com/user-attachments/assets/92b9764e-adaf-4576-9b17-974e46d85188" />

mariobehling:
> If you are okay with adding those menu items below 'Your order: code' or above "Join online event" top bar, then we can implement that directly within the ey-ticket component itself.

@Gagan-Ram It is ok to do this.

PR #814, opened 2025-07-29
Add menu items in tickets page

Gagan-Ram:
Fixes: #807


mariobehling:
Please add a screenshot. Thanks

Gagan-Ram:
> Please add a screenshot. Thanks

If you're able to see the menu items on the Tickets Shop page, then they should also be visible on the Order Tickets page.
Currently, I'm experiencing an issue on my local environment where the menu items are not appearing on the Shop page, even though the event schedule is published and the event is live.

Gagan-Ram:
A new PR is made against enext branch #1110

PR #839, opened 2025-08-19
Fix: Extra DB migration file is created

hongquan:
In additional to fix of "extra DB migration file keeps being created", this PR also does:

- Use `uv` as Python package manager.
- Use Ruff to format code.
- Make PostgreSQL the only supported DB.


norbusan:
Hi @hong_quan Nice idea, but could you please *split* this into two steps:
* first step does the real code changes (switch to uv, ...)
* second one does the ruff-part
With the current PR it is impossible to see the actual code changes to switch to ruff/uv are in the middle of all the ruffification.


Aalso, concerning the postgres only change: Please adjust the README.rst to contain a short info/explanation that when running it locally on hte live system (no docker) than one needs to provide a postgres server and set env variables.

Also, the entry DATABASE could be removed from the env.*.sample, and the code in the entrypoint scripts can be simplified to always run the postgres code since it checls for $DATABASE.

So probably the postgres should be a separate PR, too, to be able to be reviewed.

hongquan:
Close to rework (split to two smaller PRs).

PR #840, opened 2025-08-20
Fix: Extra DB migration file is created

hongquan:
In additional to fix of "extra DB migration file keeps being created", this PR also does:

- Use uv as Python package manager.
- Make PostgreSQL the only supported DB.
- Make shorter app label, like "eventaypresale" -> "presale", to generate DB tables with shorter name.


norbusan:
Thanks for the PR, a few comments:
* why is there `pyproject.toml` and `app/pyproject.toml` - we only want one
* you can leave the PR as is, but as before, I would have preferred you not to merge these three things into one PR. They should have been 3 separate PRs. But let us leave it as it is now, keep it in mind for the next time.
* Can you give an explanation why sqlite3 cannot be supported?

norbusan (changes requested):
The Dockerfiles are not updated to deal with the changes. Without that, this cannot work.

hongquan:
I understand that the "app" folder is a new project. The outer _pyproject.toml_ is for the old, original eventyay-tickets, which will be deleted someday with the old eventyay-tickets code.

For SQLite3, we will re-implement the email feature from eventyay-talk, where there is a need to store list of UUID, and it is better to use ArrayField with UUIDField. These data types are not supported by SQLite.

I will update the Dockerfile, then drop the _requirements.txt_ file.

Issue #837, opened 2025-08-18
Move Views from Components into Base System


mariobehling:
As part of the Eventyay Unified System, we are consolidating functionality from multiple sub-projects (`eventyay-tickets`, `eventyay-talk`, `eventyay-video`, etc.) into the **base system**.  Currently, many **views** still reside in individual components. This causes duplication, inconsistent UX, and makes maintenance harder.  To complete the unification, we need to **migrate all views into the base system**, while keeping components focused on their logic (APIs, services, utilities).

## Goal
- Move views from sub-components into the **base system**.
- Ensure all views follow the **unified design system** (VueJS 3).
- Maintain feature parity while consolidating.
- Track progress for each migrated view here.

### Common views at eventyay.com/[here] [Quan]
- [X] Unified Login (Gagan)
- [x] Registration Views 
- [ ] Event Dashboard (Organizer)
- [ ] Attendee Dashboard
- [ ] Settings (Profile, Organization, Teams)

### Ticketing (from `eventyay-tickets`) at eventyay.com/tickets/ [Srivatsav]
- [ ] Ticket Purchase Page (`/tickets`)
- [ ] Ticket Checkout Page (`/checkout`)
- [ ] Ticket Confirmation Page (`/confirmation`)
- [ ] Ticket QR Scanner View
- [ ] Order History / My Tickets Page

### Talks & Schedule (from `eventyay-talk`) at at eventyay.com/talk/ [Gagan]
- [ ] Schedule Overview (`/schedule`)
- [ ] Session Detail Page
- [ ] Speaker Profile Page
- [ ] Speaker Dashboard (CFP Management)
- [ ] Organizer Schedule Management View
- [ ] Session Submission Form

### Video (from `eventyay-video`) at at eventyay.com/video/ [Suhail]
- [ ] Embedded Video Session View
- [ ] Speaker Backstage / Control View
- [ ] Live Interpretation Selection View
- [ ] Recording Archive / Playback Page

### Badge System [Yakshit] as part of ticketing component
- [ ] Badge Template Designer
- [ ] Badge Preview & Print View
- [ ] Badge Export / Download Page

### Exhibitor Plugin [Yakshit] as part of ticketing component
- [ ] Exhibitor Directory / List View
- [ ] Exhibitor Detail Page

## Migration Steps [Yakshit]
1. Identify and copy view logic & templates from components.
2. Refactor views to use **unified base system components** (VueJS 3).
3. Replace component-level routing with base system routing.
[...]

PR #841, opened 2025-08-20
Add Presale

Sak1012:
Move presale views into Enext, Partially solves #837


Sak1012:
The root page is part of the presale templates.
image


### Note : Multidomain component is restored for URLs config

app/eventyay/config/urls.py
```diff
@@ -38,17 +31,30 @@ def blank_view(request):
...
     path('admin/', include('eventyay.config.urls_admin')),
 ]
 
-urlpatterns = [
-    *admin_patterns,
-    *base_patterns,
-    *control_patterns,
-    *common_patterns,
+debug_patterns = []
+if settings.DEBUG:
```
hongquan: For modern Python, you should write:

```py
import importlib.util

if settings.DEBUG and importlib.util.find_spec('debug_toolbar');
    debug_patterns.append(path('__debug__/', include('debug_toolbar.urls'))
```

Note that, we are using `include` with string to let Django lazily import the `debug_patterns` and help our app start up faster.

Sak1012: I have added the changes 👍🏿

mariobehling:
Ok, lets try this.

PR #848, opened 2025-08-24
Add Organizer views

Sak1012:
Move Organiser Views from eventyay-tickets to enext

Screenshots:
image

image

image


hongquan:
The `Item` model should be renamed, this is the comment from the original code:

https://github.com/fossasia/eventyay-tickets/blob/b6bffdd2b10fdb84e2700fa986f676611714aee0/src/pretix/base/models/items.py#L209-L212

This model was originally named `Item`, after that the pretix authors found that it is not a good name (too generic that the reader cannot understand what it is for), but because it has been referred in many places of the source code, they cannot rename it.

Now we are create a fresh new project, we don't need to retain those mistake. You may create a new PR to work on renaming `Item` model then continue this PR.

mariobehling:
> The `Item` model should be renamed, this is the comment from the original code:

What is a good name? @hongquan

rolandturner:
I'd suggest that trying to determine a better name requires an understanding of what the related objects are called, how they're connected, etc. One way to do this might be for someone to actually map out the current relationships, e.g. in something like an ER model or UML diagram, at a level of detail something like this:

image

(Note that this is **not**  a good example, please don't attempt to emulate it. It's from a "how do I improve my diagram?" question. I meant it only to make clear what kind of thing and level of detail I'm suggesting.)

hongquan:
One of the places that `Item` appear is in the "Limit to products" in this picture:

image_2025-08-04_23-44-32

The `Meta` class of this model also suggest the name that pretix authors want to use:

https://github.com/fossasia/eventyay-tickets/blob/b6bffdd2b10fdb84e2700fa986f676611714aee0/src/pretix/base/models/items.py#L476-L479

To me, `Product` is clearer than `Item` but is still vague, someone with better English, can suggest another name which matches the business of ticketing system.

rolandturner:
All looks reasonable. I do note that there is newly commented-out code in app/eventyay/eventyay_common/base_tasks.py:
- Should this not have been resolved before the pull request?
- If that was not feasible, should it not have been explained in the pull request?

PR #852, opened 2025-08-26
Reformat Python code with Ruff

hongquan:
When we copy code from eventyay-tickets, eventyay-talk, some definition code are missing.
I tried to recover some. For other things related to "metrics", I cannot identify where are the source, so I just delete them.


mariobehling (approved):

app/pyproject.toml
```diff
@@ -260,3 +260,19 @@ test = [
...
 
+[tool.ruff]
+line-length = 120
+
+[tool.ruff.format]
+indent-style = "space"
+quote-style = "single"
+
+[tool.ruff.lint]
+select = ["E", "F", "W", "I"]
```
norbusan, line 271: Maybe add 
```
    "UP",  # pyupgrade
    "PLE",
    "PLW", # Pylint
```
other than that looks fine.

norbusan (approved):

app/pyproject.toml
hongquan: Will add it in another PR.

PR #857, opened 2025-09-01
Change Item to Product

Sak1012:
Renamed `Item` model to `Product` for clarity and consistency. Changes include:

- Updated model definition
- Modified related forms
- Adjusted views to reflect new model name

Note that some permission names (e.g., `can_change_items`) remain unchanged for now. Static HTML file folder names were not updated in this commit.


mariobehling:
> Looks reasonable. @mariobehling do we allow for rewriting of migrations? This means that the server needs to be re-initialized, since this is a changed set of migrations, not an added one.

We dont have any production system running. So, we can just start from zero any time.

mariobehling (approved):

PR #871, opened 2025-09-15
Set User-Agent for HTTP client when accessing Wikimedia API

### LINKED ISSUE #796: Customize "User Agent" for the HTTTP client which we use for talking with wikimedia.org OAuth
The HTTP requests to acquire OAuth access token from wikimedia.org have been rate-limited. We should follow their [guideline](https://wikitech.wikimedia.org/wiki/Bot_traffic) to prevent it from happening again.

hongquan:
Resolve #796

## Summary by Sourcery

Configure the socialauth adapter to set a Wikimedia-compliant User-Agent and default request timeout for HTTP sessions, and update dependencies for oauthlib and django-allauth.

Enhancements:
- Override get_requests_session to add a custom User-Agent header based on site URL and admin contact and apply a default timeout.

Build:
- Bump oauthlib dependency to >=3.3.1.
- Update django-allauth[socialaccount] requirement to version >=65.11.2.

### DIFF
--- a/src/pretix/plugins/socialauth/adapter.py
+++ b/src/pretix/plugins/socialauth/adapter.py
@@ -1,14 +1,37 @@
+import functools
 import logging
+from typing import cast
 
 from allauth.core.exceptions import ImmediateHttpResponse
+from allauth.socialaccount import app_settings
 from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
-from django.http import HttpResponseRedirect
+from django.conf import settings
+from django.http import HttpRequest, HttpResponseRedirect
 from django.urls import reverse
 
 logger = logging.getLogger(__name__)
 
 
 class CustomSocialAccountAdapter(DefaultSocialAccountAdapter):
+    # Override to setup User-Agent to follow Wikimedia policy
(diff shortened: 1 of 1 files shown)


src/pretix/plugins/socialauth/adapter.py
```diff
@@ -1,14 +1,37 @@
...
+        # The self.request was populated by BaseAdapter.
+        dj_request = cast(HttpRequest, self.request)
+        site_url = dj_request.build_absolute_uri('/')
+        try:
+            contact = settings.ADMINS[0][1]
+        except (AttributeError, IndexError):
+            contact = 'webmaster@eventyay.com'
+        user_agent = f'eventyay/1.0 ({site_url}; {contact})'
+
+        import requests
```
norbusan, line 28: Why is the import in the function? Requests is most probably already loaded by whatever other part, so putting it into the top import is fine.

hongquan: The code from this line down, I only copy from the original `DefaultSocialAccountAdapter.get_requests_session()`.

PR #876, opened 2025-09-20
add platform specifics and resolve issue due to wrong copy order

Saksham-Sirohi:
Fixes for Docker setup and platform fixes for easy setup on ARM-based devices


app/Dockerfile
```diff
@@ -23,9 +26,6 @@ COPY ./entrypoint.sh .
 RUN sed -i 's/\r$//g' /usr/src/app/entrypoint.sh
 RUN chmod +x /usr/src/app/entrypoint.sh
 
-# copy project
-COPY . .
-
```
norbusan, line 28: That is not a really good change, since it makes building docker images when files changes slower.
But it can be left it either way, the uv sync is fast anyway.

PR #877, opened 2025-09-21
chore: datetime picker fix

### LINKED ISSUE #864: [enext] Date picker in event creation page doesn't work
**Describe the bug**

In the event creation page, the "Event start time" field does show date picker when clicking in. There are also some JS errors thrown in the browser console.

Image

My idea is that, we should use HTML native date picker, instead of relying on JS libraries.

Saksham-Sirohi:
fixes #864

### DIFF
--- a/app/eventyay/static/pretixcontrol/scss/_forms.scss
+++ b/app/eventyay/static/pretixcontrol/scss/_forms.scss
@@ -322,6 +322,12 @@ input[type=number].short {
   }
 }
 
+.form-group:has(.datepickerfield),
+.form-group:has(.timepickerfield),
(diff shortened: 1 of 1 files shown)


hongquan:
I don't know which issue this PR is trying to fix, but not #864 

-- Update --

I checked again, the fix in this PR seems to match the JS error in #864

Saksham-Sirohi:
Screenshot 2025-09-21 at 1 55 34 PM

hongquan (approved):

hongquan:
This PR is acceptable, though I don't favour continuing to use jQuery.

PR #878, opened 2025-09-22
Use native HTML datetime picker in event creation form

hongquan:


hongquan:
Close, in favour of https://github.com/fossasia/eventyay/pull/877

PR #895, opened 2025-10-02
chore(talk): Add name property to User model and substitue references with eventyay in mail system

Gagan-Ram:
Missed code from talk component to enext - part 1
- Substitues pretlax with eventyay.
- Adds name property to User model. This will potentially solves user model missing name attribute error which could have possibly come up.
E.g. While deleting speaker for submission proposal, in url `/orga/event/<event_slug>/submissions/<code>/speakers/`

Screenshot for above example url:
user_name_attr

## Summary by Sourcery

Replace references to the pretalx domain with eventyay in mail placeholders and admin templates, and expose a name property on the User model

New Features:
- Add name property to User model returning the user’s fullname

Enhancements:
- Update mail context and valid placeholders to use eventyay.com domain instead of pretalx.com
- Replace links in the orga admin template to point to eventyay.com and docs.eventyay.org


app/eventyay/base/models/auth.py
```diff
@@ -227,6 +227,10 @@ class Meta:
             'administrator': is_administrator,
         }
 
+    @property
+    def name(self):
```
hongquan, line 291: If other code tries to access `User.name` and fail, that code should be fixed.
Don't alias `fullname` to `name` because the `fullname` is clearer.
If you use only `name`,  people will ask: Is it first name, nick name for full name.

Gagan-Ram: In the **Talk** project, the `User` model originally had a `name` attribute. During migration to **Enext**, we replaced most instances of `name` with `fullname`.

However, while deleting a speaker from a submission proposal via the URL:
`/orga/event/<event_slug>/submissions/<code>/speakers/`,
we encountered an error stating that the `User` model has no attribute `name` — which was expected.

Now, we have two possible approaches:

1. Add a custom `name` property in the `User` model (using a Django property method).
2. Continue substituting `name` with `fullname` wherever this issue appears.
   However, identifying all such occurrences might be difficult since other models linked to `User` still reference the `name` field from Talk.

What would be the best way to proceed?

hongquan: You should give a log or screenshot to show what is that error.

hongquan:
image

You don't show this location, I cannot see the code to judge what we should do.

Gagan-Ram:
> You don't show this location, I cannot see the code to judge what we should do.

```
Internal Server Error: /orga/event/oais/submissions/QJQNPQ/speakers/delete
Traceback (most recent call last):
  File "/home/ubuntu/gsoc/finalize_pr/eventyay-tickets/app/.venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/gsoc/finalize_pr/eventyay-tickets/app/.venv/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/gsoc/finalize_pr/eventyay-tickets/app/.venv/lib/python3.12/site-packages/django/views/generic/base.py", line 105, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/gsoc/finalize_pr/eventyay-tickets/app/eventyay/orga/views/submission.py", line 225, in dispatch
    submission.remove_speaker(speaker, user=self.request.user)
  File "/home/ubuntu/gsoc/finalize_pr/eventyay-tickets/app/eventyay/base/models/submission.py", line 1033, in remove_speaker
    'name': speaker.name,
            ^^^^^^^^^^^^
AttributeError: 'User' object has no attribute 'name'
```
-----
code:
```
    def remove_speaker(self, speaker, orga=True, user=None):
        if self.speakers.filter(code=speaker.code).exists():
            self.speakers.remove(speaker)
            self.log_action(
                'eventyay.submission.speakers.remove',
                person=user or speaker,
                orga=orga,
                data={
                    'code': speaker.code,
                    'email': speaker.email,
                    'name': speaker.name,
                },
            )
```

hongquan:
Here I don't see any reason to keep `speaker.name`. Modify it to `speaker.fullname`.

Gagan-Ram:
> Here I don't see any reason to keep `speaker.name`. Modify it to `speaker.fullname`.

Sure, if we continue adding fullname, we need to make a separate PRs if we encounter the 'name' attribute error anywhere else.
Is that fine?

hongquan:
Yes, it is Ok if you cannot find all occurence of `user.name` in this PR. Try your best in a PR and if some bugs are not covered, we can continue to fix in next PRs.

hongquan:
Btw, using `ripgrep` tool, we can find more `user.name` accesses to fix in this PR:

image

Issue #906, opened 2025-10-06
Change root page message to new Eventyay installation text


mariobehling:
Change the current root page message to the following text below.

<img width="870" height="878" alt="Image" src="https://github.com/user-attachments/assets/36bec2cc-d59c-4806-a75c-026676d47241" />

This is an installation of Eventyay — the open-source event management platform.

Eventyay empowers communities and organizations to create, manage, and host events of all kinds — from conferences to online meetups.

If you’re looking to buy tickets, submit a session proposal, join an online event, or create your own, please follow the direct link to the specific event or organizer profile.

To log in and access your events, please head over here.

Enjoy!

mariobehling:
The last PR resulted in a scroll bar on the right and the link to the "head over here" directs to https://next.eventyay.com/control/admin/video/ instead of https://next.eventyay.com/common/.

Saksham-Sirohi:
Hi @mariobehling, can you share something like a video, as I'm trying this locally, and this seems to work perfectly

mariobehling:
The scrollbar probably depends on the screen size. Move the box a bit up. That should solve it. The "head over" link needs to be changed to /common.

<img width="2560" height="1291" alt="Image" src="https://github.com/user-attachments/assets/2570efe9-cd97-4b81-b582-08947bdaf4ae" />

mariobehling:
Still experiencing a scroll bar. Please reduce the area above the central eventyay logo a bit and test it on different screen sizes, e.g. laptops.

<img width="1466" height="1270" alt="Image" src="https://github.com/user-attachments/assets/c1a535d7-24f5-4829-aec5-8c23e3d32e3e" />

Saksham-Sirohi:
<img width="1308" height="801" alt="Image" src="https://github.com/user-attachments/assets/0e2247b9-32bf-4717-9995-c15a15c5b11c" />

No scrollbar I have removed additional margin and widened the text box a little. Tested with the smallest resolutions of netbooks.

PR #1042, opened 2025-10-13
chore: smaller margins and higher widths of content on welcome page

Saksham-Sirohi:
Solution #906

image

## Summary by Sourcery

Adjust layout spacing and sizing on the authentication welcome page

Enhancements:
- Reduce top margin of the footer logo and auth panel from 7.5% to 3.5%
- Remove bottom margin from the footer logo
- Increase max width of the auth form container from 380px to 420px

### DIFF
--- a/app/eventyay/cfp/templates/cfp/event/invitation.html
+++ b/app/eventyay/cfp/templates/cfp/event/invitation.html
@@ -30,7 +30,7 @@
         </p>
 
         {% if submission.abstract %}
-            <div class="card ml-auto mr-auto col-md-9">
+            <div class="card ml-auto mr-auto p-0 col-md-9">
                 <div class="card-header">{% translate "Abstract:" %} {{ submission.title }}</div>
                 <div class="card-body mt-0 mb-0 mr-3 ml-3">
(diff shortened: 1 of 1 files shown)


Saksham-Sirohi:
> Thanks! As we had issues with these let's pay attention to detail:
> 
>     1. **Remove duplicate/obsolete declarations:**
>        `margin-top` is defined twice in two places (old value and new value). Keep only the updated one to avoid confusion for future maintainers.
>        Also, make sure the line `margin-bottom: 30px;` is actually removed rather than left commented or overridden — otherwise it adds noise to the file.
> 
>     2. **Avoid using `%` for vertical spacing:**
>        Percent-based margins are calculated from the _container width_, not height. This can cause inconsistent vertical spacing across screen sizes.
>        Please switch to a unit like `rem`, `px`, or a responsive clamp, e.g.:
> 
> 
> ```css-scss
> margin-top: clamp(8px, 2vh, 24px);
> ```
> 
> This gives more predictable and accessible spacing behavior.

part 1 is not defined twice it is for two separate classes also that line is removed
part 2 changed to vh from %

PR #1058, opened 2025-10-17
chore(enext): change config file load order to prioritize system config over local eventyay.cfg

Gagan-Ram:
The `app/eventyay.cfg` file contains the following default configuration:

```
[redis]
location=redis://eventyay-next-redis/0
; Remove the following line if you are unsure about your redis' security
; to reduce impact if redis gets compromised.
sessions=true

[celery]
backend=redis://eventyay-next-redis/1
broker=redis://eventyay-next-redis/2
```

When running locally, developers need to modify these entries to use `localhost:<port>`. However, this becomes inconvenient when switching between branches, as the file must be reverted to its original state before making any branch changes.

Previously, when running `django manage.py runserver`, Django loaded the configuration from the local `eventyay.cfg` file by default.

This PR changes the configuration load order to prioritize `/etc/eventyay/` first, so developers no longer need to modify their local `eventyay.cfg` for local testing.

Additionally, the README.rst file has been updated with clear instructions on how to set up and configure Redis locally, making the development process smoother and more consistent.

## Summary by Sourcery

Reorder config file precedence to apply system configurations after local files, reducing the need to modify local eventyay.cfg, and enhance documentation with guidance for local Redis setup.

Enhancements:
- Reorder configuration loading to prioritize system-level eventyay.cfg over local copies

Documentation:
- Update README with detailed instructions for setting up and configuring Redis locally


hongquan:
No, settings in user local directory should override the system ("/etc/").
Given a machine that has two users, using the same software. Each user wants to use the software in a different way. The only way to allow that, is that the software loads different settings for each user (from local user directory). If that software prefers settings in "/etc", it will always loads the same settings for all users.

The issue you are facing:

> However, this becomes inconvenient when switching between branches, as the file must be reverted to its original state before making any branch changes.

are actually due to wrong convention of config files. We should always allow one "local" config file that each developer is free to modify and the file is NOT tracked by Git.

The configuration files should be:

- A file to contain default config. This file is tracked by Git. We can call it _evenyay.base.toml_.
- A developer-local config file, contains the change that developer needs. This file is not tracked by Git. We can name it _eventyay.local.toml_. This file should not contain all config. For example, if developer wants to change `REDIS_URL`, this file should only contain the Redis config.


But, except when we plan to distribute eventyay via OS package manager (like APT on Debian / Ubuntu), we don't need to support config file in "/etc". Let's remove support of "/etc/" config to make it simple.

PR #1062, opened 2025-10-17
fix: statistics error page

Gagan-Ram:
before:
schedule error

after:
stats_fix

## Summary by Sourcery

Fix a statistics error on the organizer views by converting submission and talk IDs to strings when filtering log entries for timeline data

Bug Fixes:
- Convert raw submission timeline IDs to strings before querying log entries
- Convert accepted talk timeline IDs to strings before querying log entries


hongquan (approved):
Approved. But this is only short-term solution.

The long-term is to fix the type of `LogEntry.object_id`. Currently, we are using `JSONField` for it. It prevents the database from doing optimization for fast lookup. In long run, it will slow down our system.

Issue #692, opened 2025-05-29
Add option to make changes to existing orders to customer data tab in global event settings


mariobehling:
There is an option to configure the time when changes to existing orders can be done in the the tab "Timeline" on Global settings of events. 

An example URL is here https://app-test.eventyay.com/tickets/control/event/MarioOrg/txnvy/settings/#tab-0-5-open (but only accessible to organiser).

**Step 1**: Move this to the bottom of the tab General settings > Customer

![Image](https://github.com/user-attachments/assets/6dca96e2-59c6-4a91-a4ea-f1e1888c7e87)

**Step 2**: Implement an additional option as in the below screenshot to "Allow customers to modify their information" Optional
- No modifications after order was submitted
- Only the person who ordered can make changes
- Both the attendee and the person who ordered can make changes

An example URL of this page is  e.g. https://app-test.eventyay.com/tickets/control/event/MarioOrg/txnvy/settings/#tab-0-2-open

![Image](https://github.com/user-attachments/assets/3a3a889f-1c22-4b01-a6df-40e14c80e8d0)

Sana0523:
@mariobehling Could you please confirm if you want the Step 1 changes to be made in this tab (below "Customer") as shown in the screenshot?

![Image](https://github.com/user-attachments/assets/cce1c13d-03ef-4874-b6a2-ae1bde801055)

This would be followed up by Step 2 changes. Please let me know if this placement works or if you'd prefer it elsewhere.

Gagan-Ram:
@mariobehling could you please check on this

PR #1085, opened 2025-10-22
feat(orders): add attendee order modification setting and move order change config to Customer tab

Gagan-Ram:
#692 

https://github.com/user-attachments/assets/e84a19d0-59c1-4fa7-96c2-ba27069be0f4

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -35,6 +35,6 @@ npminstall:
 	OUT_DIR=$(shell pwd)/eventyay/static npm run build --prefix=eventyay/frontend/schedule-editor
 
 webappbuild:
-	npm install --prefix=eventyay/webapp
+	npm ci --prefix=eventyay/webapp
 	OUT_DIR=$(shell pwd)/eventyay/static npm run build --prefix=eventyay/webapp
 
--- a/app/eventyay/presale/views/event.py
+++ b/app/eventyay/presale/views/event.py
@@ -876,7 +876,7 @@ class JoinOnlineVideoView(EventViewMixin, View):
             # no order placed yet
             return False, None, None
         # Check if Event allow all ticket type to join
-        if self.request.event.settings.venueless_all_items:
+        if self.request.event.settings.venueless_all_products:
             return True, None, order_list[0]
         list_allow_ticket_type = self.request.event.settings.venueless_products
         if not list_allow_ticket_type:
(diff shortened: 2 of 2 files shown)


app/eventyay/base/models/orders.py
```diff
@@ -725,6 +725,18 @@ def assign_code(self):
                 length += 1
                 iteration = 0
 
+    def modify_deadline(self):
+        modify_deadline = self.event.settings.get('last_order_modification_date', as_type=RelativeDateWrapper)
```
norbusan: Please don't name local variable the same as the function, that is really confusing

app/eventyay/control/templates/pretixcontrol/event/settings.html
```diff
@@ -101,6 +101,19 @@ <h4>{% trans "Other settings" %}</h4>
...
+                {% bootstrap_field sform.allow_modifications layout="control" %}
+                <div data-display-dependency-container>
+                    <div data-display-dependency="#id_settings-allow_modifications_1">
+                        {% bootstrap_field sform.last_order_modification_date layout="control" %}
+                        {% bootstrap_field sform.allow_modifications_after_checkin layout="control" %}
+                    </div>
+                    <div data-display-dependency="#id_settings-allow_modifications_2">
+                        {% bootstrap_field sform.last_order_modification_date layout="control" %}
+                        {% bootstrap_field sform.allow_modifications_after_checkin layout="control" %}
+                    </div>
```
norbusan: Will those two blocks not be rendered the very same way?

norbusan (changes requested):

app/eventyay/control/templates/pretixcontrol/event/settings.html
Gagan-Ram:
```
if (dependent.is("[data-inverse]")) {
        enabled = !enabled;
   }
```
This code was missing in `app/eventyay/static/pretixcontrol/js/ui/main.js`, that's why data-inverse attribute wasn't working in my code, which made me to use the above code.
In the new commit I have modified the code, please check.
Thank you

app/eventyay/base/models/orders.py
Gagan-Ram: Understood.
I resolved in new commit, please check.

PR #1090, opened 2025-10-22
Add webapp build step to Docker image build process

suhailnadaf509:
## Summary
Adds the webapp build process to the Docker build to ensure the frontend is built automatically.

## Changes
- Added `npm install` and `npm run build` for `eventyay/webapp` in the Dockerfile.

## Problem
The webapp wasn’t being built during Docker image creation.

## Solution
Run:
1. `npm install --prefix=eventyay/webapp`
2. `npm run build --prefix=eventyay/webapp`

## Testing
Run `docker compose up -d --build` and verify:
- The image builds successfully  
- The webapp assets are generated and served correctly  

## Impact
Ensures consistent, automatic webapp builds in Docker without manual steps.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     command: python manage.py runserver 0.0.0.0:8000
     volumes:
       - ./app/eventyay:/usr/src/app/eventyay
-      - /usr/src/app/eventyay/webapp/dist
+      - /usr/src/app/eventyay/static/webapp
       - /usr/src/app/eventyay/webapp/node_modules
     ports:
       - 8000:8000
@@ -25,7 +25,7 @@ services:
     entrypoint: celery -A eventyay worker -l info
     volumes:
       - ./app/eventyay:/usr/src/app/eventyay
-      - /usr/src/app/eventyay/webapp/dist
+      - /usr/src/app/eventyay/static/webapp
       - /usr/src/app/eventyay/webapp/node_modules
     env_file:
       - ./.env.dev


suhailnadaf509:
> make this incorporate inside npminstall

I think its better to keep this modular so i've implemented webappbuild

Saksham-Sirohi:
A better unified approach for all frontend apps is in development. This can be merged then to use the webapp currently.

suhailnadaf509:
@Sak1012 I've resolved the conflicts

Saksham-Sirohi (commented):
New changes look good to the pr can you make so that files get built inside static, following our approach in other frontend apps.

docker-compose.yml
```diff
@@ -8,6 +8,8 @@ services:
     command: python manage.py runserver 0.0.0.0:8000
     volumes:
       - ./app/eventyay:/usr/src/app/eventyay
+      - /usr/src/app/eventyay/webapp/dist
+      - /usr/src/app/eventyay/webapp/node_modules
```
norbusan, line 12: What shoud this do? That doesn't make sense, and is incorrectly formatted.
Please remove the two lines.

docker-compose.yml
```diff
@@ -23,6 +25,8 @@ services:
     entrypoint: celery -A eventyay worker -l info
     volumes:
       - ./app/eventyay:/usr/src/app/eventyay
+      - /usr/src/app/eventyay/webapp/dist
+      - /usr/src/app/eventyay/webapp/node_modules
```
norbusan, line 29: Same here

norbusan (changes requested):

suhailnadaf509:
@norbusan When ./app/eventyay is mounted, it would normally overwrite webapp/dist and webapp/node_modules from the image. The anonymous volumes (- /usr/src/app/eventyay/webapp/dist, - /usr/src/app/eventyay/webapp/node_modules) prevent that, preserving the built assets.

suhailnadaf509:
> New changes look good to the pr can you make so that files get built inside static, following our approach in other frontend apps.

@Saksham-Sirohi Several files currently rely on hardcoded paths, which would all need to be updated. Implementing custom URL patterns for /video would also require refactoring the existing logic. Additionally, the videospa view serves index.html, which itself uses a hardcoded path. Overall, this represents a significant change and might be better addressed in a later update.

app/Dockerfile.prod
```diff
@@ -76,6 +80,7 @@ RUN chown -R app:app $APP_HOME
 COPY --from=builder --chown=app:app $APP_HOME/.venv $APP_HOME/.venv
 COPY --from=builder --chown=app:app $APP_HOME/eventyay/static/global-nav-menu $APP_HOME/eventyay/static/global-nav-menu
 COPY --from=builder --chown=app:app $APP_HOME/eventyay/static/schedule-editor $APP_HOME/eventyay/static/schedule-editor
+COPY --from=builder --chown=app:app $APP_HOME/eventyay/webapp/dist $APP_HOME/eventyay/webapp/dist
```
norbusan: One more thing - that will not work in production. 
All files that need to be served via nginx need to be in the static directory. This one is shared with the nginx component.

Please follow the example above and use
```
RUN OUT_DIR=/home/app/web/eventyay/static npm run build --prefix=eventyay/webapp
...
COPY --from=builder --chown=app:app $APP_HOME/eventyay/static/webapp $APP_HOME/eventyay/static/webapp
```

mariobehling (approved):
Ok, thanks. Let's test the deployment.

PR #1095, opened 2025-10-23
Add dedicated WebSocket container with Daphne for real-time features

suhailnadaf509:
### Overview

Fixes WebSocket connection errors (Not Found: /ws/event/<slug>/) by adding a Daphne service for WebSockets while keeping Gunicorn for HTTP. Nginx now routes /ws/* to Daphne and all other requests to Gunicorn.

### Problem

Gunicorn (WSGI) doesn’t support WebSockets, causing real-time features like chat, polls, and Q&A to fail.

### Solution

- Gunicorn (8000): HTTP traffic
- Daphne (8001): WebSockets
- Nginx: Routes and upgrades /ws/* requests
- Added timeouts and upgrade headers

### Result

1. WebSocket features work again
2. No impact on existing HTTP setup
3. Easy rollback and no DB/code changes

## Summary by Sourcery

Add a dedicated WebSocket container using Daphne and update Nginx routing to restore real-time features while retaining existing HTTP setup through Gunicorn.

New Features:
- Introduce a dedicated Daphne container for handling WebSocket connections.

Enhancements:
- Update Nginx configuration to route /ws/* paths to the Daphne service with proper WebSocket upgrade and timeout headers.
- Keep Gunicorn serving HTTP traffic on port 8000 while offloading real-time traffic to Daphne on port 8001 without code or DB changes.

PR #1096, opened 2025-10-23
Fix Django settings initialization in routing.py for WebSocket container

suhailnadaf509:
### Problem

After PR #1095, the WebSocket container failed to start with an error saying Django settings weren’t configured (REDIS_USE_PUBSUB access before initialization).

### Cause

Daphne loads eventyay.config.routing:application directly, which imports the Redis utility before Django settings are initialized.

### Fix

Set DJANGO_SETTINGS_MODULE at the top of routing.py, matching how it’s done in asgi.py, wsgi.py, and manage.py.

## Summary by Sourcery

Bug Fixes:
- Preload Django settings in routing.py to fix startup failure in the WebSocket container

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -36,5 +36,5 @@ npminstall:
 
 webappbuild:
 	npm install --prefix=eventyay/webapp
-	npm run build --prefix=eventyay/webapp
+	OUT_DIR=$(shell pwd)/eventyay/static npm run build --prefix=eventyay/webapp
 


hongquan:
Both this PR and #1095 sound not right.
Daphne (or any ASGI server) should load the ASGI app from _asgi.py_ file (this file is Django convention, similar to _wsgi.py_ for WSGI app).
So the command to run Daphne should load the _asgi.py_ and the `DJANGO_SETTINGS_MODULE` environment variable should be set there, not `routing.py`.

mariobehling:
Ok, I created an issue here https://github.com/fossasia/eventyay-tickets/issues/1098
Does this look correct? Please comment.

Issue #1113, opened 2025-10-26
Resolve 500 error on talk dashboard


mariobehling:
The talk dashboard on the test deployment at https://next.eventyay.com/orga/event/ shows a 500 error. Please resolve it.

<img width="1490" height="1096" alt="Image" src="https://github.com/user-attachments/assets/52655a9f-30fa-4401-bae7-21a78cbaf6f9" />

Gagan-Ram:
@mariobehling I don't see this problem, neither in my local machine nor in production so far.
are you still facing this issue?

<img width="1917" height="971" alt="Image" src="https://github.com/user-attachments/assets/bce53545-ce99-41d2-b3cf-d2626d47d3ed" />

mariobehling:
Logging in with eventyay-devops shows this issue at the moment at https://next.eventyay.com/orga/event/

Gagan-Ram:
Could you please provide the logs

Sak1012:
Here are the logs it appears to be a `TypeError`
```
eventyay-next-web         | ERROR 2025-10-27 15:29:37,504 log: Internal Server Error: /orga/event/
eventyay-next-web         | Traceback (most recent call last):
eventyay-next-web         |   File "/home/app/web/.venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
eventyay-next-web         |     response = get_response(request)
eventyay-next-web         |                ^^^^^^^^^^^^^^^^^^^^^
eventyay-next-web         |   File "/home/app/web/.venv/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
eventyay-next-web         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
eventyay-next-web         |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
eventyay-next-web         |   File "/home/app/web/.venv/lib/python3.12/site-packages/django/views/generic/base.py", line 105, in view
eventyay-next-web         |     return self.dispatch(request, *args, **kwargs)
eventyay-next-web         |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
eventyay-next-web         |   File "/home/app/web/.venv/lib/python3.12/site-packages/django/views/generic/base.py", line 144, in dispatch
eventyay-next-web         |     return handler(request, *args, **kwargs)
[...]

PR #1126, opened 2025-10-27
fix(event): set event end date by default

Gagan-Ram:
fixes: #1113

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -35,6 +35,6 @@ npminstall:
 	OUT_DIR=$(shell pwd)/eventyay/static npm run build --prefix=eventyay/frontend/schedule-editor
 
 webappbuild:
-	npm install --prefix=eventyay/webapp
+	npm ci --prefix=eventyay/webapp
 	OUT_DIR=$(shell pwd)/eventyay/static npm run build --prefix=eventyay/webapp
 
--- a/app/eventyay/api/serializers/event.py
+++ b/app/eventyay/api/serializers/event.py
@@ -664,6 +664,7 @@ class EventSettingsSerializer(SettingsSerializer):
         'locales',
         'locale',
         'region',
+        'allow_modifications',
         'last_order_modification_date',
         'allow_modifications_after_checkin',
         'show_quota_left',
--- a/app/eventyay/base/configurations/default_setting.py
+++ b/app/eventyay/base/configurations/default_setting.py
@@ -1210,6 +1210,28 @@ DEFAULT_SETTINGS = {
             label=_('Hide all unavailable dates from calendar or list views'),
         ),
     },
+    'allow_modifications': {
+        'default': 'order',
+        'type': str,
+        'form_class': forms.ChoiceField,
+        'serializer_class': serializers.ChoiceField,
+        'serializer_kwargs': dict(
+            choices=(
+                ('no', _('No modifications after order was submitted')),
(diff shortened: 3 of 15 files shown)


hongquan (changes requested):
I expect to see DB migration to change `date_to` to non-nullable.

Gagan-Ram:
This PR requires us to drop all tables of our database.
```
python manage.py dbshell
```
Then in the PostgreSQL prompt:
```
DROP SCHEMA public CASCADE;
CREATE SCHEMA public;
GRANT ALL ON SCHEMA public TO postgres;
GRANT ALL ON SCHEMA public TO public;
```

hongquan:
You should do it in one of these two ways:

1. Accept dropping all tables (refresh our DB):

- Just delete all _migrations/xxx.py_ files and recreate. Every modules will starting from `0001_initial.py`.

2. Keep old data:

- Create new migration file (don't modify _0001_initial.py_).
- Just add a step of "[data migration](https://docs.djangoproject.com/en/5.2/topics/migrations/#data-migrations)" to the same file, where you fill a default value for `date_to` field of existing records (ones with null `date_to`), before converting the field to "non-nullable".


Your current implement is bad because it modifies an existing DB migration file.
You can modify existing DB migration files only if the changes don't cause change in DB schema (tables, columns).

The 1st option is Ok for us because we don't serve any real event yet.

Gagan-Ram:
> 3\. Keep old data:
> 
> 
>     * Create new migration file (don't modify _0001_initial.py_).
> 
>     * Just add a step of "[data migration](https://docs.djangoproject.com/en/5.2/topics/migrations/#data-migrations)" to the same file, where you fill a default value for `date_to` field of existing records (ones with null `date_to`), before converting the field to "non-nullable".

Thank you, this is what I was looking for.
I have made the changes, please check

app/eventyay/base/migrations/0006_auto_20251030_1557.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 5.2.5 on 2025-10-30 15:57
+
+from datetime import timedelta
+from django.db import migrations
+
+def fill_date_to(apps, _schema_editor):
+    Event = apps.get_model('base', 'Event')
+    for event in Event.objects.filter(date_to__isnull=True):
```
hongquan: Instead of calling `.save()` in a `for` loop, you can just use `Event.objects.update()` with [`F()`](https://docs.djangoproject.com/en/5.2/ref/models/expressions/#f-expressions). It runs faster.

hongquan (changes requested):
@mariobehling wants to have a production system running end of this week.
We should have a fresh DB. Please delete all DB migration files and recreate from 0001. We don't have many chances to do clean up of DB migration files.

app/eventyay/base/migrations/0006_auto_20251030_1557.py
Gagan-Ram: Thank you

Gagan-Ram:
@Sak1012 We need to drop all tables of our database because of newly created migration files
```
python manage.py dbshell
```
Then in the PostgreSQL prompt:
```
DROP SCHEMA public CASCADE;
CREATE SCHEMA public;
GRANT ALL ON SCHEMA public TO postgres;
GRANT ALL ON SCHEMA public TO public;
```

hongquan (approved):

Issue #984, opened 2025-10-09
Unify Event Modes Across Components (Tickets, Talk, Video)


mariobehling:
Currently, the new eventyay unified system mixes tickets-component and talk-component modes inconsistently.
When an event is created, the **Talk** component is automatically live, while the **Tickets** component starts in **Test Mode**.
We need to unify these flows and ensure a consistent mode system across all components, with clear visibility in both the **common dashboard** and **component dashboards**.

### **Goal:**

Implement a unified state flow for all components, ensuring that:

* All components start **offline** when a new event is created.
* Only the **organizer** can change the mode.
* The mode can be changed either:

  * On the **common dashboard**, or
  * Within each **component dashboard**.
* The **common dashboard** should clearly display the current mode of each component with an actionable link, e.g.:

  > “Your Ticket component is live. Click here to change it.”

#### **1. Tickets Component**

* Initial: **Offline Mode**
* Step 1: Organizer can enable **Test Mode**
* Step 2: Organizer can enable **Live Mode**

#### **2. Talk Component**

* Initial: **Offline Mode**
* Step 1: Organizer can enable **Live Mode**

#### **3. Video Component**

* Initial: **Offline Mode**
* Step 1: Organizer can enable **Live Mode**

### **User Interface**
In the dashboard propose and implement a suitable UI change to show the state of the event component, e.g. live or offline mode.

<img width="2494" height="998" alt="Image" src="https://github.com/user-attachments/assets/077d1031-f5b8-4653-b46d-d25e647f6be3" />

### **Acceptance Criteria**

* [ ] New events start with all components in **offline** mode.
* [ ] Only organizer-level users can toggle modes.
* [ ] Dashboard UI shows current mode and provides a link/button to change it.
* [ ] Ticket flow includes an intermediate **test mode** before live.
* [ ] Talk and Video components have only **offline → live** transition.
* [ ] State changes are reflected consistently in both component dashboards and the unified dashboard.
* [ ] Appropriate backend state fields and API endpoints are aligned across components.
* [ ] User Interface in the common dashboard that shows state of component and matches design.

---

### **Notes**

* Align Pretix’s `test/live` logic with Pretalx’s `is_public` (live/offline) state.
* Make sure public visibility and payment/live activation are separate but consistent.
* Example UI element:

  ```
  Tickets: 🟡 Test Mode → [Switch to Live]
  Talks: 🔴 Offline → [Make Live]
  Video: 🔴 Offline → [Make Live]
[...]

Issue #1014, opened 2025-10-10
Refactor: Align Talk and Video Component URLs with Tickets Component Schema


mariobehling:
<p>The <strong>Talk</strong> and <strong>Video</strong> components currently use different URL structures from the <strong>Tickets</strong> component.<br>
To ensure <strong>consistency across the unified Eventyay platform</strong>, all public and admin URLs in Talk and Video should follow the same schema:</p>
<pre><code>https://next.eventyay.com/{orgname}/{eventslug}/[component]/
</code></pre>
<p>There are exceptions for the Talk component where the <code inline="">/talk</code> slug is not really needed.</p>
<p>This includes all subpages such as schedule, sessions, speakers, rooms, and admin views.</p>
<hr>
<h3>🧭 Expected Unified URL Structure</h3>

Component | Example Old URL | New Unified URL
-- | -- | --
Tickets (unchanged) | — | https://next.eventyay.com/{orgname}/{eventslug}/
Talk | https://next.eventyay.com/{eventslug}/ | https://next.eventyay.com/{orgname}/{eventslug}/talk
Video | https://next.eventyay.com/video/{eventslug} | https://next.eventyay.com/{orgname}/{eventslug}/video

<p>➡️ <strong>All related subpages must follow the same pattern</strong>, for example:</p>
<ul>
<li>
<p><code inline="">{orgname}/{eventslug}/cfp</code></p>
</li>
<li>
<p><code inline="">{orgname}/{eventslug}/sessions</code></p>
</li>
<li>
<p><code inline="">{orgname}/{eventslug}/speakers</code></p>
</li>
<li>
<p><code inline="">{orgname}/{eventslug}/video/schedule</code></p>
</li>
<li>
<p><code inline="">{orgname}/{eventslug}/video/sessions</code></p>
</li>
<li>
<p><code inline="">{orgname}/{eventslug}/video/rooms/{id}</code></p>
</li>
<li>
<p><code inline="">{orgname}/{eventslug}/video/admin/...</code></p>
</li>
</ul>
<hr>
<h3>🧩 Requirements</h3>
<ul class="contains-task-list">
<li class="task-list-item">
<p><input type="checkbox" disabled=""> Update routing for all public and admin pages in Talk and Video to use <code inline="">{orgname}/{eventslug}</code>.</p>
</li>
<li class="task-list-item">
[...]

PR #1170, opened 2025-10-31
fix default event live mode in talk

Gagan-Ram:
fixes: #984

## Summary by Sourcery

Switch event visibility from the deprecated is_public flag to a new live flag and ensure it is used consistently as the default live mode across the Talk codebase

Bug Fixes:
- Fix default event live mode by renaming and redirecting is_public logic to live

Enhancements:
- Replace all model, fixture, and default event creation uses of is_public with live

Tests:
- Update API serializers, views, templates, permissions, rules, middleware, and management commands to use live instead of is_public
- Rename and adjust all tests and assertions to refer to event.live in place of event.is_public

### DIFF
--- a/app/eventyay/agenda/templates/agenda/base.html
+++ b/app/eventyay/agenda/templates/agenda/base.html
@@ -1,6 +1,7 @@
 {% extends "common/base.html" %}
 
 {% load event_tags %}
+{% load staff_session %}
 {% load i18n %}
 {% load rules %}
 {% load safelink %}
@@ -14,7 +15,7 @@
 {% endblock custom_header %}
 
 {% block nav_link %}
-    {% has_perm "base.list_schedule" request.user request.event as can_view_schedule %}
+    {% has_event_perm "base.list_schedule" request.user request request.event as can_view_schedule %}
     {% if can_view_schedule %}
         {{ request.event.urls.schedule }}
     {% else %}
--- a/app/eventyay/base/configurations/default_setting.py
+++ b/app/eventyay/base/configurations/default_setting.py
@@ -2080,11 +2080,9 @@ Your {event} team"""
             ext_whitelist=('.png', '.jpg', '.gif', '.jpeg'),
             max_size=10 * 1024 * 1024,
             help_text=_(
-                'If you provide a header image, it will be displayed instead of your event’s color '
-                'and/or header pattern at the top of all event pages. It will be center-aligned, '
-                'so when the window shrinks, the center parts will continue to be displayed, you '
-                'can increase the size with the setting below. The image will not be stretched. ' 
-                'Please do not upload files larger than 10.0MB!'
+                'This image appears at the top of all event pages, replacing the default color or pattern. '
+                'It is center-aligned and not stretched, ensuring the middle part remains visible on smaller screens. '
+                'We recommend an image at least 1170 px wide and 120 px in height for best results.'
             ),
         ),
         'serializer_class': UploadedFileField,
@@ -2102,10 +2100,9 @@ Your {event} team"""
             ext_whitelist=('.png', '.jpg', '.gif', '.jpeg'),
             max_size=10 * 1024 * 1024,
             help_text=_(
-                'If you provide a logo image, we will by default not show your event name and date '
-                'in the page header. By default, the logo will be scaled down to a height of 140px. '
-                'We recommend not using small details on the picture '
-                'as it will be resized on smaller screens.'
+                'When you upload a logo, the event name and date will not appear in the header. '
+                'The logo scales to 140 px in height while maintaining aspect ratio. '
+                'We recommend not using small details as it will be resized on smaller screens.'
             ),
         ),
         'serializer_class': UploadedFileField,
@@ -2143,10 +2140,9 @@ Your {event} team"""
             ext_whitelist=('.png', '.jpg', '.gif', '.jpeg'),
             max_size=10 * 1024 * 1024,
             help_text=_(
-                'If you provide a logo image, we will by default not show your organization name '
-                'in the page header. By default, we show your logo with a size of up to 1140x120 pixels. You '
-                'can increase the size with the setting below. We recommend not using small details on the picture '
-                'as it will be resized on smaller screens.'
+                'This image appears at the top of all organizer pages, replacing the default color or pattern. '
+                'It is center-aligned and not stretched, ensuring the middle part remains visible on smaller screens. '
+                'We recommend an image 1140 px wide and 120 px in height (can be increased with the setting below).'
             ),
         ),
         'serializer_class': UploadedFileField,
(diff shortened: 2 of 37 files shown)


Gagan-Ram:
This PR partially fixes #984 
@hongquan @mariobehling The talk component overlapped with ticket component's `is_public` field. I have made talk component to align with ticket component's `live` field.
Although this seems work, I think there are potential issues:
1. If talk component goes live, ticket component also goes live even if there products and payment gateway is not created and vice-versa.
2. Going forward we will implement the creation process of each component individually, the same above problem will again appear when user choose to enable any other component after the successfully setting up everything to make 1st component to go live.

so my suggestion is that we will have a new event model field named `talk_live` which acts similar to `live` field for tickets.

hongquan:
Based on #984:
- The ticket system should have 3 states: `offline`, `test mode`, `live mode`.
- The talk & video system each has 2 states: `offline`, `live`.
- Talk & video systems can not be `live` if ticket system is offline.
- Talk & video states are independent, but depending on ticket system.

If so, I propose to add these fields to `Event` model:

- `mode`, take the value from this enum:

```py
class EventMode(TextChoices):
    OFFLINE = 'o'
    TEST = 't'
    LIVE = 'l'
```
- `is_talk_live` (`bool`)
- `is_video_live` (`bool`)

The later two fields are boolean field because boolean type has two values.

@mariobehling I wonder if `live` is terminology correct, because when I hear it, I often think as in "live broadcast".

mariobehling:
Yes, live sounds correct.

Gagan-Ram:
> * Talk & video systems can not be `live` if ticket system is offline.

In future, we will give an option to enable ticketing component later. How do we handle event 'live' if ticket component is created later?

> * Talk & video states are independent, but depending on ticket system.

Not really when only Talk component is created independently and ticket component is created later or not created at all.

Gagan-Ram:
TLDR; The Talk component can only go live if the Tickets component is already live. However, if the Tickets component fails to go live (for example, due to missing quotas or unconfigured payments), the Talk component cannot go live either.


My understanding is that an event can be initially created with either the Talk or Ticket component, or with both simultaneously.

* The Video component depends on the Event itself.
* The Talk and Ticket components can function independently when created on their own.

Given that, it’s not entirely clear which component should act as the base. For example, Talk and Video are independent, but both depend on the Ticket system in certain scenarios.

To simplify this setup, I propose the following approach:

---

### **Ticketing Component Conditions**

For the Ticketing component to go live, two conditions must be met:

1. At least one **quota** must be configured.
2. A **payment gateway** must be enabled.

---

### **Talk Component Conditions**

The Talk component does not have strict prerequisites for going live. The system automatically creates a default `sessionType` when the component is enabled.

---

### **Possible Scenarios**

1. **User enables both Talk and Tickets together:**
   Everything works smoothly—no special handling required.

2. **User enables Tickets first, then Talk:**
   The event can stay live since Talk can be activated without additional setup. The Talk component simply needs its default session type, which is created automatically.

3. **User enables Talk first, then Tickets:**
   In this case, once the user enables Ticketing, the event must automatically switch to **offline**. The system should prompt the user to set up the necessary ticketing configurations (quota, product, payment gateway). The event can only go live again after these conditions are met.

---

### **Model Changes**

We introduce three new fields in the Event model:

* `is_tickets_enabled`
* `is_talk_enabled`
* `is_ticket_created_earlier`

These fields update automatically when the user enables or disables components.

---

### **Logic Overview**

```python
def enable_ticket_component():
    if self.is_ticket_created_earlier:
        # Verify quota and payment setup
        # If missing, mark event as offline
        self.live = False
    else:
        self.is_ticket_created_earlier = True
        self.is_tickets_enabled = True
        self._build_initial_ticket_data()  # Defaults to live=False on first setup

def enable_talk_component():
    # No strict conditions required for activation
    pass

def take_event_live():
    if self.is_tickets_enabled or (self.is_tickets_enabled and self.is_talk_enabled):
        # Require quota and payment setup before going live
        
    elif self.is_talk_enabled and not self.is_tickets_enabled:
        # Allow going live (Talk has minimal default setup)
        
```

mariobehling:
I think we need to define more clearly what “go live” means versus “is published.”

We are running into conceptual and practical issues when “live” is tied too closely to the Tickets component. As mentioned in [issue #1014](https://github.com/fossasia/eventyay/issues/1014), the **root page of an event** should, in future, always display the basic event information. Although this data technically resides in the Tickets component, it should be treated as part of a **common layer** of the event.

This has several implications:

* The event info (root page) should be **viewable when the event is published**, even if the Tickets component is not live or not used at all.
* Organizers should be able to **publish or unpublish** an event independently from enabling Tickets or Talks.
* Once an event is over and ticketing is no longer relevant (e.g., payments disabled), the event info page should remain accessible in a published state.
* Similarly, even before tickets go on sale, the Call for Speakers or other content should be accessible via this main event page.

To achieve this separation, we may need an **additional status field**, e.g., `is_published`, distinct from `is_live`.

Key question: How do we implement this distinction so that the event’s root page can be published (or unpublished) and is accessible (or inaccessible) even when the Tickets component is not live?

Gagan-Ram:
> **root page of an event**

When you referred root page of an event, is it /common/event/<org_slug>/<event_slug> or presale page where currently user buys tickets from?

mariobehling:
This is an example URL of a root page https://next.eventyay.com/a-team/summit/

Screenshot from 2025-11-04 09-35-00

hongquan:
I think that the `Event` model should have these fields:

```py
published_at = DateTimeField(null=True)  # For the `Event` itself.
is_tickets_live = BooleanField(default=False)  # For Ticket subsystem
is_talk_live = BooleanField(default=False)  # For Talk subsystem
is_video_live = BooleanField(default=False)  # For video subsystem
```

The `is_published` will be a property, computed from `published_at` (`is_published = (published_at is not None)`).
The `is_tickets_live` is boolean (two values) instead of 3-variant enum because I don't see the diference between "offline" and "test mode".
I wish I can use the name `is_on_sale` instead of `is_tickets_live`, but I don't know what name should be a replacement for the other two field.
I'm not sure if we need to track when the tickets / talk / video subsystems went live. If yes, we may follow the same pattern as `published_at`.

mariobehling:
@hongquan 

If an event is **not live**, it means it is still in draft status and **ticket sales are not yet possible** — no one can place orders or access the checkout. 

When an event is **live but in test mode**, the event is **publicly visible**, allowing organizers or testers to simulate the full purchasing process. However, all resulting orders are **test orders** that do not count as real sales and can be freely deleted once testing is complete. 

In contrast, once an event is **live and in normal mode**, all ticket sales are **real transactions** — they can no longer be deleted, only **canceled or refunded** according to the organizer’s policies.

hongquan:
Thanks. Then the tickets subsystem will have 3 states. Instead of `is_tickets_live`, we will use this field:

```py
class TicketingMode(TextChoices):
    OFFLINE = 'o', 'Offline'
    TEST = 't', 'Test'
    LIVE = 'l', 'Live'

ticketing_mode = CharField(max_length=1, choices=TicketingMode.choices, default=TicketingMode.OFFLINE)
```

`Event` model will also have `published_at`, `is_talk_live`, `is_video_live` fields as the above comment.

hpdang:
@Gagan-Ram any updates on this?

mariobehling (changes requested):
Please resolve conflicts and outstanding changes.

Sak1012:
Thank you, closing this PR as these changes will be part of a new redesign and re-implementation. Please take a look at new issues for further contributions.

PR #1172, opened 2025-11-01
chore(admin): add new command to create admin user

Gagan-Ram:
This PR adds a new command to create admin user.
From now on, we must drop using django's built in command `python manage.py createsuperuser`.
Instead, use `python manage.py createadminuser` to create admin user.

## Summary by Sourcery

Provide a custom createadminuser command and underlying UserManager helper to standardize and secure admin user creation without granting superuser privileges.

New Features:
- Add create_adminuser method to UserManager to provision admin users with is_staff and is_administrator flags
- Introduce createadminuser management command that prompts for email, password confirmation and optional full name, validates inputs, and creates admin users using the custom manager


hongquan (changes requested):
Please rename the command to `create_admin_user` (make it more readable).
After that, the PR is good to get merged.

Gagan-Ram:
> Please rename the command to `create_admin_user` (make it more readable). After that, the PR is good to get merged.

I thought of matching django's style i.e. `python manage.py createsuperuser`.
I have updated the file and also the setup docs too.

hongquan:
I think Django's built-in command has weird name scheme (like `createsuperuser` instead of `create_superuser`) because it doesn't want to clash with the commands provided by 3rd-party libs or user apps.

PR #1190, opened 2025-11-04
feat(email): implement an outbox for ticketing mails with multiple receivers

Gagan-Ram:
Fixes: #658

Video:

https://github.com/user-attachments/assets/7c7d15a0-39d0-40ef-aed5-1094d060ee86

## Summary by Sourcery

Implement an outbox system for ticketing emails to allow queuing, managing, and sending batch mails to attendees, orders, or team members.

New Features:
- Introduce EmailQueue, EmailQueueFilter, and EmailQueueToUser models to store and structure queued emails with recipient metadata
- Add outbox management views and templates (list, edit, delete, purge, sent) for queued mails
- Implement send_queued_mail Celery task with retry and error handling to process queued emails asynchronously
- Support team-based email composition through a dedicated TeamMailForm and associated views
- Provide a compose choice view to select between attendee/order emails and team emails

Enhancements:
- Refactor SenderView to queue emails into the outbox instead of sending immediately
- Add CopyDraftMixin to preload existing queued mails as drafts and QueryFilterOrderingMixin for searchable and orderable listings
- Extend the mail service to respect per-queue BCC and Reply-To overrides
- Update control navigation to include outbox and sent emails sections under Mails

### DIFF
--- a/app/eventyay/base/models/room.py
+++ b/app/eventyay/base/models/room.py
@@ -92,12 +92,15 @@ class RoomQuerySet(models.QuerySet):
         # When our user has traits, this is automatically ensured by the ALL() statement, but when traits=[] we
         # need to do a special case check since "IN ()" is not valid SQL
         for i, role in enumerate(roles):
-            if traits:
+            if traits and len(traits) > 0:
                 ext = ""
                 ext_args = []
                 if not allow_empty_traits:
                     ext = " AND jsonb_array_length(trait_grants->%s) > 0"
                     ext_args.append(role)
+
+                # Build IN clause with proper placeholders for each trait
+                in_placeholders = ','.join(['%s'] * len(traits))
 
                 qs = qs.annotate(
                     **{
@@ -108,8 +111,8 @@ class RoomQuerySet(models.QuerySet):
                             TRUE = ALL(
                                 SELECT (
                                     CASE jsonb_typeof(d{i}.elem)
-                                        WHEN 'array' THEN EXISTS(SELECT 1 FROM jsonb_array_elements(d{i}.elem) e{i}(elem) WHERE e{i}.elem#>>'{"{}"}' IN %s )
-                                        ELSE d{i}.elem#>>'{"{}"}' IN %s
+                                        WHEN 'array' THEN EXISTS(SELECT 1 FROM jsonb_array_elements(d{i}.elem) e{i}(elem) WHERE e{i}.elem#>>'{"{}"}' IN ({in_placeholders}) )
+                                        ELSE d{i}.elem#>>'{"{}"}' IN ({in_placeholders})
                                     END
                                 ) FROM jsonb_array_elements( trait_grants->%s ) AS d{i}(elem)
                             ) {ext}
@@ -117,8 +120,8 @@ class RoomQuerySet(models.QuerySet):
                             (
                                 role,  # ? check
                                 role,  # IS NOT NULL check
-                                tuple(traits),  # IN check
-                                tuple(traits),  # IN check
+                                *traits,  # IN check - expand traits as individual params
+                                *traits,  # IN check - expand traits as individual params
                                 role,  # jsonb_array_elements
                                 *ext_args,
                             ),
@@ -155,7 +158,7 @@ class Room(VersionedModel, OrderedModel, PretalxModel):
     The Room object stores some meta information. Most, like capacity,
     are not in use right now.
     """
-    
+
     log_prefix = "pretalx.room"
 
     deleted = models.BooleanField(default=False)
--- a/app/eventyay/multidomain/maindomain_urlconf.py
+++ b/app/eventyay/multidomain/maindomain_urlconf.py
@@ -70,7 +70,8 @@ class VideoSPAView(View):
         # Now expecting organizer and event from URL pattern: /{organizer}/{event}/video
         organizer_slug = kwargs.get('organizer')
         event_slug = kwargs.get('event')
-        
+        event_identifier = kwargs.get('event_identifier')
+
         # TODO remove debug logging once new video routing is stable
         event = None
         if organizer_slug and event_slug:
@@ -136,8 +137,6 @@ class VideoSPAView(View):
                 'locales': ['en', 'de', 'pt_BR', 'ar', 'fr', 'es', 'uk', 'ru'],
                 'noThemeEndpoint': True,  # Prevent frontend from requesting missing /theme endpoint
             }
-            import json as _json
-
             class EventyayJSONEncoder(DjangoJSONEncoder):
                 def default(self, obj):
                     if isinstance(obj, (Promise, LazyI18nString)):
@@ -148,14 +147,11 @@ class VideoSPAView(View):
         elif event_identifier:
             # Event identifier provided but not found -> 404
             return HttpResponse('Event not found', status=404)
-            serialized = _json.dumps(injected)
-            content = f"<script>window.eventyay={serialized};window.venueless={serialized};</script>{content}"
-            if '<base ' not in content.lower():
-                content = content.replace('<head>', f'<head><base href="{base_href}">', 1)
-        elif '<base ' not in content.lower():
-            # Legacy plain /video should still load SPA; ensure assets resolve correctly
+
+        if '<base ' not in content.lower():
+            # Ensure assets resolve correctly regardless of nested route
             content = content.replace('<head>', f'<head><base href="{base_href}">', 1)
-        
+
         resp = HttpResponse(content, content_type='text/html')
         resp._csp_ignore = True  # Disable CSP for SPA (relies on dynamic inline scripts)
         return resp
--- a/app/eventyay/plugins/eventyay-ticket-video/pretix_venueless/views.py
+++ b/app/eventyay/plugins/eventyay-ticket-video/pretix_venueless/views.py
@@ -1,7 +1,9 @@
 import datetime
 import hashlib
+import logging
 import random
 import string
+from urllib.parse import urlparse, urlunparse
 
 import jwt
 from django import forms
@@ -27,6 +29,8 @@ from eventyay.control.views.event import (
 )
 from eventyay.presale.views import EventViewMixin
 from eventyay.presale.views.order import OrderPositionDetailMixin
+
+logger = logging.getLogger(__name__)
 
 
 class VenuelessSettingsForm(SettingsForm):
@@ -178,6 +182,10 @@ class OrderPositionJoin(EventViewMixin, OrderPositionDetailMixin, View):
             "profile": profile,
             "traits": list(
                 {
+                    # Grant base attendee role so the video app allows EVENT_VIEW by default
+                    # Without this, users with valid tickets received auth.denied because
+                    # none of the event trait_grants matched the token traits.
+                    'attendee',
                     'eventyay-video-event-{}'.format(request.event.slug),
                     'eventyay-video-subevent-{}'.format(self.position.subevent_id),
                     'eventyay-video-product-{}'.format(self.position.product_id),
@@ -215,8 +223,29 @@ class OrderPositionJoin(EventViewMixin, OrderPositionDetailMixin, View):
 
         baseurl = self.request.event.settings.venueless_url
         if kwargs.get("view_schedule") == 'True':
-            return redirect(self.request.event.settings.venueless_talk_schedule_url)
+            redirect_url = self.request.event.settings.venueless_talk_schedule_url
+            logger.info('Redirecting to %s...', redirect_url)
+            return redirect(redirect_url)
         if '{token}' in baseurl:
             # Hidden feature to support other kinds of installations
-            return redirect(baseurl.format(token=token))
-        return redirect('{}/#token={}'.format(baseurl, token).replace("//#", "/#"))
+            redirect_url = baseurl.format(token=token)
+            logger.info('Redirecting to %s...', redirect_url)
+            return redirect(redirect_url)
+
+        # Ensure the URL includes the event identifier so VideoSPAView has event context
+        # Format: http://localhost:8000/organizer-slug/event-slug/video/#token=...
+        # Use Django's reverse() to properly construct the video URL path
+        video_path = reverse('video.spa', kwargs={
+            'organizer': self.request.event.organizer.slug,
+            'event': self.request.event.slug
+        })
+
+        # Parse the base URL to get scheme and netloc (domain)
+        parsed = urlparse(baseurl)
+
+        # Reconstruct the full URL with the proper path from reverse()
+        baseurl = urlunparse((parsed.scheme, parsed.netloc, video_path, '', '', ''))
+
+        redirect_url = '{}/#token={}'.format(baseurl, token).replace("//#", "/#")
+        logger.info('Redirecting to %s...', redirect_url)
+        return redirect(redirect_url)
--- a/app/eventyay/presale/views/event.py
+++ b/app/eventyay/presale/views/event.py
@@ -7,6 +7,7 @@ from collections import defaultdict
 from datetime import date, datetime, timedelta
 from decimal import Decimal
 from importlib import import_module
+from urllib.parse import urlparse, urlunparse
 
 import isoweek
 import jwt
@@ -25,6 +26,7 @@ from django.db.models import (
 )
 from django.http import Http404, HttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404, redirect, render
+from django.urls import reverse
 from django.utils.decorators import method_decorator
 from django.utils.formats import get_format
 from django.utils.timezone import now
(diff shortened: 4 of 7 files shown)


mariobehling:
> * Detected a segment of a Flask template where autoescaping is explicitly disabled with '| safe' filter. This allows rendering of raw HTML in this segment. Ensure no user data is rendered here, otherwise this is a cross-site scripting (XSS) vulnerability. ([link](https://github.com/fossasia/eventyay/pull/1190/files#diff-a9ecf9d69ff5f2638800a47dfeade93ea0f252a105340a42f3b22d4590f9d2adR133))
> * Detected a segment of a Flask template where autoescaping is explicitly disabled with '| safe' filter. This allows rendering of raw HTML in this segment. Ensure no user data is rendered here, otherwise this is a cross-site scripting (XSS) vulnerability. ([link](https://github.com/fossasia/eventyay/pull/1190/files#diff-a9ecf9d69ff5f2638800a47dfeade93ea0f252a105340a42f3b22d4590f9d2adR134))
> * Detected a segment of a Flask template where autoescaping is explicitly disabled with '| safe' filter. This allows rendering of raw HTML in this segment. Ensure no user data is rendered here, otherwise this is a cross-site scripting (XSS) vulnerability. ([link](https://github.com/fossasia/eventyay/pull/1190/files#diff-8bd23b9a10bacc4428e6b08c8e77508d905fc811aaa9806172f565decd4aea36R45))
> * Detected a segment of a Flask template where autoescaping is explicitly disabled with '| safe' filter. This allows rendering of raw HTML in this segment. Ensure no user data is rendered here, otherwise this is a cross-site scripting (XSS) vulnerability. ([link](https://github.com/fossasia/eventyay/pull/1190/files#diff-8bd23b9a10bacc4428e6b08c8e77508d905fc811aaa9806172f565decd4aea36R46))

@sourcery-ai

Thanks for the heads-up. Two clarifications and actions from our side:

1. **Framework misclassification**
   This is a **Django** project, not Flask. The `|safe` filter here is the **Django template filter** (not Jinja/Flask). The rule is still useful though, because it flags potential XSS whenever auto-escaping is bypassed.

2. **Why `|safe` appears here & how we mitigate XSS**
   The flagged lines render **HTML previews** (e.g., email body preview and helper snippets). These can contain markup by design. To prevent XSS, we will/do ensure only **sanitized HTML** is passed to templates:

* ✔️ **Server-side sanitization** before render:

  * We sanitize email body/help HTML using an allow-list (e.g. `bleach.clean(..., tags=ALLOWED_TAGS, attributes=ALLOWED_ATTRS, protocols=ALLOWED_PROTOCOLS, strip=True)`), removing `<script>`, `on*`-Handler etc.
  * Any placeholder values originating from user input are escaped during placeholder rendering; only system-generated/trusted fragments can be marked as safe.
* ✔️ **Template hardening adjustments** (follow-up patch):

  * **Subjects/titles** will **not** use `|safe` (they should be plain text); we’ll remove `|safe` there.
  * For the **HTML body preview**, we will render a distinct context var like `body_html_trusted` that is only set after sanitization; only that var is used with `|safe`.
  * We will add an inline template comment explaining why `|safe` is safe at that exact line.
* ✔️ **Tests**:

  * Add unit tests with payloads like `` and `<a href="javascript:...">` to assert they are neutralized in previews.
  * Add a regression test ensuring placeholders that include user data are escaped unless explicitly whitelisted.

**Net result:** The “Flask” label is a mis-ID, but the underlying XSS concern is addressed. We’ll keep `|safe` **only** for sanitized, trusted HTML (email **body preview**), and remove it where not needed (e.g., **subject**).

How to scope your rule to Django (or paths) to avoid Flask-specific wording going forward?

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,399 @@
...
+
+        recipient.save(update_fields=["sent", "error"])
+        return True
+
+    def get_recipient_emails(self):
+        """
+        Resolve and return the full list of unique email addresses
+        this queued mail will send to.
+        """
+        return sorted(set(r.email.strip().lower() for r in self.recipients.all() if r.email))
```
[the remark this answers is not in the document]
Gagan-Ram: No, the set() ensures recipients emails are not duplicated.

app/eventyay/plugins/sendmail/forms.py
```diff
@@ -411,3 +419,173 @@ def __init__(self, *args, **kwargs):
...
+                _("You cannot add new recipients. Only editing existing email addresses is allowed.")
+            )
+        return updated_emails
+
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+
+        updated_emails = self.cleaned_data['emails']
+
+        for i, email in enumerate(updated_emails):
```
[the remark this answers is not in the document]
Gagan-Ram: Thanks

mariobehling (changes requested):
Thanks, please add screenshots or a video.

app/eventyay/plugins/sendmail/tasks.py
```diff
@@ -110,3 +114,43 @@ def send_mails(
...
+def send_queued_mail(self, event_id: int, queued_mail_id: int):
+    from eventyay.plugins.sendmail.models import EmailQueue
+    from celery.exceptions import MaxRetriesExceededError
+
+    if isinstance(event_id, Event):
+        original_event_id = event_id.pk
+        event = event_id
+    else:
+        original_event_id = event_id
+        event = Event.objects.get(id=original_event_id)
```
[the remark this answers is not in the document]
mariobehling: Please also check this.

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,396 @@
...
+    
+    :param sent: Whether this recipient has been successfully sent the email.
+    :type sent: bool
+    
+    :param error: Error message if sending failed.
+    :type error: str or None
+    """
+    mail = models.ForeignKey(EmailQueue, on_delete=models.CASCADE, related_name="recipients")
+    email = models.EmailField()
+    orders = ArrayField(models.CharField(max_length=64), blank=True, default=list)
```
hongquan: Here is your usage:

```py
recipients[email]["orders"].add(str(order.pk))
```

Then why make this field `ArrayField(CharField())`? With that usage (storing list of primary key), it should be

```py
ArrayField(BigIntegerField())
```

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,396 @@
...
+    :type orders: list[int]
+    
+    :param teams: Team IDs to include (for team-based emails).
+    :type teams: list[int]
+    """
+    mail = models.OneToOneField(EmailQueue, on_delete=models.CASCADE, related_name="filters_data")
+
+    recipients = models.CharField(
+        max_length=10,
+        choices=[("orders", "Orders"), ("attendees", "Attendees"), ("both", "Both")],
```
hongquan: These choices should be defined as `enum` or `TextChoices`, so that it can be imported and used elsewhere.

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,396 @@
...
+        default="orders",
+        blank=True
+    )
+    sendto = ArrayField(models.CharField(max_length=20), blank=True, default=list)
```
hongquan: The field name should be `send_to`. The field type should be `ArrayField(EmailField())`.

app/eventyay/plugins/sendmail/tasks.py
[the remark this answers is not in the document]
Gagan-Ram: this has been resolved

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,396 @@
...
+    products = ArrayField(models.IntegerField(), blank=True, default=list)
```
hongquan: Should be `ArrayField(BigIntegerField(())`, because you save primary keys to it.

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,396 @@
...
+    checkin_lists = ArrayField(models.IntegerField(), blank=True, default=list)
+    filter_checkins = models.BooleanField(default=False)
```
hongquan: The field name doesn't sound like it contains boolean values.

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,396 @@
...
+    not_checked_in = models.BooleanField(default=False)
+    subevent = models.IntegerField(null=True, blank=True)
+    subevents_from = models.DateTimeField(null=True, blank=True)
+    subevents_to = models.DateTimeField(null=True, blank=True)
+    created_from = models.DateTimeField(null=True, blank=True)
```
hongquan: The field name sound like the creation date of this model instance, but the doc string says not. You should make a better name to prevent confusion.

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,396 @@
...
+    :type sent_at: datetime or None
+    """
+    event = models.ForeignKey(Event, on_delete=models.CASCADE, related_name="email_queue")
+    user = models.ForeignKey(User, null=True, blank=True, on_delete=models.SET_NULL)
+    composing_for = models.CharField(max_length=20, choices=ComposingFor.choices, default=ComposingFor.ATTENDEES)
+
+    subject = I18nTextField(null=True, blank=True)
+    message = I18nTextField(null=True, blank=True)
+
+    reply_to = models.CharField(max_length=100, null=True, blank=True)
```
hongquan: For `CharField`, you should not use `null=True`, use empty string for the case of missing data. This is Django [recommendation](https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.Field.null). 

It is also from my experience, that the `CharField(null=True)` will return value of type `str | None`. This type is annoying because before you want to use any string method (like `.lower()`, `.strip()`), you have do type check to exclude the case of `None`.

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,398 @@
...
+    :param orders: Explicit order IDs to include.
+    :type orders: list[int]
+    
+    :param teams: Team IDs to include (for team-based emails).
+    :type teams: list[int]
+    """
+    mail = models.OneToOneField(EmailQueue, on_delete=models.CASCADE, related_name="filters_data")
+
+    recipients = models.CharField(max_length=10, choices=Recipients.choices, default=Recipients.ORDERS, blank=True)
+    send_to = ArrayField(models.EmailField(), blank=True, default=list)
```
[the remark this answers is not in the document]
Gagan-Ram: Thanks a lot

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,398 @@
...
+    products = ArrayField(models.BigIntegerField(), blank=True, default=list)
+    checkin_lists = ArrayField(models.IntegerField(), blank=True, default=list)
+    has_filter_checkins = models.BooleanField(default=False)
+    not_checked_in = models.BooleanField(default=False)
+    subevent = models.IntegerField(null=True, blank=True)
+    subevents_from = models.DateTimeField(null=True, blank=True)
+    subevents_to = models.DateTimeField(null=True, blank=True)
+    order_created_from = models.DateTimeField(null=True, blank=True)
+    order_created_to = models.DateTimeField(null=True, blank=True)
```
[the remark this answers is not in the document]
Gagan-Ram: It must be DateTimeField.
Updated views.py
Thank you.

app/eventyay/plugins/sendmail/models.py
Gagan-Ram: The field name is confusing here.
I have renamed it to order_status, which stores list of order status values, which is shown later as checkbox fields in forms

Gagan-Ram:
I have tested this PR against multiple Localization too. This works as expected.

Gagan-Ram:
> Thanks, please add screenshots or a video.

Added in description.
Thanks

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,398 @@
...
+    user = models.ForeignKey(User, null=True, blank=True, on_delete=models.SET_NULL)
+    composing_for = models.CharField(max_length=20, choices=ComposingFor.choices, default=ComposingFor.ATTENDEES)
+
+    subject = I18nTextField(null=True, blank=True)
+    message = I18nTextField(null=True, blank=True)
+
+    reply_to = models.CharField(max_length=100, default='', blank=True)
+    bcc = models.TextField(null=True, blank=True)  # comma-separated
+    locale = models.CharField(max_length=16, blank=True, default='')
+    attachments = ArrayField(base_field=models.UUIDField(), null=True, blank=True)
```
hongquan: Those `ArrayField` should be non-null, with an empty list as default. By doing so, the return value will be of `list` type, instead of `list | None`. We don't have to check type before calling list methods.

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,398 @@
...
+                user=self.user,
+                auto_email=False,
+            )
+            recipient.sent = True
+            recipient.error = None
+        except SendMailException as se:
+            recipient.sent = False
+            recipient.error = str(se)
+            recipient.save(update_fields=["sent", "error"])
+            logger.exception("SendMailException error while sending to %s: %s", email, str(se))
```
hongquan: You are using `logger.exception`. You don't need to pass `se` to the logger, because `logger.exception` always capture the error.

app/eventyay/plugins/sendmail/templatetags/sendmail_extras.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+
+register = template.Library()
+
+@register.filter
+def split(value, delimiter=','):
```
hongquan, line 6: This is the disadvantage of using Django template instead of Jinja.
With Jinja, you don't need to define this filter. In Jinja template, you just use normal Python code:

```nunjucks
{{ values.split(',') }}
```

app/eventyay/plugins/sendmail/templates/pretixplugins/sendmail/outbox_list.html
```diff
@@ -0,0 +1,164 @@
...
+</div>
+
+<table class="table mt-4">
+    <thead>
+        <tr>
+            {% for field, label in headers %}
+                <th>
+                    {{ label }}
+                    {% with asc=field desc="-"|add:field %}
+                        <a href="?ordering={{ asc }}{% if query %}&q={{ query }}{% endif %}"
```
hongquan: It is a bad practice to compose URL query string in template. You should do it in Python code, processing it with [`urllencode`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode), before passing down to template, because some characters are illegal in URL.

hongquan (changes requested):

app/eventyay/plugins/sendmail/templatetags/sendmail_extras.py
Gagan-Ram: ok, but I will retain .html files for now.
when we shift to jinja, we'll make changes

app/eventyay/plugins/sendmail/templates/pretixplugins/sendmail/outbox_list.html
```diff
@@ -57,12 +57,12 @@ <h1>
...
                     {% with asc=field desc="-"|add:field %}
-                        <a href="?ordering={{ asc }}{% if query %}&q={{ query }}{% endif %}"
-                           class="{% if current_ordering == asc %}fw-bold{% else %}text-muted{% endif %}">
+                        <a href="?{% if encoded_query %}{{ encoded_query }}&{% endif %}ordering={{ asc }}"
+                        class="{% if current_ordering == asc %}fw-bold{% else %}text-muted{% endif %}">
                             &#9650;
                         </a>
-                        <a href="?ordering={{ desc }}{% if query %}&q={{ query }}{% endif %}"
-                           class="{% if current_ordering == desc %}fw-bold{% else %}text-muted{% endif %}">
+                        <a href="?{% if encoded_query %}{{ encoded_query }}&{% endif %}ordering={{ desc }}"
```
hongquan: Any reason why the `ordering` parameter is not inside `encoded_query`?

Gagan-Ram: When generating the sorting arrows (the ▲/▼ links), you are creating new ordering options.
If encoded_query already included the current ordering parameter, clicking one of those arrows would produce a URL like this:
```
?q=foo&ordering=-subject&ordering=subject
```
That would lead to duplicate ordering parameters, and Django would only use the first one, effectively ignoring the new sort you just clicked.

hongquan: It means that the "the ▲/▼ links" are dumb, they should overwrite the _ordering_ parameter, instead of appending. Those links should be fixed.

app/eventyay/plugins/sendmail/tasks.py
```diff
@@ -1,112 +1,45 @@
...
+    from celery.exceptions import MaxRetriesExceededError
 
-                if p.product_id not in products and not any(a.product_id in products for a in p.addons.all()):
-                    continue
+    if isinstance(event_id, Event):
+        original_event_id = event_id.pk
+        event = event_id
+    else:
+        original_event_id = event_id
+        event = Event.objects.get(id=original_event_id)
```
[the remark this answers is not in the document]
Gagan-Ram: @hongquan any comments

app/eventyay/plugins/sendmail/models.py
```diff
@@ -0,0 +1,398 @@
...
+            else:
+                return get_email_context(event=self.event)
+        except Exception as e:
+            logger.exception("Error while generating email context")
+            recipient.error = f"Context error: {str(e)}"
+            recipient.save(update_fields=["error"])
+            return None
+
+    def _finalize_send_status(self):
+        self.sent_at = now() if all(r.sent for r in self.recipients.all()) else None
```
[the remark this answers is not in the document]
Gagan-Ram: if mails are failed to send, then we will show unsent emails in frontend, so that organizer can try to resend the mail to failed ones,

app/eventyay/plugins/sendmail/views.py
```diff
@@ -125,19 +108,19 @@ def form_valid(self, form):
...
 
         orders = orders.annotate(match_pos=Exists(opq)).filter(match_pos=True).distinct()
 
-        self.output = {}
         if not orders:
             messages.error(self.request, _('There are no orders matching this selection.'))
             return self.get(self.request, *self.args, **self.kwargs)
 
         if self.request.POST.get('action') == 'preview':
+            self.output = {}
```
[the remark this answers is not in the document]
Gagan-Ram: we do use `ctx['output'] = getattr(self, 'output', None)` in get_context_data

hongquan (approved):

Issue #1214, opened 2025-11-09
SSO Username Mapping and Original Username Retention for Wikimedia Login Integration


mariobehling:
Wikimedia requires enhanced SSO handling and username mapping to improve participant identification and Trust & Safety workflows during event registrations on **Eventyay**.

The requested functionality primarily concerns how usernames from SSO (e.g., Wikimedia OAuth) are **pre-filled, displayed, and stored** for **audit and vetting purposes**.

This ensures that Wikimedia usernames are applied consistently across public-facing sections of the platform, while organizers have control over whether this behavior is active through event settings.

### Requirements

**1. Organizer-Controlled Setting**

* Introduce a new toggle in **Event Settings → General / SSO Settings**:
  **“Use Wikimedia username as the value for the public Name field”**
* When enabled:

  * Users who log in through Wikimedia SSO will automatically have their Wikimedia username applied as the value for the **public “Name” field**.
  * This applies across all visible contexts — user profiles, registration forms, badges, and public attendee lists.
* When disabled:

  * The “Name” field remains user-editable and behaves as standard, without pre-filling from SSO.
* The toggle works per event, giving Wikimedia events full control while preserving default behavior for others.

**2. Username Pre-Fill and Display Mapping**

* Upon login via Wikimedia SSO:

  * The **Wikimedia username** should pre-populate the “Name” field on the registration form.
  * The user may still edit the field (unless the organizer locks it in a later enhancement).
  * The name value should propagate consistently across all parts of the platform that display the participant’s name (profiles, badges, talk listings, and attendee lists).

**3. Persistent Storage of Original SSO Username**

* Add a permanent, non-editable field (e.g., `original_sso_username`) to store the username received from Wikimedia OAuth.
* This field remains tied to the user account even if the user changes their display or public name.
* Visible only to admins and organizers for **Trust & Safety review**; never shown to public users or included in open API responses.

**4. Admin Access for Trust & Safety**

* Add an **admin-only view** (or user detail field) displaying the original Wikimedia SSO username.
* Optionally include it in secure exports for event vetting or moderation.

**5. Data Integrity and Privacy**

* The `original_sso_username` must not be editable by the user.
* SSO re-logins must preserve the stored username (no overwrite).
* Maintain backward compatibility for all non-SSO accounts.

### Possible Implementation Notes

* Extend the Wikimedia OAuth login handler to store the `username` claim as `original_sso_username`.
[...]

PR #1230, opened 2025-11-11
Register Social Auth plugin URLs in main routing

### LINKED ISSUE #1226: Recover Social login (OAuth) for enext
This feature is lost: The URL for initiating login flow and the callback URL are not working.

ArnavBallinCode:
Fixes #1226 

## Summary  
To streamline our authentication pipeline and unlock frictionless OAuth workflows, this update integrates the Social Auth plugin URL patterns directly into the core routing architecture. By baking these routes into the primary URL configuration, all OAuth endpoints become first-class citizens in the application’s request-resolution lifecycle. Previously, `/oauth_login/<provider>/` and `/oauth_return/` generated 404s because the plugin endpoints were not present in the main dispatcher.

## What I Changed  
- Registered `socialauth_patterns` inside `urls.py`  
- Included `eventyay.plugins.socialauth.urls` via the plugin loader  
- Inserted `socialauth_patterns` into `common_patterns`, ensuring these routes are resolved early in the core routing pipeline  

## Why  
The Social Auth plugin already ships with fully implemented view logic:
- `OAuthLoginView`
- `OAuthReturnView`
- `SocialLoginView`

However, none of these were wired into Django’s top-level URL dispatcher. As a result, OAuth flows were effectively broken. 

## Related Issues  
- **Fixes:** #1226  
- **Related to:** Wikimedia SSO / username mapping workflow (#1214)

## Summary by Sourcery

Bug Fixes:
- Restore OAuth login flows by registering Social Auth plugin URL patterns in the main routing configuration

### DIFF
--- a/app/eventyay/config/urls.py
+++ b/app/eventyay/config/urls.py
@@ -1,6 +1,5 @@
 import importlib.util
 
-from django.apps import apps
 from django.conf import settings
 from django.conf.urls.static import static
 from django.urls import include, path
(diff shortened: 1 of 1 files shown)


hongquan:
Please don't use "Fix/xxx" pattern for the PR title.

hongquan:
Note that, this PR brings back the URL for initiating OAuth flow, but it doesn't resolve the "callback URL":

image

app/eventyay/config/urls.py
```diff
@@ -60,6 +61,22 @@
 # Note: agenda and cfp patterns are now included under {organizer}/{event} in maindomain_urlconf.py
 # They are no longer at the root level
 
+# Auto-discover and load plugin URLs (similar to maindomain_urlconf.py)
```
hongquan: Your previous implementation is better. Because when we debug, we see a URL, we will want to trace back to the module which generates that URL.
A static registration like previous implementation helps us trace back easier (by searching for the string in the code base).

With this implementation, it is very hard to do so (there is no string to search).

hongquan:
Btw, don't force-push your commits, because sometimes we need to refer to previous commits. By doing force-push, we have nothing to refer to.

hongquan:
Thanks, I can login successfully with this fix.

image

Though there is error at the last step (but it can be fixed in the next PR):

image

hongquan:
Note that, the callback URL to register with MediaWiki is in this form: *http://localhost:8000/accounts/mediawiki/login/callback/*, not the */oauth_return/*.

hongquan:
Hi @ArnavBallinCode
To let you know, the changes made by this PR is incompatible with other places in this code base. We temporarily revert this PR at PR #1241 until we find a better solution.

PR #1258, opened 2025-11-13
fix: unable to export order data – HTTP 500 error

### LINKED ISSUE #1250: Ticket Component: Unable to Export Order Data – HTTP 500 Error
Go to Orders -> Export -> HTTP 500 Error https://next.eventyay.com/control/event/a-team/pgday/orders/export/

ArnavBallinCode:
Fixes  #1250

# (fix) unable to Export Order Data – HTTP 500 Error
located in the `ExportMixin` class in `/app/eventyay/control/views/orders.py` at line 2503.

```python
test_form = ExporterForm(data=self.request.GET, prefix=ex.identifier)
test_form.fields = ex.export_form_fields
test_form.is_valid()
initial = {k: v for k, v in test_form.cleaned_data.items() if ex.identifier + '-' + k in self.request.GET}
```

The code calls `test_form.is_valid()` on line 2502, but **doesn't check the result** , on line 2503, it tries to access `test_form.cleaned_data` unconditionally, if `test_form.is_valid()` returns `False`, `cleaned_data` may not exist or may be incomplete.
Usually when a form is invalid, `cleaned_data` only contains fields that passed individual field validation....

If the exporter has certain form fields that fail validation (or if there are no GET parameters), accessing `cleaned_data` would raise an `AttributeError` or something of such sort...

### Code Changes
```python
# AFTER (Fixed code):
test_form = ExporterForm(data=self.request.GET, prefix=ex.identifier)
test_form.fields = ex.export_form_fields
if test_form.is_valid():  #  Check result
    initial = {k: v for k, v in test_form.cleaned_data.items() if ...}
else:
    initial = {}  # Safe fallback
```

1. **Explicit validation check**: Only accesses `cleaned_data` when form is valid
2. **Safe fallback**: Uses empty dict when form is invalid
3. **Preserves functionality**: Valid GET parameters still work as before
4. **Prevents errors**: No more AttributeError or unexpected behavior

## Summary by Sourcery

Add explicit form validation check in exporters views to prevent HTTP 500 errors when exporting data and rename the 'id' variable to 'identifier' for clarity.

Bug Fixes:
- Check ExporterForm.is_valid() before accessing cleaned_data to prevent AttributeError on invalid or missing GET parameters
- Provide empty initial fallback when form validation fails

Enhancements:
- Rename variable 'id' to 'identifier' to avoid shadowing built-in and improve clarity

Screenshot 2025-11-14 at 1 00 38 PM

## Summary by Sourcery

Prevent export errors by safely accessing form.cleaned_data, rename "items" field to "products" in check-in list exporters, and add a JSON_FIELD_AVAILABLE setting flag

Bug Fixes:
- Safely access ExporterForm.cleaned_data with getattr to avoid HTTP 500 when exporting orders and organizers

Enhancements:
- Rename check-in list exporter field and filter parameter from 'items' to 'products'
- Add JSON_FIELD_AVAILABLE setting flag to detect PostgreSQL JSON field support

### DIFF
--- a/app/eventyay/base/exporters/json.py
+++ b/app/eventyay/base/exporters/json.py
@@ -31,30 +31,30 @@ class JSONExporter(BaseExporter):
                     }
                     for category in self.event.categories.all()
                 ],
-                'items': [
+                'products': [
                     {
-                        'id': item.id,
-                        'name': str(item.name),
-                        'internal_name': str(item.internal_name),
-                        'category': item.category_id,
-                        'price': item.default_price,
-                        'tax_rate': item.tax_rule.rate if item.tax_rule else Decimal('0.00'),
-                        'tax_name': str(item.tax_rule.name) if item.tax_rule else None,
-                        'admission': item.admission,
-                        'active': item.active,
+                        'id': product.id,
+                        'name': str(product.name),
+                        'internal_name': str(product.internal_name),
+                        'category': product.category_id,
+                        'price': product.default_price,
+                        'tax_rate': product.tax_rule.rate if product.tax_rule else Decimal('0.00'),
+                        'tax_name': str(product.tax_rule.name) if product.tax_rule else None,
+                        'admission': product.admission,
+                        'active': product.active,
                         'variations': [
                             {
                                 'id': variation.id,
                                 'active': variation.active,
                                 'price': variation.default_price
                                 if variation.default_price is not None
-                                else item.default_price,
+                                else product.default_price,
                                 'name': str(variation),
                             }
-                            for variation in item.variations.all()
+                            for variation in product.variations.all()
                         ],
                     }
-                    for item in self.event.items.select_related('tax_rule').prefetch_related('variations')
+                    for product in self.event.products.select_related('tax_rule').prefetch_related('variations')
                 ],
                 'questions': [
                     {
@@ -83,7 +83,7 @@ class JSONExporter(BaseExporter):
                         'positions': [
                             {
                                 'id': position.id,
-                                'item': position.item_id,
+                                'product': position.product_id,
                                 'variation': position.variation_id,
                                 'price': position.price,
                                 'attendee_name': position.attendee_name,
@@ -107,10 +107,10 @@ class JSONExporter(BaseExporter):
                     {
                         'id': quota.id,
                         'size': quota.size,
-                        'items': [item.id for item in quota.items.all()],
+                        'products': [product.id for product in quota.products.all()],
                         'variations': [variation.id for variation in quota.variations.all()],
                     }
-                    for quota in self.event.quotas.all().prefetch_related('items', 'variations')
+                    for quota in self.event.quotas.all().prefetch_related('products', 'variations')
                 ],
             }
         }
--- a/app/eventyay/base/forms/user.py
+++ b/app/eventyay/base/forms/user.py
@@ -54,16 +54,13 @@ class UserSettingsForm(forms.ModelForm):
     class Meta:
         model = User
         fields = ['fullname', 'wikimedia_username', 'locale', 'timezone', 'email']
-        labels = {
-            'wikimedia_username': 'Nick name',
-        }
         widgets = {'locale': SingleLanguageWidget}
 
     def __init__(self, *args, **kwargs):
         self.user = kwargs.pop('user')
         super().__init__(*args, **kwargs)
         self.fields['email'].required = True
-        self.fields['wikimedia_username'].widget.attrs['readonly'] = True
+        self.fields['wikimedia_username'].disabled = True
         if self.user.auth_backend != 'native':
             del self.fields['old_pw']
             del self.fields['new_pw']
--- a/app/eventyay/base/models/tax.py
+++ b/app/eventyay/base/models/tax.py
@@ -173,7 +173,7 @@ class TaxRule(LoggedModel):
         return (
             not OrderFee.objects.filter(tax_rule=self, order__event=self.event).exists()
             and not OrderPosition.all.filter(tax_rule=self, order__event=self.event).exists()
-            and not self.event.items.filter(tax_rule=self).exists()
+            and not self.event.products.filter(tax_rule=self).exists()
             and self.event.settings.tax_rate_default != self
         )
 
--- a/app/eventyay/base/services/cancelevent.py
+++ b/app/eventyay/base/services/cancelevent.py
@@ -210,11 +210,11 @@ def cancel_event(
             user=user,
         )
 
-        for i in event.items.filter(active=True):
+        for i in event.products.filter(active=True):
             i.active = False
             i.save(update_fields=['active'])
             i.log_action(
-                'pretix.event.item.changed',
+                'pretix.event.product.changed',
                 user=user,
                 data={'active': False, '_source': 'cancel_event'},
             )
--- a/app/eventyay/base/services/event.py
+++ b/app/eventyay/base/services/event.py
@@ -257,7 +257,10 @@ def _create_room(data, with_channel=False, permission_preset="public", creator=N
         data["trait_grants"] = {}
     has_modules = bool(data.get("module_config"))
     data.setdefault("setup_complete", has_modules)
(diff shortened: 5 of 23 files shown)


ArnavBallinCode:
@sourcery-ai review

app/eventyay/control/views/orders.py
```diff
@@ -2499,8 +2499,10 @@ def exporters(self) -> list[BaseExporter]:
             # Use form parse cycle to generate useful defaults
             test_form = ExporterForm(data=self.request.GET, prefix=ex.identifier)
             test_form.fields = ex.export_form_fields
-            test_form.is_valid()
-            initial = {k: v for k, v in test_form.cleaned_data.items() if ex.identifier + '-' + k in self.request.GET}
+            if test_form.is_valid():
+                initial = {k: v for k, v in test_form.cleaned_data.items() if f'{ex.identifier}-{k}' in self.request.GET}
+            else:
+                initial = {}
```
Sak1012, line 2504: Unnecessary conditional `is_valid()` is called unconditionally to populate `cleaned_data`, even if the form isn’t fully valid. This change drops partial defaults that are still useful for exporter initialization.

app/eventyay/control/views/organizer.py
```diff
@@ -46,19 +46,21 @@ def exporters(self):
...
         )
         responses = register_multievent_data_exporters.send(self.request.organizer)
-        id = self.request.GET.get('identifier') or self.request.POST.get('exporter')
+        identifier = self.request.GET.get('identifier') or self.request.POST.get('exporter')
         for ex in sorted(
             [response(events) for r, response in responses if response],
             key=lambda ex: str(ex.verbose_name),
         ):
-            if id and ex.identifier != id:
+            if identifier and ex.identifier != identifier:
```
Sak1012: The variable `id` here is local and harmless renaming adds no functional value and touches unrelated lines.

app/eventyay/control/views/organizer.py
```diff
@@ -46,19 +46,21 @@ def exporters(self):
...
 
             # Use form parse cycle to generate useful defaults
             test_form = ExporterForm(data=self.request.GET, prefix=ex.identifier)
             test_form.fields = ex.export_form_fields
-            test_form.is_valid()
-            initial = {k: v for k, v in test_form.cleaned_data.items() if ex.identifier + '-' + k in self.request.GET}
+            if test_form.is_valid():
+                initial = {k: v for k, v in test_form.cleaned_data.items() if f'{ex.identifier}-{k}' in self.request.GET}
+            else:
+                initial = {}
```
Sak1012, line 63: Same as above this logic should not depend on validation success. We still want to extract cleaned values even when the form is partially invalid.

app/eventyay/plugins/checkinlists/exporters.py
```diff
@@ -749,10 +749,10 @@ def _fields(self):
                 (
                     'items',
```
Sak1012: Please also rename to `products` for consistency.

Sak1012 (changes requested):
Thank you for the PR. While it addresses a real issue, it doesn’t quite resolve the underlying problem, as the export functionality still isn’t working as intended, with only 2 or 3 export options functioning.

The root cause seems to come from logic inside other `exporters.py` files elsewhere in the codebase, not from the view changes here. Those modules need to be updated in coordination for the export process to function correctly.

Please open a new issue describing the problem, so we can properly track it and address the related fixes in the appropriate `exporters.py` files. This PR can then be linked to that issue for context.

app/eventyay/plugins/checkinlists/exporters.py
ArnavBallinCode: Yup, done

ArnavBallinCode:
@sourcery-ai review

app/eventyay/control/views/orders.py
ArnavBallinCode: I've updated the code to use `getattr(test_form, 'cleaned_data', {})` instead of the if/else block.

app/eventyay/control/views/organizer.py
ArnavBallinCode: yup... reverted the variable rename from 'identifier' back to 'id'.

app/eventyay/control/views/organizer.py
ArnavBallinCode: Fixed with the same getattr() approach as in orders.py, partial cleaned data is now preserved.

app/eventyay/control/views/orders.py
```diff
@@ -2499,8 +2499,8 @@ def exporters(self) -> list[BaseExporter]:
             # Use form parse cycle to generate useful defaults
             test_form = ExporterForm(data=self.request.GET, prefix=ex.identifier)
             test_form.fields = ex.export_form_fields
-            test_form.is_valid()
-            initial = {k: v for k, v in test_form.cleaned_data.items() if ex.identifier + '-' + k in self.request.GET}
+            test_form.is_valid()  # Populate cleaned_data with valid fields
```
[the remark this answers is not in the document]
ArnavBallinCode: code already has this oof..

app/eventyay/config/settings.py
```diff
@@ -380,6 +380,9 @@ def instance_name(request):
     }
 }
 
+# JSON field support is available in PostgreSQL
+JSON_FIELD_AVAILABLE = DATABASES['default']['ENGINE'].split('.')[-1] == 'postgresql'
```
hongquan: We only supports PostgreSQL, meaning `JSONField` is always available. Don't add and don't depend on this setting.

app/eventyay/control/views/orders.py
```diff
@@ -2500,8 +2500,8 @@ def exporters(self) -> list[BaseExporter]:
             # Use form parse cycle to generate useful defaults
             test_form = ExporterForm(data=self.request.GET, prefix=ex.identifier)
             test_form.fields = ex.export_form_fields
-            test_form.is_valid()
-            initial = {k: v for k, v in test_form.cleaned_data.items() if ex.identifier + '-' + k in self.request.GET}
+            test_form.is_valid()  # Attempt to populate cleaned_data; use getattr for safe access as validation result is not checked
```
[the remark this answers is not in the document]
hongquan: This comment sounds like AI-generated, not matching the code at all. Delete it.

ArnavBallinCode:
Hey, just to clarify, those changes were actually suggested by Copilot during its automatic review. 
I should’ve checked them more carefully before committing, since they seemed harmless and everything still worked. My mistake there , cleaning them now..

app/eventyay/config/settings.py
```diff
@@ -811,7 +811,6 @@ def instance_name(request):
 
 REDIS_URL = config.get('redis', 'location')
 HAS_REDIS = bool(REDIS_URL)
-JSON_FIELD_AVAILABLE = True
```
Sak1012, line 814: `JSON_FIELD_AVAILABLE` this field is required for the export feature please add it back if not it'll throw a 500 error

Image

hongquan: It is better that the export doesn't depend on this `JSON_FIELD_AVAILABLE` settings, because we only support Postgres, and with Postgres, `JSONField` is always available, no need to handle the case of `JSON_FIELD_AVAILABLE is False`.

hongquan:
It is better that the export doesn't depend on `JSON_FIELD_AVAILABLE` settings, because we only support Postgres, and with Postgres, JSONField is always available, no need to handle the case of `JSON_FIELD_AVAILABLE is False`.

So please scan in codebase, where `JSON_FIELD_AVAILABLE` is used, and just adjust code to not care about `JSON_FIELD_AVAILABLE`.

ArnavBallinCode:
@Sak1012 @hongquan 

Since eventyay only supports PostgreSQL, I've removed the `JSON_FIELD_AVAILABLE` checks from the three exporter files and simplified the conditionals to:

if settings.JSON_FIELD_AVAILABLE and len(name_scheme['fields']) > 1
  
if len(name_scheme['fields']) > 1

hongquan (approved):

PR #1260, opened 2025-11-13
Fix: gift card 500 error

### LINKED ISSUE #1252: Ticket Component: Gift Card Received from Refund Cannot Be Used
I received a gift card after requesting a refund for a previous purchase. However, when I try to use the gift card during checkout, the system returns a HTTP 500 Error

**Expected Behaviour**

The gift card should be accepted and applied correctly during the checkout process.

**Steps to Reproduce**

1. Add any product to the cart.
2. Proceed to the checkout page.
3. Choose Gift Card -> Enter the gift card code received after the refund.
4. Observe the error message displayed by the system.

Image

Image

ArnavBallinCode:
(Fix)gift card 500 error

The issue is in the `GiftCardPayment` class in `/app/eventyay/base/payment.py` at 2 places:

### 2: `payment_prepare` method (line ~1338)

```python
gc = self.event.organizer.accepted_gift_cards.get(secret=request.POST.get('giftcard'))
```

1. `accepted_gift_cards` is a `@property` that returns an **annotated queryset**:
   ```python
   @property
   def accepted_gift_cards(self):
       return GiftCard.objects.annotate(
           accepted=Exists(...)
       ).filter(Q(issuer=self) | Q(accepted=True))
   ```

2. When calling `.get()` directly on this property, there might be issues with:
   - Case sensitivity in the secret matching
   - The queryset not being properly evaluated
   - Missing `.all()` before `.get()`

3. Django's `.get()` on an annotated queryset can sometimes fail if the secret comparison isn't exact or if there's a case mismatch

## Solution

Use `.filter()` with `.first()` instead of `.get()` for more robust handling, or use `iexact` for case-insensitive matching similar to how vouchers are handled elsewhere in the codebase.

Actually, looking at line 1309:
```python
if self.event.vouchers.filter(code__iexact=request.POST.get('giftcard')).exists():
```

The code already uses `code__iexact` for vouchers, suggesting that case-insensitive matching is expected. The gift card lookup should use the same pattern!

### The Fix:
Change from:
```python
gc = self.event.organizer.accepted_gift_cards.get(secret=request.POST.get('giftcard'))
```

To:
```python
gc = self.event.organizer.accepted_gift_cards.filter(secret__iexact=request.POST.get('giftcard')).first()
if not gc:
    raise GiftCard.DoesNotExist()
```

## Summary by Sourcery

Fix gift card lookup in checkout_prepare and payment_prepare to prevent server errors and support case-insensitive matching

Bug Fixes:
- Replace direct get() on accepted_gift_cards with filter(secret__iexact).first() and raise GiftCard.DoesNotExist when missing to avoid 500 errors

Enhancements:
- Align gift card secret matching with vouchers by using case-insensitive lookup

fixes #1252

## Summary by Sourcery

Use filter(secret__iexact).first() for gift card lookup in checkout_prepare and payment_prepare, remove DoesNotExist exceptions, and add user feedback for voucher misuse or invalid gift cards.

Bug Fixes:
- Prevent server errors when looking up gift cards by replacing direct get() calls with case-insensitive filter().first()

Enhancements:
- Match gift card secrets using case-insensitive lookup consistent with voucher code handling
- Add warning when a voucher code is entered in place of a gift card and error for unknown gift cards


app/eventyay/base/payment.py
```diff
@@ -1254,7 +1254,9 @@ def checkout_prepare(self, request: HttpRequest, cart: Dict[str, Any]) -> Union[
 
         cs = cart_session(request)
         try:
-            gc = self.event.organizer.accepted_gift_cards.get(secret=request.POST.get('giftcard'))
+            gc = self.event.organizer.accepted_gift_cards.filter(secret__iexact=request.POST.get('giftcard')).first()
+            if not gc:
+                raise GiftCard.DoesNotExist()
```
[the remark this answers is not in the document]
ArnavBallinCode: done

ArnavBallinCode:
@sourcery-ai review

app/eventyay/base/payment.py
```diff
@@ -1254,7 +1254,20 @@ def checkout_prepare(self, request: HttpRequest, cart: Dict[str, Any]) -> Union[
 
         cs = cart_session(request)
         try:
-            gc = self.event.organizer.accepted_gift_cards.get(secret=request.POST.get('giftcard'))
+            gc = self.event.organizer.accepted_gift_cards.filter(secret__iexact=request.POST.get('giftcard')).first()
```
hongquan: The GiftCard code is a secret code, it is case-sensitive, we should not looking up with `iexact`.

hongquan (changes requested):
You must chose one of two ways of error handling: via returning `None` or via raising exception.

You are mixing them: You replaced `GiftCard.DoesNotExist` exception with `.first()` (returning `None`), but still leave `GiftCard.MultipleObjectsReturned`, making the code worse: The block of handling `GiftCard.MultipleObjectsReturned` can not be reach.

hongquan:
Do you have any log of this error 500? Because your analysis sounds wrong:

> Django's .get() on an annotated queryset can sometimes fail if the secret comparison isn't exact or if there's a case mismatch

This should not cause error 500, because there is already a block to handle `GiftCard.DoesNotExist` at line `1308`:

https://github.com/fossasia/eventyay/blob/4c92d7efbff2e5a15add539bc73d10142ac97a47/app/eventyay/base/payment.py#L1308

ArnavBallinCode:
> Do you have any log of this error 500? Because your analysis sounds wrong:
> 
> > Django's .get() on an annotated queryset can sometimes fail if the secret comparison isn't exact or if there's a case mismatch
> 
> This should not cause error 500, because there is already a block to handle `GiftCard.DoesNotExist` at line `1308`:
> 
> https://github.com/fossasia/eventyay/blob/4c92d7efbff2e5a15add539bc73d10142ac97a47/app/eventyay/base/payment.py#L1308


I did run into a 500 earlier while testing this flow and wrote that comment based on what I thought was happening at the time.

I reproduced right now and its working right now🤷

ArnavBallinCode:
Sorry for the messy commit history with multiple force pushes. I was iterating on 2 different branches to reproduce and messed some things up...

I tested the reproduction and confirmed it's working correctly.

app/eventyay/base/payment.py
```diff
@@ -1375,6 +1389,10 @@ def payment_prepare(self, request: HttpRequest, payment: OrderPayment) -> Union[
                     'Please contact the organizer of this event.'
                 ),
             )
+        except Exception as e:
```
hongquan: Don't use the general `Exception` class. There may be some bugs from developers mistake that needs to be exposed early to let developers fix.

ArnavBallinCode: Done, removed

PR #1300, opened 2025-11-15
Fix Speaker Acceptance Confirmation Link Returning 500 Error

### LINKED ISSUE #1276: Talk Component: Speaker Acceptance Email Contains Broken Link (Error 500)
In the speaker acceptance email, the “Confirm your attendance” link is broken. When the speaker opens the link, the system redirects to a 500 Internal Server Error page instead of confirming their attendance.

**Expected behavior**

The “Confirm your attendance” link should open the appropriate confirmation page and allow the speaker to confirm participation successfully.

**Steps to reproduce**

- Speaker receives the acceptance email.
- Notice that the “Confirm your attendance” link cannot be clicked directly in the email.
- The speaker must manually copy and paste the link into a new browser tab to open it.
- The system returns an HTTP 500 error.

Image

ArnavBallinCode:
# Fix Speaker Acceptance Confirmation Link Returning 500 Error

The `SubmissionConfirmView.dispatch()` method tried to access `self.submission` (a cached property) before calling `super().dispatch()`, which caused an `AttributeError` because `self.request` and `self.kwargs` weren't initialized yet.

Modified `dispatch()` to manually initialize `self.request`, `self.args`, and `self.kwargs` before accessing `self.submission`:

```python
    if request.user.is_anonymous:
        return get_login_redirect(request)
    
    # Initialize request attributes before accessing self.submission
    self.request = request
    self.args = args
    self.kwargs = kwargs
    
    if not request.user.has_perm('base.is_speaker_submission', self.submission):
        self.template_name = 'cfp/event/user_submission_confirm_error.html'
    return super().dispatch(request, *args, **kwargs)
```

Fixes #1276

## Summary by Sourcery

Bug Fixes:
- Initialize request, args, and kwargs in SubmissionConfirmView.dispatch before accessing self.submission to prevent AttributeError and 500 error on confirmation link

### DIFF
--- a/app/eventyay/cfp/views/user.py
+++ b/app/eventyay/cfp/views/user.py
@@ -220,12 +220,6 @@ class SubmissionConfirmView(LoggedInEventPageMixin, SubmissionViewMixin, FormVie
         if request.user.is_anonymous:
             return get_login_redirect(request)
         
-        # Initialize request attributes before accessing self.submission
-        # to avoid AttributeError when submission property tries to access self.request/self.kwargs
-        self.request = request
-        self.args = args
-        self.kwargs = kwargs
-        
         if not request.user.has_perm('base.is_speaker_submission', self.submission):
             self.template_name = 'cfp/event/user_submission_confirm_error.html'
         return super().dispatch(request, *args, **kwargs)


ArnavBallinCode:
@copilot review

Sak1012:
There is no `Attribute Error` if that was the case the mail most likely wouldn't be sent, and there are no logs of any kind of error, so i'm not sure if these changes fixes the issue. Also I tested it with a few different cases to check if it's just this link, while testing locally it seems to works fine

image

also tried adding other urls in the mail sent from the deployment the other link works fine

image

ArnavBallinCode:
Screenshot 2025-11-27 at 7 52 15 PM

app/eventyay/cfp/views/user.py
```diff
@@ -219,6 +219,13 @@ def get_object(self):
     def dispatch(self, request, *args, **kwargs):
         if request.user.is_anonymous:
             return get_login_redirect(request)
+        
+        # Initialize request attributes before accessing self.submission
+        # to avoid AttributeError when submission property tries to access self.request/self.kwargs
+        self.request = request
+        self.args = args
+        self.kwargs = kwargs
+
```
Sak1012: This change isn’t needed Django’s `View.dispatch()` already sets `self.request`, `self.args`, and `self.kwargs` before invoking the custom `dispatch()` method. These attributes are guaranteed to be available to any mixins or properties (like `self.submission`) during the normal request lifecycle. Manually re-assigning them is redundant and can even hide deeper issues if another mixin misbehaves.

ArnavBallinCode: Done

Sak1012 (approved):

PR #1301, opened 2025-11-15
fix(permissions): ensure traits are correctly parsed as list for RawSQL ANY clause in RoomQuerySet

Sak1012:
This PR fixes a recurring `psycopg.errors.SyntaxError` occurring in WebSocket login and event room queries due to malformed SQL generated from trait-based permissions. The issue arose when `traits` were passed as a single string instead of a list, leading to SQL fragments like `e0.elem#>>'{}' IN '(eventyay-video-variation-None,...)'` instead of a valid `= ANY(ARRAY[...])` expression. The fix normalizes `traits` to a list of strings before building the RawSQL query, ensuring correct parameter binding and eliminating the syntax errors during permission evaluation.

## Summary by Sourcery

Bug Fixes:
- Fix SQL syntax errors by normalizing string traits to a list for the RawSQL ANY clause in room permission queries

### DIFF
--- a/app/eventyay/base/models/room.py
+++ b/app/eventyay/base/models/room.py
@@ -43,6 +43,10 @@ class RoomQuerySet(models.QuerySet):
 
         traits = traits or user.traits
         allow_empty_traits = not user or user.type == User.UserType.PERSON
+        # Ensure traits is always a proper list of strings for SQL parameterization
+        if traits and isinstance(traits, str):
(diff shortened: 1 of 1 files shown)


suhailnadaf509:
@Sak1012  well this would not entirely fix the issue , the problem here might be that the traits is actualy coming in as list but is getting converted to to string in the SQL lines so we might have to add something like ::text[] to cast it to a list

PR #1374, opened 2025-11-29
fix: featured toggle fails due to incorrect CSRF cookie lookup (#1373)

### LINKED ISSUE #1373: Featured toggle does not work due to incorrect CSRF cookie lookup
### Bug
The "Featured" toggle in the organizer interface does not work correctly because the CSRF token is only retrieved from a single cookie (pretalx_csrftoken).

Eventyay deployments use different CSRF cookie names depending on configuration.

### Expected
The featured toggle should correctly retrieve the CSRF token from any valid cookie.

### Actual
The AJAX request fails when the CSRF token is not found in pretalx_csrftoken.

### Fix Implemented in PR
Introduces getCsrfToken() helper that checks:
- eventyay_csrftoken
- pretalx_csrftoken
- csrftoken

### Steps to Reproduce
1. Open Organizer -> Sessions page
2. Click on the “Featured” toggle
3. Observe that the request fails

### Environment
Live / local deployment

MohammedAnasNathani:
### Description
The "Featured" toggle in the organizer interface was failing because the CSRF token was hardcoded to look only for the `pretalx_csrftoken` cookie. Eventyay deployments (especially ENext) often use `eventyay_csrftoken` or the standard Django `csrftoken`.

This PR fixes the issue by introducing a helper function `getCsrfToken()` that checks for all common cookie names in order of priority.

### Issues
Closes #1373

### Fix
- Replaced hardcoded `getCookie("pretalx_csrftoken")` with a robust `getCsrfToken()` function.
- `getCsrfToken()` now checks for:
  1. `eventyay_csrftoken` (Primary)
  2. `pretalx_csrftoken` (Legacy/Fallback)
  3. `csrftoken` (Standard Django)

### Type of Change

### How Has This Been Tested?
I manually verified in the browser console that `getCsrfToken()` correctly returns the value of `eventyay_csrftoken` if present, falling back to `pretalx_csrftoken` or `csrftoken` if the others are missing. I also verified the Featured toggle network request now includes the token.

## Summary by Sourcery

Bug Fixes:
- Fix featured toggle failures by replacing a hardcoded CSRF cookie lookup with a helper that supports multiple common CSRF cookie names.

### DIFF
--- a/app/eventyay/static/orga/js/main.js
+++ b/app/eventyay/static/orga/js/main.js
@@ -24,7 +24,7 @@ const handleFeaturedChange = (element) => {
         method: "POST",
         headers: {
             "Content-Type": "application/json",
-            "X-CSRFToken": getCookie("eventyay_csrftoken"),
+            "X-CSRFToken": getCsrfToken(),
         },
         credentials: "include",
     }
@@ -69,6 +69,16 @@ const getCookie = (name) => {
         }
     }
     return cookieValue
+}
+
+/* Helper to get CSRF token from common cookie names with priority */
+const getCsrfToken = () => {
+    // Priority 1: eventyay_csrftoken (New standard)
+    // Priority 2: pretalx_csrftoken (Legacy support)
+    // Priority 3: csrftoken (Django default)
+    return getCookie("eventyay_csrftoken") || 
+           getCookie("pretalx_csrftoken") || 
+           getCookie("csrftoken");
 }
 
 onReady(() => {


MohammedAnasNathani:
Thanks for the review! I've kept the implementation scoped to main.js to strictly fix the reported bug (#1373) without risking regressions elsewhere. I am happy to open a separate PR to refactor/unify CSRF token handling across the entire codebase if the maintainers prefer that.

Saksham-Sirohi:
not required as functionality is already working and fixed in a earlier pull request

PR #1378, opened 2025-11-29
fix: race condition in checkout allows overselling (#1359)

### LINKED ISSUE #1359: Bug: Cart Expiry Race Condition Allows Order Overselling
# Cart checkout allows overselling when lock is skipped

## The Problem

Found a race condition in `eventyay/base/services/orders.py` (lines 1180-1188) that lets multiple people buy the last ticket simultaneously.

The code has a "performance optimization" that skips locking the event during checkout when:
- No voucher is used
- Cart isn't expiring soon (>2 minutes left)
- No seat selection
```python
if positions.filter(
    Q(voucher__isnull=False) | Q(expires__lt=now() + timedelta(minutes=2)) | Q(seat__isnull=False)
).exists():
    # Performance optimization comment...
    locked = True
    lockfn = event.lock

with lockfn() as now_dt:
    # order creation
```

When that condition is false, `lockfn` stays as a no-op function and multiple checkouts can happen at once.

## What Happens

**Expected:** If there's 1 ticket left and 2 people checkout, only 1 should succeed.

**Actual:** Both succeed. You get 2 orders for a quota of 1.

## How to Reproduce

1. Create an event with 1 ticket available
2. Open two browser windows
3. Add the product to cart in both (make sure cart expires >2 mins away, no vouchers)
4. Click checkout in both windows at the same time
5. Both orders go through → quota shows 2/1 sold

Works consistently in testing, confirmed on test event `aer/xgzvvq` on local device 

## Related

Found this while fixing #1344 (payment duplicate confirmation bug). Same pattern.

MohammedAnasNathani:
### Description
Identified a critical race condition in the checkout flow where the database lock was being skipped for "performance optimization" in certain scenarios (e.g., no voucher used). This allowed multiple users to purchase the last remaining ticket simultaneously, resulting in overselling.

### Issues
Closes #1359

### Fix
- Removed the conditional check that skipped `event.lock()`.
- The checkout process now **always** acquires a lock on the event. This ensures that quota checks are atomic and strictly serialized, preventing concurrent overselling.

### Type of Change

### Verification
- Reviewed the `_perform_order` logic. The previous code explicitly set `locked = False` and `lockfn = NoLockManager` under specific conditions.
- The new code forces `locked = True` and `lockfn = event.lock` in all cases, utilizing Django's `select_for_update` mechanism via the `lock()` context manager to handle concurrency correctly.

## Summary by Sourcery

Bug Fixes:
- Fix a race condition in checkout by removing the conditional that skipped event locking in some scenarios, preventing concurrent purchases from overselling inventory.

### DIFF
--- a/app/eventyay/base/email.py
+++ b/app/eventyay/base/email.py
@@ -522,7 +522,7 @@ def base_placeholders(sender: Event, **kwargs):
         SimpleFunctionalMailTextPlaceholder(
             'expire_date',
             ['event', 'order'],
-            lambda event, order: LazyExpiresDate(order.expires.astimezone(ZoneInfo(event.timezone))),
+            lambda event, order: LazyExpiresDate(order.expires.astimezone(ZoneInfo(event.settings.timezone))),
             lambda event: LazyDate(djnow() + timedelta(days=15)),
         ),
         SimpleFunctionalMailTextPlaceholder(
@@ -824,4 +824,3 @@ def base_placeholders(sender: Event, **kwargs):
         )
 
     return ph
-
--- a/app/eventyay/base/services/mail.py
+++ b/app/eventyay/base/services/mail.py
@@ -172,7 +172,7 @@ def mail(
         bcc = []
 
         if event:
-            timezone = event.timezone
+            timezone = event.settings.timezone
             renderer = event.get_html_mail_renderer()
             if not auto_email:
                 if event_bcc:  # Use custom BCC if specified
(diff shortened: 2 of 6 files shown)


ArnavBallinCode (approved):

ArnavBallinCode:
This does solve the issue , I dont see anywhere it could go wrong except we have too much of traffic , entire flow passing through that 1 lock, but very rare case and would need too much traffic

hongquan:
Looks fine enough, though I wonder why you still keep `locked` variable. It doesn't change its value throughout its life.

MohammedAnasNathani:
Good point! I've removed the redundant locked variable and updated the payment confirmation call to explicitly use lock=False. Thank you!

PR #1416, opened 2025-12-02
fix: NoReverseMatch error in quota delete template

### LINKED ISSUE #1404: Error page when trying to delete a quota tied to a product
**Describe the bug**
When trying to delete the quota attached to a product, we get a error page like this

Image

**Expected behavior**
A simple message saying that quota is attached to a product and cannot be deleted.

**Steps to reproduce**

1. Create a new quota
2. Add a product in the quota
3. Try to delete the qouta

**Additional context**
Editing the quota and removing any products attached to it , and then deleting the quota works fine. If this is not done, an appropriate popup shall be displayed.
Please provide more clarity on the expected behaviour.

AKRITI-ENG:
Fixes #1404 
 
How I fixed it
Updated the template to use the correct URL pattern name that actually exists in the codebase:
- Changed the URL from `"control:event.item"` to `"control:event.product"`
- Updated the parameter from `item=item.pk` to `product=item.pk` to match what the URL expects

Testing done
I've tested this locally and confirmed:
- Deleting quotas without products works fine
- Deleting quotas with products now shows the proper confirmation page 
- The Cancel button takes you back to the quotas list as expected

image
image

## Summary by Sourcery

Bug Fixes:
- Resolve NoReverseMatch errors on the quota deletion page by using the existing product URL pattern and parameter.

### DIFF
--- a/app/eventyay/base/notifications.py
+++ b/app/eventyay/base/notifications.py
@@ -184,7 +184,7 @@ class ParametrizedOrderNotificationType(NotificationType):
         else:
             n.add_attribute(_('Event date'), order.event.get_date_range_display())
 
-        positions = list(order.positions.select_related('item', 'variation', 'subevent'))
+        positions = list(order.positions.select_related('product', 'variation', 'subevent'))
         fees = list(order.fees.all())
 
         n.add_attribute(_('Order code'), order.code)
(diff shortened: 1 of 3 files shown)


AKRITI-ENG:
@Sak1012 Rebased as requested

hongquan (approved):

Sak1012 (approved):

PR #1425, opened 2025-12-03
chore: room complete setup pass

Saksham-Sirohi:
## Summary by Sourcery

Ensure room setup completion status and sidebar visibility stay consistent with configured modules in both backend and admin UI.

Bug Fixes:
- Keep room.setup_complete and sidebar_hidden in sync with the presence or absence of configured modules on save.

Enhancements:
- Derive and persist the setup_complete flag on room config updates in the admin UI based on whether any modules are configured.


app/eventyay/base/services/room.py
Sak1012, line 1: The new logic in `save_room()` duplicates what the existing code already ensures.

Originally, `setup_complete` was marked `True` when `room.module_config` existed and `False` otherwise, so adding an explicit `len(...) > 0` check doesn’t change behavior. Python already treats **empty lists as false**.

The additional “defensive” branch that flips `setup_complete` back to `False` when there are no modules also introduces unnecessary churn; the field will naturally remain `False` unless explicitly set.
In short, the patch adds complexity without altering actual semantics, the existing implementation already handles both cases correctly (modules present vs. empty).

Saksham-Sirohi: addressed

PR #1438, opened 2025-12-04
Fix: Prevent Scroll-to-Top Behavior on Logout Click

### LINKED ISSUE #1437: Logout button scrolls page to top due to href="#"
**Issue**: Logout link scrolls page to top before redirect

When clicking Logout from the global navigation menu, the browser briefly scrolls the page to the top.  
This occurs because the HTML contains `href="#"` which triggers
default navigation behavior.

Image

**Current Behavior**
- Clicking Logout triggers a scroll jump before redirecting
- Happens inconsistently depending on viewport position + response time

**Expected Behavior**
- Clicking Logout should smoothly trigger logout without scroll jump

**Observations**
- The issue may not always be visible on fast local/dev servers
- The scroll jump may occur more often in production or on slow networks
- Inspecting code showed the root cause (href="#")

**Proposed Fix**
Replace `<a href="#">` with a `<button type="button">`
and keep existing logout logic with `@click.prevent="handleLogout"`.

**Verification**
- Tested locally
- No UI change
- Logout still works as expected

htssp:
**Problem**
When clicking the Logout option in the global navigation menu, the browser performs a default navigation action due to the href="#".
This causes the page to scroll to the top before logout is fully processed, leading to a poor user experience.

 **Solution Implemented**
Replaced the logout `<a href="#">` element with a semantic `<button type="button">`.

Preserved all UI styling by keeping the same utility classes

Added @click to block default navigation behavior

Logout behavior remains unchanged — handleLogout() still submits the form normally

**Result**

No scroll jump
No UI changes
Logout still works properly

Fixes #1437

## Summary by Sourcery

Improve global navigation behavior and align supported locale lists across the application.

Bug Fixes:
- Prevent page scroll-to-top on logout by replacing the logout anchor with a button that no longer triggers default navigation.

Enhancements:
- Streamline and synchronize configured locale lists for moment.js and i18n helpers by removing unused or unsupported locales.

### DIFF
--- a/app/eventyay/frontend/global-nav-menu/src/App.vue
+++ b/app/eventyay/frontend/global-nav-menu/src/App.vue
@@ -115,10 +115,14 @@ onBeforeUnmount(() => {
         <div class='i-fa-cog h-3 w-3'></div>
         <div>Admin</div>
       </a>
-      <a :class='[ENTRY_CLASSES, WITH_ICON_CLASSES, WITH_BORDER_CLASSES, "w-full text-left"]' @click='handleLogout' href="#">
+      <button
+        type='button'
+        :class='[ENTRY_CLASSES, WITH_ICON_CLASSES, WITH_BORDER_CLASSES, "w-full text-left"]'
+        @click='handleLogout'
+      >
         <div class='i-fa-sign-out h-3 w-3'></div>
         <div>Logout</div>
-      </a>
+      </button>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


mariobehling:
This is probably related to a revert of several PRs which we had to make.

Aqil-Ahmad:
@htssp please rebase. its difficult to find what is related to the fix.

PR #1454, opened 2025-12-07
fix: Validate event locale to prevent KeyError on unsupported languages (Fixes #1388)

### LINKED ISSUE #1388: KeyError when creating event with unsupported language
**Describe the bug**
When creating a new event with an unsupported language, the system allows the event to be created and added to the event list. 

**Steps to Reproduce**
* Create a new event and Select a language 
* Enter an event name in a language which doesn't match the selected language 
* Set the default language to an unsupported language and submit the event 

**Expected behavior**
* The event creation form should validate the language code against supported languages
*  Event should not be created or added to the event list if the language code is invalid

**Screenshots**

Image

MohammedAnasNathani:
**Fixes #1388**

**Issue:**
Previously, the event creation form did not validate if the submitted `locale` was present in `settings.LANGUAGES`. This allowed invalid language codes to pass through, causing a `KeyError` later in the application (specifically when accessing `settings.LANGUAGES[locale]`) and resulting in a 500 Server Error.

**Fix:**
I have added a `clean_locale` method to the `EventForm` class in `app/eventyay/orga/forms/event.py`. 
- This method checks if the submitted `locale` exists in the supported `settings.LANGUAGES`.
- If the language is unsupported, it raises a `ValidationError` with a user-friendly message ("This language is not supported") instead of letting the application crash.

**Changes:**
- Added `clean_locale` validation to `EventForm`.

**Type of Change:**

## Summary by Sourcery

Bug Fixes:
- Add locale field validation to EventForm so unsupported languages raise a user-facing validation error instead of causing a server error.

### DIFF
--- a/app/eventyay/orga/forms/event.py
+++ b/app/eventyay/orga/forms/event.py
@@ -111,9 +111,11 @@ class EventForm(ReadOnlyFlag, I18nHelpText, JsonSubfieldMixin, I18nModelForm):
             + ' '
             + str(_('You can find the page <a {href}>here</a>.')).format(href=f'href="{self.instance.urls.featured}"')
         )
+
     def clean_locale(self):
         locale = self.cleaned_data.get('locale')
-        if locale and locale not in dict(settings.LANGUAGES):
+        valid_locales = {l[0] for l in settings.LANGUAGES}
+        if locale and locale not in valid_locales:
             raise forms.ValidationError(_("This language is not supported."))
         return locale
 


mariobehling:
Please see my comment here https://github.com/fossasia/eventyay/issues/1388#issuecomment-3635311744

mariobehling:
The correct fix would be to fix Ukrainian language instead. Hence closing this. Thanks.

PR #1497, opened 2025-12-13
feat: Add Active Languages and Content Languages to event creation form

### LINKED ISSUE #1472: Event Creation: Ensure Active Languages and Content Languages options are available
When creating a new event, the event creation form at:
https://next.eventyay.com/common/events/add

does **not** include options to configure **Active Languages** or **Content Languages**.
These settings are only available later in the **Common Settings** page of an already-created event.

To make event setup faster and more consistent, these language options should also be available during event creation.

Image

### Expected Behavior

* The Event Creation page includes:

  * **Active Languages**
  * **Content Languages**
* Both fields should use the same UI as the language options on the Common Settings page (checkbox lists, alphabetical sorting, unified list of languages, etc.).
* Default language behavior should match the existing system logic.

### Current Behavior

* Event Creation page does not provide any language configuration fields.
* Organizers must create the event first and then navigate to settings to configure languages.

### Tasks

* Add **Active Languages** selection to the event creation form.
* Add **Content Languages** selection to the event creation form.
* Reuse the exact same UI component as used in the Common Settings page.
* Ensure the initial event creation process stores these values correctly.
* Test event creation with:

  * Single language
  * Multiple active languages
  * Multiple content languages
* Confirm correct behavior on desktop and mobile.

ArnavBallinCode:
https://github.com/user-attachments/assets/e00bd8d7-0776-47df-8cf6-3d95315a43e9

fixes #1472

## Summary by Sourcery

Add separate active language and content language configuration to the event creation wizard and ensure content languages are constrained to active languages.

New Features:
- Introduce a dedicated content languages field in the event foundation form to control which languages proposals can be submitted in.

Bug Fixes:
- Ensure content languages default to and remain a subset of the selected active languages, preventing invalid configuration states.

Enhancements:
- Refine localization labels and help text in the event foundation form and group language settings under a dedicated Localization section in the UI.

### DIFF
--- a/app/eventyay/eventyay_common/tasks.py
+++ b/app/eventyay/eventyay_common/tasks.py
@@ -69,83 +69,6 @@ def send_team_webhook(self, user_id, team):
             logger.error('Max retries exceeded for sending organizer webhook.')
 
 
-@shared_task(
-    bind=True, max_retries=5, default_retry_delay=60, base=CreateWorldTask
-)  # Retries up to 5 times with a 60-second delay
-def create_world(self, is_video_creation: bool, event_data: dict) -> Optional[dict]:
-    """
-    Create a video system for the specified event.
-
-    :param self: Task instance
-    :param is_video_creation: A boolean indicating whether the user has chosen to add a video.
-    :param event_data: A dictionary containing the following event details:
-        - id (str): The unique identifier for the event.
-        - title (str): The title of the event.
-        - timezone (str): The timezone in which the event takes place.
-        - locale (str): The locale for the event.
-        - token (str): Authorization token for making the request.
-        - has_permission (bool): Indicates if the user has 'can_create_events' permission or is in admin session mode.
-
-    To successfully create a world, both conditions must be satisfied:
-    - The user must have the necessary permission.
-    - The user must choose to create a video.
-    """
-
-    def _create_world(payload: dict, headers: dict) -> Optional[dict]:
-        try:
-            response = requests.post(
-                urljoin(settings.VIDEO_SERVER_HOSTNAME, 'api/v1/create-world/'),
-                json=payload,
-                headers=headers,
-            )
-            response.raise_for_status()
-            return response.json()
-        except (
-            requests.exceptions.JSONDecodeError,
-            requests.exceptions.ConnectionError,
-            requests.exceptions.Timeout,
-            requests.exceptions.RequestException,
-        ) as e:
-            error_type = type(e).__name__
-            logger.error('%s during video world creation: %s', error_type, e)
-            raise
-
-    has_permission = event_data.get('has_permission', False)
-    if not (is_video_creation and has_permission):
-        logger.info(
-            'Skipping video world creation - Video enabled: %s, Has permission: %s',
-            is_video_creation,
-            has_permission,
-        )
-        return None
-
-    event_slug = event_data.get('id', '')
-    payload = {
-        'id': event_slug,
-        'title': event_data.get('title', ''),
-        'timezone': event_data.get('timezone', ''),
-        'locale': event_data.get('locale', ''),
-        'traits': {
-            'attendee': 'eventyay-video-event-{}'.format(event_slug),
-        },
-    }
-
-    try:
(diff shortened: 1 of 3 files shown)


app/eventyay/control/forms/event.py
```diff
@@ -53,9 +53,20 @@
 class EventWizardFoundationForm(forms.Form):
     locales = forms.MultipleChoiceField(
         choices=settings.LANGUAGES,
-        label=_('Use languages'),
+        label=_('Active languages'),
         widget=MultipleLanguagesWidget,
-        help_text=_('Choose all languages that your event should be available in.'),
+        help_text=_(
```
hongquan, line 58: This help text is too long, is it necessary?

app/eventyay/control/forms/event.py
```diff
@@ -99,6 +110,26 @@ def __init__(self, *args, **kwargs):
             self.fields['organizer'].initial = qs.first()
             self.fields['organizer'].required = False
 
+    def clean(self):
+        cleaned_data = super().clean()
+        locales = cleaned_data.get('locales', [])
+        content_locales = cleaned_data.get('content_locales')
+        
+        if not isinstance(locales, list):
```
hongquan: The `MultipleChoiceField` already guarantees this value to be a list, the `isinstance` is not necessary.
If your purpose is to help Pyright / your IDE infer type correctly, you can use:

```py
assert isinstance(locales, list)
```

to keep the code flat.

app/eventyay/control/forms/event.py
```diff
@@ -99,6 +110,26 @@ def __init__(self, *args, **kwargs):
...
+            locales = list(locales)
+        
+        if content_locales is not None:
```
hongquan: This block is too deeply nested, try to make code flatter.

hongquan (changes requested):

app/eventyay/control/forms/event.py
```diff
@@ -99,6 +106,25 @@ def __init__(self, *args, **kwargs):
...
+        cleaned_data = super().clean()
+        locales = cleaned_data.get('locales', [])
+        content_locales = cleaned_data.get('content_locales')
+        
+        assert isinstance(locales, list)
+        
+        if content_locales is None or not content_locales:
+            return cleaned_data
+        
+        assert isinstance(content_locales, list)
```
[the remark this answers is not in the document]
ArnavBallinCode: @hongquan shall I revert it back then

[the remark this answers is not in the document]
ArnavBallinCode: here too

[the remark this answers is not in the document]
hongquan: Copilot knowledge is not up-to-date and doesn't understand the context of this code block. The `assert` here is for "type guard", not "validate", because you don't need to "validate", as the reason I mentioned above.

It is just a trick to help type checker tool or your IDE infer the type correctly. Ideally, you shouldn't need this trick, but because many type checker tools fail with Django (due to the fact that Django is too flexible and the type checker like Pyright is written in JS), you will need this trick or the tool will give false report.

app/eventyay/control/forms/event.py
```diff
@@ -99,6 +106,25 @@ def __init__(self, *args, **kwargs):
...
             self.fields['organizer'].required = False
 
+    def clean(self):
+        cleaned_data = super().clean()
+        locales = cleaned_data.get('locales', [])
+        content_locales = cleaned_data.get('content_locales')
+        
+        assert isinstance(locales, list)
+        
+        if content_locales is None or not content_locales:
```
hongquan: At this line, `content_locales` is of `list | None` type. `list` because it is returned by `MultipleChoiceField`, `None` because this field is optional, so the `if` check can just be as short as:

```py
if not content_locales:
```

That already covers the case of `None` and empty list.

hongquan (changes requested):

app/eventyay/control/forms/event.py
ArnavBallinCode: yeah, right, changed

PR #1511, opened 2025-12-16
Fix AttributeError in GiftCardPayment: use 'product' instead of 'item'

ArnavBallinCode:
This happens at `/checkout/payment/` when the `GiftCardPayment.checkout_prepare()` method iterates over cart positions and incorrectly accesses `p.item.issue_giftcard`.

Screenshot 2025-12-16 at 9 21 10 AM

## Root Cause

The `GiftCardPayment` class in `app/eventyay/base/payment.py` uses `p.item` to access the product, but both `CartPosition` and `OrderPosition` models inherit from `AbstractPosition` which defines the product relationship as `product`, not `item`:

```python
# In AbstractPosition (orders.py line 1290)
product = models.ForeignKey(Product, verbose_name=_('Product'), on_delete=models.PROTECT)
```

## Summary by Sourcery

Bug Fixes:
- Fix AttributeError in gift card payment processing by using the correct product relationship on cart and order positions.

### DIFF
--- a/app/eventyay/orga/templatetags/staff_session.py
+++ b/app/eventyay/orga/templatetags/staff_session.py
@@ -10,6 +10,8 @@ register = template.Library()
 
 @register.simple_tag()
 def has_event_perm(perm, user, request, obj=None):
+    if not user.is_authenticated:
+        return False
     return is_admin_mode_active(request) or User.has_perm(user, perm, obj)
 
 


mariobehling:
> Could you scan `.item.` in the source code to fix more occurences like this?

I think this could be handled in a separate issue. This PR seems to solve the specific issue here already.

PR #1516, opened 2025-12-16
Fix : Address Already in Use (Port Clash)

Yaannko:
- **Issue**: Attempting to run the server when it is already running.
- **Status**: **Resolved** (Modified `manage.py` to detect if the port is in use and suggest a kill command).
- **Action**: None required.

## Summary by Sourcery

Detect when the Django development server port is already in use and provide a helpful kill command hint, while updating auth settings and tightening the supported Python version range.

New Features:
- Add a pre-run check when invoking `runserver` to detect if the target host/port is already in use and display guidance on how to terminate the conflicting process.

Enhancements:
- Adjust django-allauth configuration to explicitly set email-based login and signup fields in alignment with updated setting names.
- Constrain the supported Python version range in pyproject configuration to exclude Python 3.13 and above.

### DIFF
--- a/app/manage.py
+++ b/app/manage.py
@@ -17,6 +17,44 @@ def main():
             'available on your PYTHONPATH environment variable? Did you '
             'forget to activate a virtual environment?'
         ) from exc
+    if len(sys.argv) > 1 and sys.argv[1] == 'runserver':
+        import socket
+        import re
+
+        # Default port
+        port = 8000
+        host = '127.0.0.1'
+
+        # Try to parse port/host from arguments
+        for arg in sys.argv[2:]:
+            if arg.startswith('--noreload'):
+                continue
+            # Handle [addr:]port format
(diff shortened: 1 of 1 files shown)


hongquan (changes requested):
This is unnecessary change.

- `runserver` is Django built-in command. Don't override it, because you may cause different behaviour from Django documentation and surprise the developers.
- The handling of CLI argument is more complex than you thought, you are trying to replicate it but with naiver method and it can introduce bugs.
- Port `8000` is too common, many softwares use it, not only Django. Don't assume that user is running another Django instance. If other software is using this port, the user should stop that software in proper way, don't tell user to stop it by `kill` command. It is like telling user to shutdown computer by disconnecting the power plug!

Sak1012:
Thank you for the PR this change is currently not of priority for the system, please continue with other contributions.

Issue #1481, opened 2025-12-11
Enhance Attendee Login Requirement Before Checkout With Login Pop Up and Seamless Return to Checkout


mariobehling:
Some events require that attendees be authenticated before placing an order. Currently, the system allows anonymous checkout. The setting for organizers to enforce login is in the tickets settings (a bit hidden).

1. The default should be "attendees should be logged in".
2. Attendees should see a pop up to login.
3. The user should be returning to the checkout page after successful login, also when using external providers such as Wikimedia.

<img width="2448" height="774" alt="Image" src="https://github.com/user-attachments/assets/cc1c1155-638c-419a-89cd-adf256d1ab61" />

### Behavior When Enabled

1. A user browsing tickets can select tickets normally.

2. When the user clicks **“Checkout”**, if they are **not logged in**:

   * A **login popup** should appear.
   * The popup should show all available authentication methods (email, SSO options, Wikimedia, Google, etc.).
   * The user must log in before proceeding to the order form.

3. After successful login:

   * The system should return the user directly to the **checkout flow**, not the homepage or dashboard.

### Expected Behavior

* Unauthenticated users see a login popup instead of a broken redirect or error page.
* Checkout process resumes smoothly after login.

### Current Behavior

* Unauthenticated users can reach the checkout page.
* Some events need restricted access and cannot enforce it.

mariobehling:
Basically our goal is to get the same process that was working well on the legacy eventyay.com app. Test this currently here still for example https://eventyay.com/e/88882f3e

<img width="2364" height="1306" alt="Image" src="https://github.com/user-attachments/assets/b79a97de-9204-4936-b155-d8afb86afc45" />

mariobehling:
We knew that this feature was already in the backend, but we did not realize that there is actually also an option in the ticket settings. Please see screenshot below. So, I adjusted the description in the issue to reflect this. An important goal of the issue is still that attendees get redirected back to the tickets checkout after logging in with external SSOs, e.g. Wikimedia. Please test this in your PR.

![Image](https://github.com/user-attachments/assets/2bb351c9-f9a5-46ee-a369-4c1c6d4b358b)

Issue #813, opened 2025-07-29
Show error message for not-logged in users when clicking Join Online Event


mariobehling:
When a user access the event at https://wikimedia.eventyay.com/tickets/wikimedia/wikimania2025/ and clicks "Join online event" there is no error message.

Expected show the same error message as for logged in users. See below.

<img width="1944" height="710" alt="Image" src="https://github.com/user-attachments/assets/c411ac8c-dc48-4210-aa1e-8eb01028a1e6" />

vijay-x-Raj:
hey @mariobehling can i try this actually i am in the process of understanding the tech stack

Gagan-Ram:
@mariobehling 
I tested this issue in 3 different accounts and here what I found.
1. Account from which I bought a virtual ticket for the event. Here when I click the "Join online event", it does take me to online video.
2. Account from which I didn't bought a ticket (Non ticket holder). It seems the error message is shown as below.

<img width="1390" height="480" alt="Image" src="https://github.com/user-attachments/assets/78b9a662-eeb3-4645-9572-5549a2fda31e" />

3. when an anonymous user clicks "Join online event", it shows the below.

<img width="1408" height="526" alt="Image" src="https://github.com/user-attachments/assets/13f206f3-0ce8-43ef-8e44-a890ba1f0709" />

mariobehling:
<img width="1136" height="426" alt="Image" src="https://github.com/user-attachments/assets/643cc4a9-3dac-4995-ac19-b13ab38951b7" />

You can test this on https://next.eventyay.com/a-team/wikipedia25/

It is not working here yet.

mariobehling:
Identical with https://github.com/fossasia/eventyay/issues/1470

On the event start page, if a user **is not logged in** and clicks **“Join Online Event”**, the system displays a **full error page** instead of showing the expected **login-required popup**.

The correct behavior is already implemented in other components, but the start page still redirects to an error page.

<img width="2458" height="1292" alt="Image" src="https://github.com/user-attachments/assets/e61d7a72-816c-41f1-a01a-4aabf5cbcaf3" />

### Expected Behavior

* When a non-logged-in user clicks **“Join Online Event”**, a **login-required popup/modal** should appear.
* The user should not be redirected to an error page.
* Behavior should match the talk component and other areas that already use the popup.

### Current Behavior

* The user is taken to a full error page.
* No popup is shown.
* This prevents users from understanding that login is required.

### Tasks

[...]

PR #1523, opened 2025-12-16
enhance: Enforce Login Before Checkout With Modal Login and Post Login Redirect to Checkout

ArnavBallinCode:
Implements https://github.com/fossasia/eventyay/issues/1481
Fixes https://github.com/fossasia/eventyay/issues/1481

https://github.com/user-attachments/assets/0265a161-30e8-4304-ba79-d46903c2be41

## Summary by Sourcery

Require a registered account for checkout by introducing a login modal on the event page and adding an organizer setting to control this behavior.

New Features:
- Add a checkout login modal with email/password and SSO options that intercepts the checkout button when login is required.
- Introduce an event setting to require a registered account for obtaining tickets, exposed in the event settings form.

Enhancements:
- Improve the checkout flow by using an in-page modal and AJAX login instead of redirecting directly to the login page.

## Summary by Sourcery

Require users to log in via an in-page modal before starting checkout when guest checkout is disabled, while preserving existing behavior for current events.

New Features:
- Add a checkout login modal on the event page with email/password and SSO options that intercepts the checkout button when login is required.
- Expose an event setting to require a registered account for obtaining tickets in the event settings form.

Enhancements:
- Streamline the checkout flow by handling login in-page with AJAX instead of redirecting to the standalone login page.

Chores:
- Backfill the new require-registered-account setting to be disabled for all existing events via a data migration to maintain current behavior.

### DIFF
--- a/app/eventyay/base/configurations/default_setting.py
+++ b/app/eventyay/base/configurations/default_setting.py
@@ -2470,6 +2470,7 @@ Your {event} team"""
                 'The system generates by default {}-character long gift card codes. However, if a different length '
                 'is required, it can be set here.'.format(settings.ENTROPY['giftcard_secret'])
             ),
+            min_value=4,
         ),
     },
     'giftcard_expiry_years': {
@@ -2481,9 +2482,12 @@ Your {event} team"""
             label=_('Validity of gift card codes in years'),
             help_text=_(
                 'If you set a number here, gift cards will by default expire at the end of the year after this '
-                'many years. If you keep it empty, gift cards do not have an explicit expiry date.'
+                'many years. For example, 1 means expiry at the end of next year, 2 means end of the year after, etc. '
+                'If you keep it empty, gift cards do not have an explicit expiry date.'
             ),
+            min_value=1,
         ),
+        'serializer_kwargs': dict(validators=[MinValueValidator(0)]),
     },
     'privacy_policy': {
         'default': None,
--- a/app/eventyay/base/forms/__init__.py
+++ b/app/eventyay/base/forms/__init__.py
@@ -75,6 +75,10 @@ class SettingsForm(i18nfield.forms.I18nFormMixin, HierarkeyForm):
             if isinstance(field, i18nfield.forms.I18nFormField):
                 field.widget.enabled_locales = self.locales
             self.fields[fname] = field
+            if fname not in self.initial or self.initial[fname] is None:
+                default_value = DEFAULTS[fname].get('default')
+                if default_value:
+                    self.initial[fname] = default_value
         for k, f in self.fields.items():
             if isinstance(f, (RelativeDateTimeField, RelativeDateField)):
                 f.set_event(self.obj)
--- a/app/eventyay/base/models/event.py
+++ b/app/eventyay/base/models/event.py
@@ -2206,8 +2206,12 @@ class Event(
     @cached_property
     def _visible_logo_path(self):
         """
-        Resolve a usable logo path/URL from common settings (event_logo_image/logo_image).
+        Resolve a usable logo path/URL from event_logo_image setting.
         Returns a storage-relative path (e.g. ``pub/...``) or an absolute URL.
+        
+        NOTE: This method ONLY checks for event_logo_image, NOT logo_image.
+        The logo_image setting is actually used for HEADER images (see default_setting.py),
+        so we must NOT use it here to prevent header images from appearing as logos.
         """
         def _extract_path(obj):
             if not obj:
@@ -2220,7 +2224,8 @@ class Event(
                 return obj.url
             return str(obj)
 
-        for key in ('event_logo_image', 'logo_image'):
+        # Only check event_logo_image - NOT logo_image (which is for header images)
+        for key in ('event_logo_image',):
             settings_logo = self.settings.get(key, default=None) or getattr(self.settings, key, None)
             path = _extract_path(settings_logo)
             if not path:
--- a/app/eventyay/common/templates/common/forms/field.html
+++ b/app/eventyay/common/templates/common/forms/field.html
@@ -3,7 +3,7 @@
     {% block field_label_left %}
         {% if field.field.widget.input_type != "checkbox" or field.field.widget.allow_multiple_selected %}
             <label for="{{ field.auto_id }}" class="{{ label_class }}">
-                {{ field.label }}
+                <span>{{ field.label }}<span class="d-inline text-danger">{% if field.field.required %} *{% endif %}</span></span>
                 {% if not field.field.required %}<span class="optional">{% translate "Optional" %}</span>{% endif %}
             </label>
         {% endif %}
--- a/app/eventyay/common/templates/common/fragment_nav.html
+++ b/app/eventyay/common/templates/common/fragment_nav.html
@@ -41,8 +41,9 @@
                 href="{% url 'cfp:event.start' event=current_event.slug organizer=current_event.organizer.slug %}">
                 <i class="fa fa-bullhorn"></i> {% translate "Call for Speakers" %}
             </a>
-        {% else %}
-            <a class="header-tab disabled" tabindex="-1" aria-disabled="true" href="#">
+        {% elif not cfp.settings.hide_after_deadline %}
+            <a class="header-tab {% if '/cfp' in path %}active{% endif %}"
+                href="{% url 'cfp:event.start' event=current_event.slug organizer=current_event.organizer.slug %}">
                 <i class="fa fa-bullhorn"></i> {% translate "Call for Speakers (Closed)" %}
             </a>
         {% endif %}
--- a/app/eventyay/common/templatetags/event_tags.py
+++ b/app/eventyay/common/templatetags/event_tags.py
@@ -40,14 +40,19 @@ def cfp_locale_switch_url(context, locale_code):
 def short_user_label(user):
     """
     Compact user display: prefer first name, then name, then email local part.
-    Truncate to 10 chars with ellipsis when longer.
+    Truncate to 11 chars with ellipsis when longer.
     """
     if not user:
         return ''
-    first = getattr(user, 'first_name', None) or getattr(user, 'firstname', None) or getattr(user, 'name', None)
+    first = getattr(user, 'first_name', None) or getattr(user, 'firstname', None)
+    if not first:
+        fullname = getattr(user, 'fullname', None) or getattr(user, 'name', None)
+        if fullname:
+            parts = fullname.split()
+            first = parts[0] if parts else fullname
     email = getattr(user, 'email', '') or ''
     label = (first or '').strip() or (email.split('@')[0] if email else '')
-    if len(label) > 10:
-        label = label[:10] + '...'
+    if len(label) > 11:
+        label = label[:11] + '…'
     return label
 
--- a/app/eventyay/eventyay_common/templates/eventyay_common/dashboard/dashboard.html
+++ b/app/eventyay/eventyay_common/templates/eventyay_common/dashboard/dashboard.html
@@ -31,6 +31,8 @@
                   {% endif %}
               </div>
           </div>
+        {% empty %}
+        <div class="alert alert-info">{% trans "You have no upcoming events" %}</div>
       {% endfor %}
     </div>
     {% if upcoming %}
--- a/app/eventyay/presale/templates/pretixpresale/fragment_login_status.html
+++ b/app/eventyay/presale/templates/pretixpresale/fragment_login_status.html
@@ -1,6 +1,7 @@
 {% load i18n %}
 {% load eventurl %}
 {% load login_url %}
+{% load event_tags %}
 
 {{ 'popover-profile'|json_script:"popover_toggle" }}
 <nav class="login-hdr" aria-label="{% translate 'account' %}" style="{% if request.event and request.event.settings.event_logo_image %}margin-right: 18rem{% else %}margin-right: 0{% endif %}">
@@ -8,7 +9,7 @@
     {% if request.user.is_authenticated %}
         <div class="navigation-button">
             <a target="_blank" class="header-nav btn btn-outline-success" data-toggle="popover-profile">
-                <i class="fa fa-user"></i> {{ request.user.fullname|default:request.user.email }}
+                <i class="fa fa-user"></i> {{ request.user|short_user_label }}
                 <i class="fa fa-solid fa-caret-down"></i>
             </a>
         </div>


app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,288 @@ <h3>
...
+            e.preventDefault();
+            checkoutUrl = this.getAttribute('data-checkout-url') || '';
+            if (nextUrlInput) nextUrlInput.value = checkoutUrl;
+
+            // Update SSO links with next URL
+            var ssoLinks = modal.querySelectorAll('.checkout-sso-btn');
+            ssoLinks.forEach(function(link) {
+                var baseUrl = link.href.split('?')[0];
+                var backend = link.getAttribute('data-provider');
+                link.href = baseUrl + '?backend=' + backend + '&next=' + encodeURIComponent(checkoutUrl);
```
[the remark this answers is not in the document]
ArnavBallinCode: need to do this  myself and test

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,288 @@ <h3>
...
+            // Check if login was successful
+            var checkLogin = setInterval(function() {
+                try {
+                    if (popup.closed) {
+                        clearInterval(checkLogin);
+                        // Reload to check if user is now logged in
+                        window.location.reload();
+                    }
+                } catch(e) {}
+            }, 500);
```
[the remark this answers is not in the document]
ArnavBallinCode: okay, fixing this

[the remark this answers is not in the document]
ArnavBallinCode: yup, fixing this

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,288 @@ <h3>
...
+
+            // Check if login was successful
+            var checkLogin = setInterval(function() {
+                try {
+                    if (popup.closed) {
+                        clearInterval(checkLogin);
+                        // Reload to check if user is now logged in
+                        window.location.reload();
+                    }
+                } catch(e) {}
```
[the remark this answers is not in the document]
ArnavBallinCode: Not needed, it's just checking popup.closed

ArnavBallinCode:
yup, done in previous commit itself.. updated recording.. 

> In this follow up PR please also remove the red note box
> 
> Screenshot from 2025-12-17 01-11-18
>

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,322 @@ <h3>
...
+    color: #666;
+}
+
+.checkout-login-footer a {
+    color: #337ab7;
+}
+</style>
+
+<script>
+(function() {
```
hongquan: This style of JS (wrapping code in anonymous function) is outdated and verbose. Please use `<script type='module'>` JS.

hongquan (changes requested):

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,322 @@ <h3>
...
+        loginBtn.addEventListener('click', function(e) {
+            e.preventDefault();
+            checkoutUrl = this.getAttribute('data-checkout-url') || '';
+            if (nextUrlInput) nextUrlInput.value = checkoutUrl;
+
+            // Update SSO links with next URL
+            var ssoLinks = modal.querySelectorAll('.checkout-sso-btn');
+            ssoLinks.forEach(function(link) {
+                var baseUrl = link.href.split('?')[0];
+                var backend = link.getAttribute('data-provider');
```
hongquan: Use [`.dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) instead of `getAttribute`.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,322 @@ <h3>
...
+                link.href = baseUrl + '?backend=' + backend + '&next=' + encodeURIComponent(checkoutUrl);
+            });
+
+            // Update register link
+            var registerLink = modal.querySelector('.checkout-login-footer a');
+            if (registerLink) {
+                try {
+                    var url = new URL(registerLink.href, window.location.origin);
+                    url.searchParams.set('next', checkoutUrl);
```
hongquan: Please extract the "next" string to a constant so that we can rename it later. I think we need to revise the login / register feature somedays. I feel that there is conflict with the use of "next".

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,320 @@ <h3>
...
+
+            <!-- SSO Provider Buttons -->
+            <div class="checkout-login-sso">
+                <a href="{% url 'eventyay_common:auth.login' %}?backend=mediawiki" class="btn btn-default btn-block checkout-sso-btn" target="_blank" rel="noopener noreferrer" data-provider="mediawiki">
+                    {% trans "Login with MediaWiki" %}
+                </a>
+                <a href="{% url 'eventyay_common:auth.login' %}?backend=google" class="btn btn-default btn-block checkout-sso-btn" target="_blank" rel="noopener noreferrer" data-provider="google">
+                    {% trans "Login with Google" %}
+                </a>
+            </div>
```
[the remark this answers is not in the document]
ArnavBallinCode: done

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,333 @@ <h3>
...
+    position: absolute;
+    left: 0;
+    top: 50%;
+    width: 100%;
+    height: 1px;
+    background: #ddd;
+}
+
+.checkout-login-divider span {
+    background: white;
```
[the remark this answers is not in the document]
ArnavBallinCode: This is intentional. This is a standard UX pattern.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,333 @@ <h3>
...
+                });
+
+                // Check for successful response (200-299) or redirect (302)
+                // Login endpoint returns 302 on success, not 200
+                if (response.ok || response.status === 302) {
+                    // Login successful, redirect to checkout
+                    window.location.href = checkoutUrl;
+                } else {
+                    throw new Error('{% trans "Invalid email or password. Please try again." %}');
+                }
```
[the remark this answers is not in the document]
ArnavBallinCode: Already implemented on line 336: `if (response.ok || response.status === 302).` The frontend correctly handles both 200 and 302 responses.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,333 @@ <h3>
...
+
+    // Clean up interval when modal closes
+    if (closeBtn) {
+        closeBtn.addEventListener('click', () => {
+            if (activeCheckLogin) {
+                clearInterval(activeCheckLogin);
+                activeCheckLogin = null;
+            }
+        });
+    }
```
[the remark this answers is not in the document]
ArnavBallinCode: done, removed

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,333 @@ <h3>
...
+    let checkoutUrl = '';
+
+    // Constants for SSO popup
+    const POPUP_WIDTH = 600;
+    const POPUP_HEIGHT = 700;
+    const POPUP_CHECK_INTERVAL = 500;
+    let activeCheckLogin = null;
+
+    // URL parameter name for redirect after login/register
+    const REDIRECT_PARAM = 'next';
```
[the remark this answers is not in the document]
ArnavBallinCode: Not needed. These are component-specific configuration values, not application-wide constants. They're already `const`, preventing reassignment

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,333 @@ <h3>
...
+                    if (popup.closed) {
+                        clearInterval(activeCheckLogin);
+                        activeCheckLogin = null;
+                        // Reload to check if user is now logged in
+                        window.location.reload();
+                    }
+                } catch(e) {
+                    // Ignore expected cross-origin access errors
+                }
+            }, POPUP_CHECK_INTERVAL);
```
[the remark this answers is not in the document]
ArnavBallinCode: This is not a bug - it's intentional behavior. When the SSO popup closes, we reload the page to check authentication state. The user expects to be redirected after logging in via SSO. The checkoutUrl redirect only applies to the AJAX form login (different flow).

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -64,3 +64,333 @@ <h3>
...
+            });
+
+            const registerLink = modal.querySelector('.checkout-login-footer a');
+            if (registerLink) {
+                try {
+                    const url = new URL(registerLink.href, window.location.origin);
+                    url.searchParams.set(REDIRECT_PARAM, checkoutUrl);
+                    registerLink.href = url.toString();
+                } catch (e) {
+                    // fallback
```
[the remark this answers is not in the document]
ArnavBallinCode: not needed, this is sufficient for anyone to understand !

app/eventyay/base/migrations/0005_set_require_login_default_for_existing_events.py
```diff
@@ -0,0 +1,37 @@
...
+def set_existing_events_login_not_required(apps, schema_editor):
+    """
+    Set require_registered_account_for_tickets to False for all existing events.
+    New events will get the default value of True from default_setting.py.
+    """
+    Event = apps.get_model('base', 'Event')
+    
+    # Use iterator() to stream results for better performance with large datasets
+    for event in Event.objects.iterator():
+        event.settings.set('require_registered_account_for_tickets', False)
```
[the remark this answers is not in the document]
ArnavBallinCode: done

app/eventyay/base/migrations/0005_set_require_login_default_for_existing_events.py
Sak1012, line 1: Please avoid adding migration files manually. Migration files should only be committed when they are generated via `python manage.py makemigrations` to ensure consistency and correctness.

Sak1012 (changes requested):

app/eventyay/base/migrations/0005_set_require_login_default_for_existing_events.py
ArnavBallinCode: there is no model change here ! 
it is a data migration, no schema change , created byt python manage.py makemigrations --empty and then manually edited

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +14,374 @@ <h3></h3>
...
+            e.preventDefault();
+            checkoutUrl = loginBtn.dataset.checkoutUrl || '';
+            if (nextUrlInput) nextUrlInput.value = checkoutUrl;
+
+            // Update SSO links with next URL
+            const ssoLinks = modal.querySelectorAll('.checkout-sso-btn');
+            ssoLinks.forEach((link) => {
+                const baseUrl = link.href.split('?')[0];
+                const backend = link.dataset.provider;
+                link.href = `${baseUrl}?backend=${backend}&${REDIRECT_PARAM}=${encodeURIComponent(checkoutUrl)}`;
```
hongquan: It is cleaner if you build the whole query string with `URLSearchParams`.
By the way, you should check if `checkoutUrl` is non-empty before adding "next" param.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +14,374 @@ <h3></h3>
...
+            });
+
+            const registerLink = modal.querySelector('.checkout-login-footer a');
+            if (registerLink) {
+                try {
+                    const url = new URL(registerLink.href, window.location.origin);
+                    url.searchParams.set(REDIRECT_PARAM, checkoutUrl);
+                    registerLink.href = url.toString();
+                } catch (e) {
```
hongquan: The code in the `try` block looks safe, I wonder you foresee which error can happen, that need to catch?

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +14,374 @@ <h3></h3>
...
+            <button id="join-online-close-button" class="btn btn-default join-online-close-button">Close</button>
+        </div>
+    </div>
+</div>
+
+<!-- Checkout Login Popup Modal -->
+<div id="checkout-login-modal" class="checkout-login-overlay" style="display: none;">
+    <div class="checkout-login-popup" role="dialog" aria-modal="true" aria-labelledby="checkout-login-modal-title">
+        <div class="checkout-login-header">
+            <h3 id="checkout-login-modal-title">{% trans "Login" %}</h3>
```
hongquan: Please use [`translate`](https://docs.djangoproject.com/en/5.2/topics/i18n/translation/#translate-template-tag) tag, not `trans` (older).
Btw, Django supports shorter syntax:

```twig
{{ _('Login') }}
```

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +14,374 @@ <h3></h3>
...
+            activeCheckLogin = setInterval(() => {
+                try {
+                    if (popup.closed) {
+                        clearInterval(activeCheckLogin);
+                        activeCheckLogin = null;
+                        // Reload to check if user is now logged in
+                        window.location.reload();
+                    }
+                } catch(e) {
+                    // Ignore expected cross-origin access errors
```
hongquan: Don't make the error silent, it make us difficult to debug.

hongquan (changes requested):

app/eventyay/base/migrations/0005_set_require_login_default_for_existing_events.py
hongquan: You should use data migration only when that data is essential for normal operation, for example, you are about to convert a nullable column to non-nullable, then right before you convert the column, you need a "data migration" to fill some value to the field of all records, or the migration process will break.

If that data is not so essential, you should create a Django command to fill data (then remove the command). The migration files need special care (like we need to maintain their order), so we try to keep the number of files to minimum.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
Sak1012, line 1: Some of the changes of this file seems to have overlapping fixes for #813 which was solved recently in #1533, please take a look.

Sak1012, line 1: The large inline <script> block is inline JavaScript and is not CSP-safe, it also bypasses the existing static asset pipeline. The JS should be moved to a static file and loaded via {% static %}, with the template only providing markup and data attributes.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
ArnavBallinCode: like if URL fails with invalid URL ?

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
ArnavBallinCode: Should I add console.debug for the message or like how ?

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
ArnavBallinCode: yup, resolved merge conflict before itself

ArnavBallinCode: The script uses `type = "module"` which is CSP-safe

ArnavBallinCode: also this is a template not a static asset , wont bypass the pipeline

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
ArnavBallinCode: Done

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
ArnavBallinCode: Fixed

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
ArnavBallinCode: added console.debug for now

app/eventyay/base/migrations/0005_set_require_login_default_for_existing_events.py
ArnavBallinCode: Removed migration !

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
hongquan: Agree with Srivatsa, but let it be future improvement.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +14,380 @@ <h3></h3>
...
+                    }
+                });
+
+                // Check for successful response (200-299) or redirect (302)
+                // Login endpoint returns 302 on success, not 200
+                if (response.ok || response.status === 302) {
+                    // Login successful, redirect to checkout
+                    window.location.href = checkoutUrl;
+                } else {
+                    throw new Error('{% trans "Invalid email or password. Please try again." %}');
```
hongquan: Just let the original error (`HTTPError`) raise because it has more detail for debugging and also more accurate.

By replacing with `new Error`, you obscure the true error, it may not be "wrong user name or password".

hongquan (changes requested):

mariobehling:
I updated the title of this PR in order to reflect the updates in the issue and as this is the second follow up PR to solve the issue.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +14,406 @@ <h3></h3>
...
+
+            const registerLink = modal.querySelector('.checkout-login-footer a');
+            if (registerLink) {
+                try {
+                    const url = new URL(registerLink.href, window.location.origin);
+                    url.searchParams.set(REDIRECT_PARAM, checkoutUrl);
+                    registerLink.href = url.toString();
+                } catch (e) {
+                    // Fallback if URL constructor fails with invalid URL
+                    // Build URL manually as last resort
```
hongquan: Building URL manually is less safe than using the JS built-in utilities, why using it as fallback?

If error occurs, you should print it to see what is the cause, in order to fix.

hongquan (changes requested):

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
ArnavBallinCode: yup, had made this changefor sso pop up too , let me make it here too

hongquan: You still ignore this:

```
Building URL manually is less safe than using the JS built-in utilities, why using it as fallback?
```

ArnavBallinCode: using URLSearchParams in fallback, had made changes on local , also making UX changes according to copilot , need to test and push

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +14,406 @@ <h3></h3>
...
+                    }
+                });
+
+                // Login endpoint returns 302 on success
+                if (response.status === 302) {
+                    window.location.href = checkoutUrl;
+                    return;
+                }
+
+                // 200 response from Django form means validation failed (invalid credentials)
```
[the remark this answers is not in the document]
ArnavBallinCode: The fetch API automatically follows redirects by default
A 302 redirect will be followed getimg a 200 response from the final destination
already handling this correctly by checking status codes

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +14,406 @@ <h3></h3>
...
+            const left = (window.innerWidth - POPUP_WIDTH) / 2;
+            const top = (window.innerHeight - POPUP_HEIGHT) / 2;
+            const popup = window.open(link.href, 'sso_login', `width=${POPUP_WIDTH},height=${POPUP_HEIGHT},left=${left},top=${top}`);
+
+            if (!popup) {
+                // Show user-friendly error in the modal instead of alert
+                if (errorDiv) {
+                    errorDiv.textContent = '{% trans "Please allow popups for this site to use SSO login." %}';
+                    errorDiv.style.display = 'block';
+                }
```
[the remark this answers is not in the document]
ArnavBallinCode: done

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +14,406 @@ <h3></h3>
...
+                } catch (e) {
+                    // Fallback if URL constructor fails with invalid URL
+                    // Build URL manually as last resort
+                    registerLink.href = `${registerLink.href.split('?')[0]}?${REDIRECT_PARAM}=${encodeURIComponent(checkoutUrl)}`;
+                }
+            }
+
+            modal.style.display = 'flex';
+        });
+    }
```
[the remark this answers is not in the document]
ArnavBallinCode: done

hongquan:
Failed to test, with this error: 
image

norbusan:
> Failed to test, with this error

Yup, I have seen the same

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -110,10 +111,10 @@ <h3 id="checkout-login-modal-title">{% translate "Login" %}</h3>
 
             <!-- SSO Provider Buttons -->
             <div class="checkout-login-sso">
-                <a href="{% url 'eventyay_common:auth.login' %}?backend=mediawiki" class="btn btn-default btn-block checkout-sso-btn" rel="noopener noreferrer" data-provider="mediawiki">
+                <a href="/oauth_login/mediawiki/" class="btn btn-default btn-block checkout-sso-btn" rel="noopener noreferrer" data-provider="mediawiki">
```
hongquan, line 114: Please use Django URL reverse instead of hardcoding the URL.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +15,421 @@ <h3></h3>
...
+    position: absolute;
+    left: 0;
+    top: 50%;
+    width: 100%;
+    height: 1px;
+    background: #ddd;
+}
+
+.checkout-login-divider span {
+    background: white;
```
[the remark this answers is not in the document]
hongquan: Our website doesn't support dark theme yet.

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +16,424 @@ <h3></h3>
...
+
+    // AJAX form submission
+    if (loginForm) {
+        loginForm.addEventListener('submit', async (e) => {
+            e.preventDefault();
+
+            const submitBtn = document.getElementById('checkout-login-submit');
+            const originalText = submitBtn.innerHTML;
+            submitBtn.disabled = true;
+            submitBtn.innerHTML = '<i class="fa fa-spinner fa-spin"></i> {% trans "Logging in..." %}';
```
hongquan: You still use deprecated `trans` tag...

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +16,424 @@ <h3></h3>
...
+                // Also show in a more prominent global error container
+                const globalErrorDiv = document.getElementById('ajaxerr');
+                if (globalErrorDiv) {
+                    globalErrorDiv.textContent = popupErrorMessage;
+                    globalErrorDiv.style.display = 'block';
+                    globalErrorDiv.setAttribute('role', 'alert');
+                    globalErrorDiv.setAttribute('aria-live', 'assertive');
+                    // Set tabindex to allow focus for screen reader announcement
+                    globalErrorDiv.setAttribute('tabindex', '-1');
+                    globalErrorDiv.focus?.();
```
hongquan: Is this a valid JS syntax?

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +16,424 @@ <h3></h3>
...
+                    if (popup.closed) {
+                        clearInterval(activeCheckLogin);
+                        activeCheckLogin = null;
+                        // Small delay to ensure session cookie is set before reload
+                        setTimeout(() => {
+                            window.location.href = checkoutUrl;
+                        }, 300);
+                    }
+                } catch(e) {
+                    // Silently ignore errors while checking popup status to avoid
```
hongquan: A previous comment said that, don't silent the error, or we have no clue to debug!

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
ArnavBallinCode: The socialauth plugin URLs are registered at the root level in the multidomain URL configuration, but the namespace isn't accessible from template URL tags. Using direct paths `(/oauth_login/<provider>/)` which
are the actual registered URL patterns, matching the approach used in the codebase for root-level plugin URLs."

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +16,424 @@ <h3></h3>
...
+            <div class="checkout-login-divider">
+                <span>{% translate "or" %}</span>
+            </div>
+
+            <!-- SSO Provider Buttons -->
+            <div class="checkout-login-sso">
+                <a href="/oauth_login/mediawiki/" class="btn btn-default btn-block checkout-sso-btn" rel="noopener noreferrer" data-provider="mediawiki">
+                    {% translate "Login with MediaWiki" %}
+                </a>
+                <a href="/oauth_login/google/" class="btn btn-default btn-block checkout-sso-btn" rel="noopener noreferrer" data-provider="google">
```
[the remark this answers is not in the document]
ArnavBallinCode: The socialauth plugin URLs are registered at the root level in the multidomain URL configuration, but the namespace isn't accessible from template URL tags. Using direct paths (/oauth_login/<provider>/) which
are the actual registered URL patterns, matching the approach used in the codebase for root-level plugin URLs."


Already tried and tested with this

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
ArnavBallinCode: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +16,424 @@ <h3></h3>
...
+    /* Use semi-transparent black for overlay, respects system themes */
+    background: rgba(0, 0, 0, 0.6);
+    z-index: 10000;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.checkout-login-popup {
+    background: white;
```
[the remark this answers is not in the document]
ArnavBallinCode: our website doesnt support dark theme yet

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +16,424 @@ <h3></h3>
...
+    position: absolute;
+    left: 0;
+    top: 50%;
+    width: 100%;
+    height: 1px;
+    background: #ddd;
+}
+
+.checkout-login-divider span {
+    background: inherit;
```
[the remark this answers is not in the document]
ArnavBallinCode: our website doesnt support dark theme yet

app/eventyay/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -14,4 +15,421 @@ <h3></h3>
...
+<style>
+.checkout-login-overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    /* Use semi-transparent black for overlay, respects system themes */
+    background: rgba(0, 0, 0, 0.6);
+    z-index: 10000;
```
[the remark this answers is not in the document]
hongquan: Please use lower z-index, or later on we have everything at z-index 10000.

hongquan (approved):

Issue #1526, opened 2025-12-17
Video room link is shared as localhost


hongquan:
**Describe the bug**

When clicking to the "QR code" button in Video room page, a popup appears with a URL of the room for sharing. But the URL is "localhost:8000/..." instead of the real domain.

**Expected behavior**

The URL should match the site domain, like "next.eventyay.com/..."

**Screenshots**

<img width="1154" height="746" alt="Image" src="https://github.com/user-attachments/assets/64a1bf53-6ffa-4b86-a89e-271ff0899a4d" />

**Additional context**

rakshitjain23:
@hongquan  can u aasign this to me.

hongquan:
@rakshitjain23 Just start doing it and create a draft PR to let people know that you are working on it.
We don't explicitly assign to new contributors because they may be busy after being assigned, and other contributors may think that the task is already taken and lose the chance to contribute.

rakshitjain23:
@hongquan okay sure thank you. I'll do my best

rakshitjain23:
@hongquan i opened a pr.

mariobehling:
Solved in https://github.com/fossasia/eventyay/pull/1527

PR #1527, opened 2025-12-17
fix: QR code URL generation and WebSocket robustness

rakshitjain23:
I fixed an issue where the QR code for rooms was always showing localhost instead of the actual site URL. I updated the settings logic so it now correctly uses the configured domain (like test.eventyay.com) if a short URL isn't explicitly set.

Also, I found that if the BigBlueButton server isn't set up (which happens often in local dev), the whole WebSocket connection would crash. This side effect was causing other features, like the QR code generation, to fail or load infinitely. I added a check to handle this error gracefully so the app stays stable even without a video server.

### Result:

QR codes now point to the correct domain instead of localhost.
The app no longer crashes/freezes when checking for the video server.

### Screenshots:
Before:
Screenshot from 2025-12-17 12-43-20
After:
Screenshot from 2025-12-17 13-09-42

#1526

## Summary by Sourcery

Handle missing BigBlueButton configuration more gracefully and align QR code short URL behavior with the primary site URL configuration.

Bug Fixes:
- Prevent WebSocket handlers for BigBlueButton rooms, calls, and recordings from crashing when no BBB server is configured by surfacing a controlled error instead.
- Ensure QR code and other short URL consumers use the main site URL when no explicit short_url is configured.

Enhancements:
- Allow short_url configuration to be optional and automatically default to site_url via a settings model validator.

### DIFF
--- a/app/eventyay/webapp/src/views/schedule/schedule-components/GridSchedule.vue
+++ b/app/eventyay/webapp/src/views/schedule/schedule-components/GridSchedule.vue
@@ -278,7 +278,7 @@ export default {
 			const roomIndex = this.rooms.indexOf(session.room)
 			return {
 				'grid-row': `${getSliceName(session.start)} / ${getSliceName(session.end)}`,
-				'grid-column': roomIndex > -1 ? roomIndex + 2 : null
+				'grid-column': roomIndex > -1 ? roomIndex + 2 : this.rooms.length + 2
 			}
 		},
 		getOffsetTop() {


rakshitjain23:
@ArnavBallinCode hey any changes needed?

rakshitjain23:
@hongquan I've ensured short_url no longer accepts None. I removed the validator and instead explicitly added short_url to both eventyay.development.toml and eventyay.production.toml with appropriate default values.
bbb.py - improvements: I removed the broad try...except IntegrityError blocks and fixed the indentation issues as pointed out.

rakshitjain23:
@mariobehling  any changes needed? or it'll be closed?

hongquan:
Please delete unused import. After that, this PR is OK to merge.

rakshitjain23:
@hongquan ok done i removed the unused import..

Issue #1528, opened 2025-12-17
YouTube video stream: translation stream not applied and minimized player restarts on page navigation


mariobehling:
The YouTube-based video stream in the video component shows two related issues affecting usability and live interpretation.

**Issues**

1. **Translation stream not applied**

   * Organizers can add a translation stream in the UI.
   * This translation stream is not attached or selectable in the live video player.
   * As a result, viewers cannot access the intended interpretation stream.

2. **Minimized stream restarts on page navigation**

   * When a user navigates to another page (e.g. sessions, speakers), the minimized player restarts the stream.
   * Previously, the expected behavior was:

     * The stream minimizes automatically.
     * Playback continues uninterrupted.
     * Restart only occurs when switching to a different video.

<img width="1442" height="1120" alt="Image" src="https://github.com/user-attachments/assets/d70b71c9-4afe-420d-83d0-3dba4456c85a" />

**Expected Behavior**

* Translation streams configured by organizers are available and usable in the video player.
* When navigating between pages that do not change the video source:

  * The player minimizes.
  * Playback continues without restarting.
* Stream restart should occur only when switching to a different video.

**Current Behavior**

* Translation streams are ignored or not attached to the player.
* The minimized player restarts playback on every page navigation.

**Tasks**

* Investigate and fix application of organizer-configured translation streams for YouTube videos.
* Persist player state across internal page navigation.
* Prevent reinitialization of the player when the video source has not changed.
* Ensure switching to a different video still triggers a proper restart.
* Test behavior across:

  * Page navigation within the event
  * Minimized player mode
  * Events with and without translation streams

rakshitjain23:
@mariobehling can you assign this to me.

RakshithaMH24:
Hi @mariobehling 
I’d like to work on this issue by addressing the player lifecycle and translation stream handling separately, ensuring the player is not reinitialized unless the video source changes.  
I’ll start by reproducing both issues locally and share findings before opening a PR.

norbusan:
This has been fixed already

PR #1532, opened 2025-12-17
fix(video): Fix YouTube translation streams and player navigation persistence (#1528)

norbusan:
## Summary
Fixed two critical issues with YouTube video streaming:
1. Translation streams configured by organizers were not being applied to the player
2. Minimized player was restarting on page navigation instead of maintaining state

### Translation Stream Fix
- Changed event emission from $root.$emit to Vuex store commit in views/rooms/item.vue
- Removed event bus listener code from MediaSource.vue
- Simplified initializeIframe() by removing unused mute parameter
- Fixed field name consistency: changed 'url' to 'youtube_id' for Original language option
- Existing Vuex watcher for youtubeTransUrl now properly triggers iframe rebuild

### Navigation Persistence Fix
- Added previousRoomId tracking in views/rooms/item.vue
- Modified initializeLanguages() to only reset translation URL when room ID actually changes
- Player state is now preserved when navigating within the same room (e.g., between sessions, speakers)
- Existing isEqual check in module watcher prevents unnecessary reloads

## Test Plan
- [x] Translation streams are now accessible and selectable in the video player
- [x] Minimized playback persists across internal page navigation
- [x] Player only restarts when switching to different video content or rooms
- [x] Tested across multiple scenarios with/without translation streams

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Fix YouTube livestream translation handling and preserve player state when navigating within a room.

Bug Fixes:
- Ensure YouTube translation audio selection is applied via Vuex so the main player rebuilds with the correct translation stream.
- Preserve minimized YouTube player state when navigating within the same room so playback no longer restarts unnecessarily.
- Reset translation state only when the user actually switches rooms, avoiding unwanted player reloads on internal navigation.

Enhancements:
- Simplify the YouTube iframe initialization flow by removing unused mute logic and the secondary translation iframe.
- Replace the legacy event-bus–based language change handling with a Vuex-based approach for more predictable state management.

### DIFF
--- a/app/eventyay/webapp/src/views/rooms/item.vue
+++ b/app/eventyay/webapp/src/views/rooms/item.vue
@@ -104,6 +104,10 @@ export default {
 		}
 		this.initializeLanguages()
 	},
+	beforeUnmount() {
+		// Reset previousRoomId to avoid leaking stale room IDs if component is reused
+		this.previousRoomId = null
+	},
 	methods: {
 		changedTabContent(tab) {
 			if (tab === this.activeSidebarTab) return


norbusan:
Thanks for the review! I've addressed the feedback:

## 1. Property shape consistency

The `youtube_id` property is actually consistent across all language options. The original PR already fixed this by changing the 'Original' language entry from `{language: 'Original', url: ''}` to `{language: 'Original', youtube_id: null}`.

All `languageUrls` entries are created in the admin stage editor with the structure `{language: '', youtube_id: ''}` (see `views/admin/rooms/types-edit/stage.vue:179`), and the `AudioTranslationDropdown` component expects and uses the `youtube_id` property (line 38).

So the property shape is now consistent across all language options.

## 2. previousRoomId lifecycle management

Good catch! I've added a `beforeUnmount` lifecycle hook to reset `previousRoomId` to `null` when the component is destroyed. This prevents potential stale room ID leaks if the component is reused across different navigations.

Changes pushed in commit dcc08c15e.

PR #1539, opened 2025-12-17
Let Celery run in "development" environment by default

hongquan:
Currently, in development, our developers have to use long command like this to run Celery:

```sh
EVY_RUNNING_ENVIRONMENT=development celery -A eventyay worker -l info
```

It is not convenient, let set "development" as default, so that we only need to run:

```sh
celery -A eventyay worker -l info
```

## Summary by Sourcery

Enhancements:
- Change the EVY_RUNNING_ENVIRONMENT default from production to development to simplify local Celery and app execution.

### DIFF
--- a/app/eventyay/__main__.py
+++ b/app/eventyay/__main__.py
@@ -2,7 +2,7 @@ import os
 import sys
 
 if __name__ == '__main__':
-    os.environ.setdefault('EVY_RUNNING_ENVIRONMENT', 'production')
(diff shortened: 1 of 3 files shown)


norbusan (approved):
Looks fine, but we need to make sure that we update the deployments for this ... because now we need to set EYY_RUNNING_ENVIRONMENT explicitely for the deployments.

Does that work with .env files? That is, is it enough to put
```
EYY_RUNNING_ENVIRONMENT=production
```
into a .env file for docker-compose?

PR #1540, opened 2025-12-17
Remove VIDEO_SERVER_URL need by using organizer/event slug

norbusan:
## Summary by Sourcery

Use event and organizer slugs to construct the video creation endpoint URL instead of relying on a separate video server base URL.

Enhancements:
- Extend event video world creation to include organizer slug in event data and URL construction.
- Refine video world creation requests to target a per-event URL derived from SITE_URL, organizer slug, and event slug rather than a global video server hostname.

### DIFF
--- a/app/eventyay/eventyay_common/tasks.py
+++ b/app/eventyay/eventyay_common/tasks.py
@@ -80,6 +80,7 @@ def create_world(self, is_video_creation: bool, event_data: dict) -> Optional[di
     :param is_video_creation: A boolean indicating whether the user has chosen to add a video.
     :param event_data: A dictionary containing the following event details:
         - id (str): The unique identifier for the event.
+        - organizer_slug (str): The slug of the event organizer.
         - title (str): The title of the event.
         - timezone (str): The timezone in which the event takes place.
         - locale (str): The locale for the event.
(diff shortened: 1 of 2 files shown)


app/eventyay/eventyay_common/tasks.py
```diff
@@ -91,10 +92,11 @@ def create_world(self, is_video_creation: bool, event_data: dict) -> Optional[di
     - The user must choose to create a video.
     """
 
-    def _create_world(payload: dict, headers: dict) -> Optional[dict]:
+    def _create_world(payload: dict, headers: dict, organizer_slug: str, event_slug: str) -> Optional[dict]:
         try:
+            video_url = f"{settings.SITE_URL}/{organizer_slug}/{event_slug}/video/api/v1/create-world/"
```
hongquan, line 97: I don't find "create-world" URL to be defined or registered any where, even in original [venueless](https://github.com/venueless/venueless).

But, given the fact that now Video is in the same source base as Ticket and Talk, it doesn't make sense to create 
"world" by calling HTTP API, we should doing so by just import Python function and call. I don't know if this function has been defined, you can make this whole `create_world()` task empty, and we will implement it later.

hongquan (changes requested):

norbusan:
Closed in favor of https://github.com/fossasia/eventyay/pull/1549

Issue #1565, opened 2025-12-21
Remove unused “Redirection for custom checkout” option from Ticketing settings


mariobehling:
The ticketing component still exposes the setting **“Redirection for custom checkout”**, which is no longer used by eventyay. Keeping this option creates confusion and suggests functionality that is not supported or maintained.

This setting should be removed from the ticketing settings UI and any related configuration.
<img width="2016" height="1312" alt="Image" src="https://github.com/user-attachments/assets/a44ca479-b539-4e68-b4e8-3d290c1301a6" />

**Expected Behavior**

* The **“Redirection for custom checkout”** option is no longer visible in the ticketing settings.
* Ticketing configuration reflects only actively supported features.
* No references to custom checkout redirection remain in the UI.

**Current Behavior**

* An unused setting for custom checkout redirection is still shown in the ticketing settings section.

**Tasks**

* Remove the “Redirection for custom checkout” option from the ticketing settings UI.
* Remove or deprecate related backend configuration if no longer used.
* Ensure there are no remaining references in templates, documentation, or help texts.
* Verify that ticket checkout behavior is unaffected after removal.

rakshitjain23:
@mariobehling i opened a pr.

amreshanand:
@mariobehling I have reviewed this issue. Please assign this issue to me.

Saksham-Sirohi:
closing as issue is no longer relevant as this feature is no longer enabled by default as said by @mariobehling https://github.com/fossasia/eventyay/pull/1569#issuecomment-3679074062 here please reopen if removal is needed

Issue #1567, opened 2025-12-21
Plugins are active even when shown as disabled in Ticketing plugin settings


mariobehling:
On the **Ticketing component plugin page**, plugins appear as **disabled** in the UI. However, in practice **all plugins are active and their functionality is applied**, regardless of their displayed state.

This creates a mismatch between what organizers see in the plugin settings and the actual system behavior.

**Expected Behavior**

* Plugins that are shown as **disabled** in the ticketing plugin settings are:

  * Not loaded
  * Not executed
  * Do not affect ticketing behavior
* Only plugins explicitly enabled should be active.

**Current Behavior**

* Plugins appear disabled in the UI.
* Despite this, their functionality is active and applied.
* Plugin enable or disable state in the UI has no real effect.

**Impact**

* Organizers cannot reliably control plugin behavior.
* Disabled plugins may unintentionally modify ticketing logic or UI.
* The plugin settings page is misleading.

**Tasks**

* Verify how plugin enable or disable state is stored and evaluated.
* Ensure the runtime behavior respects the configured plugin state.
* Fix plugin loading logic so disabled plugins are not executed.
* Align plugin UI state with actual plugin activation state.
* Test with:

  * All plugins disabled
  * Single plugin enabled
  * Multiple plugins enabled
* Confirm behavior consistency across ticketing, talk, and video components.

PR #1569, opened 2025-12-21
feat: Remove "Redirection for custom checkout" feature

rakshitjain23:
## Overview
Removes the "Redirection for custom checkout" feature (implemented in the returnurl plugin) as it is no longer used or supported. This action cleans up the Event Settings UI and the codebase.

## Changes

- Configuration: Removed eventyay.plugins.returnurl from INSTALLED_APPS in prod, dev, and test settings.

- Code Removal: Deleted the full plugin directory app/eventyay/plugins/returnurl/.

- Tests: Deleted associated tests in src/tests/plugins/test_returnurl.py .

- Documentation: Removed the guide doc/api/guides/custom_checkout.rst .

- Packaging: Updated  MANIFEST.in to exclude the deleted plugin assets.

## Verification

- Checked INSTALLED_APPS ensuring no lingering references.

- Verified make local build (manually or via CI).

- Confirmed "Redirection" menu item is removed from the Event Settings sidebar.

## Screenshots
Settings Menu After Removal
Screenshot from 2025-12-21 21-21-10

#1565

## Summary by Sourcery

Remove the deprecated custom checkout redirection plugin and its configuration references.

Enhancements:
- Clean up project settings by dropping the unused return URL plugin from all INSTALLED_APPS configurations.
- Remove the obsolete return URL plugin code, templates, tests, and related packaging entries.
- Delete outdated custom checkout documentation associated with the removed plugin.

Build:
- Update MANIFEST configuration to reflect removal of return URL plugin assets.

Documentation:
- Remove the custom checkout guide that documented the deprecated redirection feature.

Tests:
- Delete tests specific to the removed return URL plugin.


mariobehling:
After looking further into this, I think this issue is actually related to https://github.com/fossasia/eventyay/issues/1567. All plugins are activated and this is the reason that this plugin is active. So the way forward in this case would be rather to figure out why all plugins are activated and simply deactivate the plugin.

rakshitjain23:
@mariobehling Thanks for pointing that out. Agreed - this seems to be related to #1567. I’ll look into why all plugins are activated and work on a proper fix there.

PR #1573, opened 2025-12-21
Fix plugin enable/disable issue 

### LINKED ISSUE #1567: Plugins are active even when shown as disabled in Ticketing plugin settings

ArnavBallinCode:
## Fix: Plugin Enable/Disable State Now Properly Enforced

Fixes #1567

---

## Problem

Plugins were shown as **disabled in the UI** but continued to remain **active at runtime**, still responding to signals.

For example:
- `eventyay.plugins.badges`
- `eventyay.plugins.ticketoutputpdf`

Both continued to function even when marked as disabled in the plugin settings, leading to a mismatch between UI state and actual system behavior.

---

## Root Cause

The issue stemmed from a logic flaw in `EventPluginSignal._is_active()`.

Plugins listed under `CORE_MODULES` bypassed the enable/disable check entirely due to an OR short-circuit in the condition:

```python
if core_module or (sender and app and app.name in sender.get_plugin_modules()):
```

## How It Works Now 

```
Signal Sent (e.g., register_ticket_outputs)
    ↓
For each registered receiver:
    ↓
1. Get receiver's module path → 'eventyay.plugins.badges.signals'
    ↓
2. Is it in CORE_MODULES? → YES (badges is in CORE_MODULES)
    ↓
3. Resolve to Django app → BadgesApp
    ↓
4. Get event.available_plugins → {'eventyay.plugins.badges': <Meta>, ...}
    ↓
5. Is 'eventyay.plugins.badges' in available_plugins? → YES
    ├─→ It's a PLUGIN! Check if enabled
    ↓
6. Get event.get_plugins() → ['eventyay.plugins.ticketoutputpdf']
    ↓
7. Is 'eventyay.plugins.badges' in enabled list?
    ├─→ NO → Return False →  Receiver NOT called
    ├─→ YES → Check exclusions & compatibility
              ↓
              Return True →  Receiver IS called
```

**For pure core modules:**
```
Signal Sent
    ↓
1. Get module path → 'eventyay.base.signals'
    ↓
2. Is in CORE_MODULES? → YES
    ↓
3. Resolve to app → eventyay.base
    ↓
4. Is in available_plugins? → NO (not a plugin!)
    ├─→ It's a CORE MODULE!
    ↓
5. Return True → Always active
```

## Summary by Sourcery

Ensure signal receivers from plugins only run when the plugin is actually enabled while still allowing core modules to remain active as expected.

Bug Fixes:
- Fix plugins marked as disabled in the UI continuing to handle signals at runtime by enforcing plugin enable/disable state in signal dispatching.

Enhancements:
- Introduce shared helper logic and cached app resolution to consistently determine whether a receiver belongs to an active plugin or core module across event and common signal handling.

### DIFF
--- a/app/eventyay/base/services/bbb.py
+++ b/app/eventyay/base/services/bbb.py
@@ -269,6 +269,7 @@ class BBBService:
         scheme = (
             "http://" if settings.DEBUG else "https://"
         )  # TODO: better determinator?
+        domain = self.event.domain or settings.SITE_NETLOC
         return get_url(
             "join",
             {
@@ -288,7 +289,7 @@ class BBBService:
                     else "false"
                 ),
                 "userdata-bbb_custom_style_url": scheme
-                + self.event.domain
+                + domain
                 + reverse("live:css.bbb"),
                 "userdata-bbb_show_public_chat_on_login": "false",
                 # "userdata-bbb_mirror_own_webcam": "true",  unfortunately mirrors for everyone, which breaks things
@@ -329,6 +330,7 @@ class BBBService:
         scheme = (
             "http://" if settings.DEBUG else "https://"
         )  # TODO: better determinator?
+        domain = self.event.domain or settings.SITE_NETLOC
         return get_url(
             "join",
             {
@@ -339,7 +341,7 @@ class BBBService:
                 "password": create_params["moderatorPW"],
                 "joinViaHtml5": "true",
                 "userdata-bbb_custom_style_url": scheme
-                + self.event.domain
+                + domain
                 + reverse("live:css.bbb"),
                 "userdata-bbb_show_public_chat_on_login": "false",
                 # "userdata-bbb_mirror_own_webcam": "true",  unfortunately mirrors for everyone, which breaks things
--- a/app/eventyay/cfp/flow.py
+++ b/app/eventyay/cfp/flow.py
@@ -13,6 +13,7 @@ from django.core.files.uploadedfile import UploadedFile
 from django.db.models import Q
 from django.forms import ValidationError
 from django.http import HttpResponseNotAllowed
+from django.utils.datastructures import MultiValueDict
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.utils.functional import Promise, cached_property
@@ -219,14 +220,40 @@ class FormFlowStep(TemplateFlowStep):
         return copy.deepcopy({**initial_data, **previous_data})
 
     def get_form(self, from_storage=False):
-        if self.request.method == 'GET' or from_storage:
+        # Cache form initial data to avoid repeated work
+        form_initial = self.get_form_initial()
+
+        if self.request.method == 'GET':
+            # For initial GET requests, use an unbound form populated from session data
+            # This shows saved values but intentionally does not display validation errors
             return self.form_class(
-                data=self.get_form_initial() if from_storage else None,
-                initial=self.get_form_initial(),
+                data=None,
+                initial=form_initial,
+                files=None,
+                **self.get_form_kwargs(),
+            )
+        if from_storage:
+            # For validation checks, create a bound form with session data
+            return self.form_class(
+                data=form_initial,
+                initial=form_initial,
                 files=self.get_files(),
                 **self.get_form_kwargs(),
             )
-        return self.form_class(data=self.request.POST, files=self.request.FILES, **self.get_form_kwargs())
+        # For POST requests, merge new uploads with existing session files
+        # This allows users to navigate back without losing previously uploaded files
+        session_files = self.get_files() or {}
+
+        # Preserve MultiValueDict semantics for proper multi-file field support
+        files = MultiValueDict()
+        # Add session files first
+        for field, file_obj in session_files.items():
+            files[field] = file_obj
+        # For each field, new uploads completely replace any existing session files
+        for field, file_list in self.request.FILES.lists():
+            files.setlist(field, file_list)
+
+        return self.form_class(data=self.request.POST, files=files, **self.get_form_kwargs())
 
     def is_completed(self, request):
         self.request = request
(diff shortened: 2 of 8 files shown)


Saksham-Sirohi:
@ArnavBallinCode, can you please add a video showcasing the changes

ArnavBallinCode:
https://github.com/user-attachments/assets/fde9eb5b-09a5-46be-876d-1a3b3d49225c



> @ArnavBallinCode, can you please add a video showcasing the changes

app/eventyay/base/signals.py
```diff
@@ -18,6 +19,85 @@ def _populate_app_cache():
...
+        core_module: Boolean indicating if this is a core module by path
+        excluded_plugins: List of excluded plugin names
+        get_plugin_list_callable: Callable that accepts sender and returns plugin list
+                                  (e.g., lambda s: s.get_plugins() or lambda s: s.plugin_list)
+    
+    Returns:
+        Boolean indicating if the receiver should be active
+    """
+    # Track plugin metadata when we have a sender and resolved app
+    if sender and app:
```
norbusan, line 70: The previous implementation returned `True` for core modules when `sender=None`.
The logic here (bottom `return False`) will return `False` for this case.

Is this on purpose?

Should we add something like:

```
if core_module and sender is None:
    return True
```
to get back the previous behavior?
`register_notification_types` for example documents that sender=None is possible.

app/eventyay/common/signals.py
```diff
@@ -34,26 +35,19 @@ def get_live_receivers(self, sender):
...
-            while True:
-                app = app_cache.get(searchpath)
-                if '.' not in searchpath or app:  # pragma: no cover
-                    break
-                searchpath, _ = searchpath.rsplit('.', 1)
-            return app and app.name in sender.plugin_list
-        return False
+        module_path = receiver.__module__
+        core_module = any(module_path.startswith(cm) for cm in settings.CORE_MODULES)
+
```
norbusan, line 35: The previous code removed had a fast-track line:
```
        if sender and not sender.plugin_list:
            return False
```
Now we call the possibly heavy `_resolve_app_for_module` on each signal.

Should we re-introduce a short-cut here, too?
Not sure the following works, but something along these lines:
```
      # Fast path for events without plugins
      if sender and not sender.plugin_list:
          module_path = receiver.__module__
          if not any(module_path.startswith(cm) for cm in settings.CORE_MODULES):
              return False
```

app/eventyay/common/signals.py
```diff
@@ -10,6 +10,7 @@
 from django.dispatch.dispatcher import NO_RECEIVERS
 
 from eventyay.base.models import Event
+from eventyay.base.signals import _resolve_app_for_module, _check_plugin_active
 
 app_cache = {}
```
norbusan: Is this still used? I don't think so. Better remove it.

hongquan (changes requested):
Don't name the function with "_" prefix.

app/eventyay/base/signals.py
ArnavBallinCode: Yup, Ill add this

app/eventyay/common/signals.py
ArnavBallinCode: The import line: `from eventyay.base.signals import _resolve_app_for_module, _check_plugin_active`
is definitely needed and actively used. The `_is_active()` method calls both of these functions.

However, if you're referring to `app_cache` (which was on line 15-22), then yes, Thats unused and redundant now . Ill remove that

app/eventyay/common/signals.py
ArnavBallinCode: yup,
Instead of checking in _is_active(), Ill add the fast path inside the shared helper

```
if sender and app:
    plugin_list = get_plugin_list_callable(sender)
    if not plugin_list:
        if core_module:
            return True
        return False
```

app/eventyay/common/signals.py
```diff
@@ -34,26 +28,19 @@ def get_live_receivers(self, sender):
...
-            app = None
-            while True:
-                app = app_cache.get(searchpath)
-                if '.' not in searchpath or app:  # pragma: no cover
-                    break
-                searchpath, _ = searchpath.rsplit('.', 1)
-            return app and app.name in sender.plugin_list
-        return False
+        module_path = receiver.__module__
+        core_module = any(module_path.startswith(cm) for cm in settings.CORE_MODULES)
```
hongquan: Rename to `is_core_module`.

hongquan (changes requested):

hongquan (approved):

Saksham-Sirohi (approved):

PR #1577, opened 2025-12-21
Fix BBB live URL routing and domain fallback

suhailnadaf509:
**Context:**
BBB join flows rely on `live` routes (e.g. the custom CSS endpoint), but these URLs were not included in the main domain URL configuration. In addition, BBB join URL generation assumed `event.domain` was always set, which caused broken `userdata-bbb_custom_style_url` values when an event had no domain.

**What changed:**

* Added `eventyay.features.live.urls` to the main domain URL patterns so `live:` routes resolve correctly.
* Updated BBB join URL generation to use `event.domain` with a fallback to `settings.SITE_NETLOC`.

**Why this works:**
With `live` routes now exposed and a safe domain fallback in place, BBB’s HTML5 client can correctly load the custom CSS endpoint during room and call joins.

**Files changed:**

* `app/eventyay/base/services/bbb.py`
* `app/eventyay/multidomain/maindomain_urlconf.py`

## Summary by Sourcery

Ensure BigBlueButton (BBB) join flows correctly resolve live routes and safely handle events without a custom domain.

Bug Fixes:
- Fix BBB join URL generation by falling back to the global site domain when an event domain is not set.
- Expose live/BBB-related routes (such as the custom CSS endpoint) in the main domain URL configuration so they can be resolved correctly.

Enhancements:
- Centralize live feature URL patterns into the main domain URL configuration for BBB and related video features.


norbusan:
@suhailnadaf509 can you give an example how to test this PR?

Issue #1621, opened 2025-12-29
Tests for badge APIs assume fixed ordering and exact response structure


td15:
While reviewing the badge API tests, I noticed that some tests rely on exact
equality checks of the full API response and implicitly assume a fixed ordering
of results.

For example, in `test_api_list`, the test asserts full equality on:

    r['results'] == [res]

and later relies on `.first()` for item assignments. This assumes:
- deterministic ordering of API results
- presence of only a single badge layout/item
- no additional fields being added to the API response in the future

Since API ordering is not guaranteed unless explicitly defined, these assumptions
can make the tests fragile and lead to false failures across database backends
or future extensions.

A more robust approach could be:
- asserting the length of results
- validating key fields (IDs, names) instead of full object equality
- explicitly sorting results before comparison if ordering is important

If this aligns with the project’s expectations, I’d be happy to submit a PR
to make the tests more resilient.

Pradyuman-aviator:
Hi @td15, thanks for pointing this out..... this makes sense.

I agree that relying on exact response equality and implicit ordering can make these tests fragile, especially across different DB backends or future API changes.

I’d like to take this up and work on making the badge API tests more robust by:

avoiding full object equality where not required

asserting key fields (id, name, etc.)

and sorting results explicitly only where ordering is intentional

I’ll keep the changes minimal and aligned with existing test expectations.
Please let me know if this approach works before I submit a PR

td15:
Hi @its-me-meax  , I have already raised a PR for this issue. You may look at other issues.

Saksham-Sirohi:
the tests are currently non working and just recently moved so this is currently out of scope for us

PR #1626, opened 2025-12-31
Make badge API tests robust against ordering assumptions

td15:
This PR follows up on the discussion in #1621.

It improves the robustness of badge API list tests by avoiding full object
equality, asserting only key fields, and removing implicit ordering assumptions.

Changes are kept minimal and aligned with existing test expectations.

## Summary by Sourcery

Relax badge API list test assertions to avoid relying on full object equality and result ordering.

Tests:
- Update badge layout list tests to assert only key fields and counts instead of full response objects, removing ordering assumptions.
- Adjust badge item list tests to focus on critical fields and presence checks rather than exact object matches.

### DIFF
--- a/src/tests/plugins/badges/test_api.py
+++ b/src/tests/plugins/badges/test_api.py
@@ -55,7 +55,7 @@ def test_api_list(env, client):
         ).content.decode('utf-8')
     )
 
-    # ✅ robust assertions (no ordering / no full equality)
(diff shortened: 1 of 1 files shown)


td15:
Thanks for the review!

I’ve simplified the inline comments to keep the test style consistent and avoid visual noise. I’ve kept the assertions explicit for readability, but I’m happy to refactor into helpers if maintainers prefer that approach.

Issue #1455, opened 2025-12-08
Convert webcheckin app from Vue 2 to Vue 3


hongquan:
There is one plugin which is using Vue 2 code:

```sh
❯ rg -F .vue -t html
eventyay/plugins/webcheckin/templates/pretixplugins/webcheckin/index.html
47:    <script type="text/vue" src="{% static 'pretixplugins/webcheckin/components/checkinlist-item.vue' %}"></script>
48:    <script type="text/vue" src="{% static 'pretixplugins/webcheckin/components/checkinlist-select.vue' %}"></script>
49:    <script type="text/vue" src="{% static 'pretixplugins/webcheckin/components/searchresult-item.vue' %}"></script>
50:    <script type="text/vue" src="{% static 'pretixplugins/webcheckin/components/datetimefield.vue' %}"></script>
51:    <script type="text/vue" src="{% static 'pretixplugins/webcheckin/components/datefield.vue' %}"></script>
52:    <script type="text/vue" src="{% static 'pretixplugins/webcheckin/components/timefield.vue' %}"></script>
53:    <script type="text/vue" src="{% static 'pretixplugins/webcheckin/components/app.vue' %}"></script>
```

Because Vue 2 is end-of-life, let's convert it to Vue 3.

This app is currently built with Django-Compressor + Rollup. The https://github.com/fossasia/eventyay/pull/1151 will move the location of the build tool to _eventyay/frontend/webcheckin-bundler_.

Saksham-Sirohi:
the pr #1634 gets the component working, will start working on migrations as soon as this is merged

PR #1634, opened 2026-01-03
chore: got webcheckin working

Saksham-Sirohi:
Part of fixing #1455 gets the webcheckin app component working

## Summary by Sourcery

Enable the web check-in frontend to work with the updated API and authentication model, and align check-in logic with the new product-based schema.

Bug Fixes:
- Fix multiple-choice answer binding in the web check-in form so options are correctly detected as selected.
- Correct check-in result display to use product-based naming instead of the deprecated item field.
- Update check-in service logic to use product_id consistently and respect question required flags when enforcing required check-in questions.
- Ensure check-in list filtering uses product-based constraints when validating order positions and auto-creating check-ins on order placement.
- Make web check-in and check-in list selection API calls send session credentials so authenticated API endpoints can be used.
- Restore the public API v1 URL routing so the web check-in app can reach the backend API.

Enhancements:
- Switch REST framework default permission to an event-aware permission class and modernize authentication setup to support sessions, team tokens, device tokens, and OAuth2.

### DIFF
--- a/app/eventyay/base/services/checkin.py
+++ b/app/eventyay/base/services/checkin.py
@@ -93,7 +93,7 @@ class LazyRuleVars:
 
     @property
     def product(self):
-        return self._position.item_id
+        return self._position.product_id
 
     @property
     def variation(self):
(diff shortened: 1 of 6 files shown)


Saksham-Sirohi:
rebuild frontend vue apps using make npminstall then in tickets setting enable webchekin plugin then inside checkin section go to webcheckin section it would now work without showing error code 500 you can see webchekin beta here in screenshot below open that and component will work
Screenshot 2026-01-03 at 5 52 15 PM

norbusan (approved):
LGTM, page loads

PR #1636, opened 2026-01-03
fix: Allow SSO-only users to change email and set password without cu…

### LINKED ISSUE #1394: Email and password cannot be changed for MediaWiki SSO accounts
For users whose accounts were created via **MediaWiki SSO**, it is impossible to change either the **email** or the **password**.

The problem is caused by the account settings form requiring the **“current password”** field to be filled out before any change can be saved.
However, MediaWiki SSO accounts do **not** have a local password, and no password is shown or stored for the user. As a result:

* Users cannot change their email address.
* Users cannot set or update a password.
* The form cannot be submitted because “current password” cannot be provided.

**Expected Behavior**

* MediaWiki SSO accounts should allow email changes without requiring a local password.
* If a password field is shown, it should allow setting a new password **without** requiring a “current password”.
* Alternatively, password fields should be hidden entirely for SSO-only accounts.

**Current Behavior**

* “Current password” is mandatory but cannot be provided.
* Email and password changes are blocked for all MediaWiki SSO accounts.

**Tasks**

* Update profile settings logic to detect SSO-only accounts.
* Remove or bypass “current password” requirement for SSO accounts.
* Allow email updates for SSO accounts without entering a password.
* Allow setting a new password directly (optional) or hide password fields.
* Test flows for:

  * MediaWiki SSO–only accounts
  * Accounts with both SSO and email/password login

Ashutoshx7:
Fixes #1394

Users who logged in via SSO (e.g., MediaWiki) do not have a local password set. Previously, these users were unable to:
- Change their email address (required current password)
- Set a new local password (required current password they don't have)

Changes:
- Modified LoginInfoForm and UserSettingsForm to detect SSO-only accounts using user.has_usable_password()
- For SSO-only accounts:
  - Hide the 'current password' field using HiddenInput widget
  - Skip password validation in clean_old_password()
  - Allow email changes and password setting without current password
- Added informative SSO message in account settings templates
- Regular users with local passwords still require current password

Fixes: SSO users unable to update profile settings

## Summary by Sourcery

Allow SSO-only users to update login details while preserving password verification for local accounts.

Bug Fixes:
- Fix inability for SSO-only users to change their email address or set a local password due to mandatory current-password checks.

Enhancements:
- Detect SSO-only accounts based on whether the user has a usable local password and adjust validation accordingly.
- Hide the current-password field and update password help text for SSO-only accounts in account settings and profile templates.
- Add informational messaging in account settings views to clarify SSO login status and available options for setting a local password.


hongquan:
I feel that this PR is going in wrong direction. You should check django-allauth first to see if has a flow for social account user to set a new password or not.

mariobehling:
Thanks @Ashutoshx7 for the contribution. The PR is reasonable given the original issue description and the effort is appreciated.

After a deeper review, we concluded that we should align more closely with Django Allauth’s intended approach. With that in mind, this implementation goes in a different direction than what we want to pursue moving forward.

Please have a look at the additional context and discussion in the issue comment here:
https://github.com/fossasia/eventyay/issues/1394#issuecomment-3709361024

Thanks again for taking the time to work on this, and we would be happy to see more contributions from you in the future. Closing this PR.

Issue #1661, opened 2026-01-04
BigBlueButton (BBB) integration: broken features & hardening checklist


suhailnadaf509:
## Context

A BigBlueButton (BBB) server was recently set up  and connected to Eventyay. Early testing indicates that several parts of the integration are not yet working as expected(joining rooms, admin settings not applying, recordings, server selection / room moves, etc.).

This issue is a **master tracking issue** to list what’s broken, create focused sub-issues, and fix things incrementally via small PRs.

Relevant code areas:

* Frontend BBB module: `call.bigbluebutton` (`MediaSource.vue`)
* Video Channel admin UI: `channel-bbb.vue`
* Backend BBB service: `bbb.py`
* Admin room move UI: `bbb_moveroom.html`
* Docs: `bbb.rst`

## Expected behavior (high level)

* Video Channels using BBB reliably resolve join URLs and create meetings.
* Admin-configured settings actually apply in BBB.
* Server selection / room moves are validated and safe.
* Recordings are retrievable and visible when permitted.
* Failures are observable (clear UI errors + backend logs).

## Actual behavior

TBD — to be filled as we reproduce issues during testing (see checklist).

### A) Connectivity & join flow

* [ ] Join URL works for rooms (`bbb.room_url`)
* [ ] Private calls (`bbb.call_url`) work where applicable
* [ ] Missing profile handled cleanly (`bbb.join.missing_profile`)
* [ ] Clear UI errors for `bbb.failed` (server down vs config vs permission)

### B) Video Channel settings actually apply

Verify end-to-end wiring for settings defined in `channel-bbb.vue`:

* [ ] Recording
* [ ] Hide presentation
* [ ] Waiting room
* [ ] Auto mic / auto cam
* [ ] Auto-mute users
* [ ] Disable cam/chat for non-mods
* [ ] Voice bridge
* [ ] Prefer server
* [ ] Initial presentation upload

### C) Server selection & room moving

* [ ] `prefer_server` is respected
* [ ] Clear error when no active BBB servers exist
* [ ] “Move room to new server” works end-to-end:

  * validates room
  * handles kicking users safely
  * updates server mapping
  * avoids / documents split-room cases
* [ ] Cost update job (`bbb_update_cost`) documented and stable

### D) Recordings

* [ ] Backend fetches recordings correctly
* [ ] UI displays recordings with proper permissions
* [ ] Clear errors for no recordings / server down
[...]

PR #1678, opened 2026-01-06
Bbb ensure display name

singhmahipal:
This PR fixes backend handling of BBB joins when a user profile or display name
is missing.

It ensures a safe, non-empty display name is always passed to BBB for both
room joins and private calls, and avoids unsafe access to user.profile.

Addresses checklist item A) Connectivity & join flow → bbb.join.missing_profile
from the BBB hardening tracker (#1661).

## Summary by Sourcery

Ensure BigBlueButton join requests always use a safe, non-empty display name and avoid unsafe access to user profile data.

Bug Fixes:
- Guarantee a fallback display name for BBB room joins and private calls when user profile or display name is missing.
- Prevent attribute errors when accessing avatar URLs by safely handling missing or null user.profile on BBB joins.

### DIFF
--- a/app/eventyay/base/services/bbb.py
+++ b/app/eventyay/base/services/bbb.py
@@ -13,6 +13,7 @@ from django.db.models import Count, F, OuterRef, Q, Subquery, Value
 from django.db.models.functions import Coalesce
 from django.urls import reverse
 from django.utils.html import escape
+from django.utils.translation import gettext_lazy as _
 from lxml import etree
 from yarl import URL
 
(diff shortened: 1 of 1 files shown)


mariobehling:
Thank you for the contribution.

We cannot accept this PR as is, and we are closing it. This change touches the BBB join flow, but the PR does not provide evidence of end to end testing, reproducible steps, or any screenshots or short walkthrough showing the user facing behavior. Without that, it is difficult for reviewers to validate that the join flow is actually fixed and does not introduce regressions.

In addition, the PR appears largely AI edited and the core logic introduces questionable fallbacks such as a hard coded “Guest” label that is not translatable, plus an unsafe email local part split that can yield empty or invalid display names. The PR also mixes in a documentation change that is not clearly tied to the bug fix, which further increases review overhead.

If you want to continue, please open a new PR that includes:

* Clear reproduction steps and what is fixed
* Evidence of manual testing for both room joins and private calls
* Screenshots or a short walkthrough
* Address the fallback name i18n and input validation concerns
* Keep documentation changes separate unless they are strictly required for the fix

PR #1684, opened 2026-01-06
Let Wikimedia users set new password

hongquan:
Resolves part of #1394 (for password only)

For users who signed up via Wikimedia, hence, doesn't have password, we will show a form for them to start the flow of setting password. This flow is made by django-allauth:

image

This is a page that user will be brought to, after clicking the link in the email. It looks poor because it is using the default HTML provided by django-allauth, we can customize later.

Screenshot From 2026-01-06 19-01-16

## Summary by Sourcery

Allow accounts created via social login without a usable password to request a password setup link from the account settings page and adjust project metadata and development email/debug configuration.

New Features:
- Expose a password setup request form in account settings for users whose accounts were created via social login and currently lack a usable password.

Enhancements:
- Hide password change fields and relax current-password validation when a password reset is required so that users without a password can still update their profile.
- Show provider-specific messaging in account settings explaining why a password is missing and guiding users to request a password setup link.
- Tighten the filename format for file-based email backend output to be more compact and collision-resistant, and configure a dedicated directory for development emails.
- Simplify settings and i18n imports by using Django settings directly and cleaning up unused modules.

Build:
- Update project metadata, supported Python versions, and dependency sets, moving various tooling libraries from runtime to docs/test extras and bumping some package versions.

### DIFF
--- a/app/eventyay/eventyay_common/views/account/basic.py
+++ b/app/eventyay/eventyay_common/views/account/basic.py
@@ -27,15 +27,16 @@ logger = getLogger(__name__)
 PASSWORD_RESET_INTENT = 'password_reset'
 
 
-def password_reset_provider(user: User) -> str | None:
-    social = SocialAccount.objects.filter(user=user).order_by('-pk').first()
-    return social.provider if social else None
-
-
-def password_reset_provider_label(user: User) -> str:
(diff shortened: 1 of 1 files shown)


hongquan:
Tested on https://dev.eventyay.com, it works.

Issue #1688, opened 2026-01-06
Bug: Unable to submit session after completing CFS form


hpdang:
We are unable to submit a session proposal after filling out the Call for Speakers (CFS) form. The form can be completed, but when clicking submit, it shows HTTP 500 error. 

### Steps to Reproduce
- Open the Call for Speakers: https://next.eventyay.com/fossasia/exhibition/cfp 
- Fill in all required fields in the session submission form
- Click Submit, it goes to https://next.eventyay.com/fossasia/exhibition/submit/e2ucop/profile/ 

<img width="1879" height="872" alt="Image" src="https://github.com/user-attachments/assets/e71cb7ec-c0e2-415d-bd39-9471574becb9" />

tusharbhatt7:
Hi! I’d like to work on this issue.
I have gone through the problem and will start by reproducing the HTTP 500 error and checking the failing endpoint and logs.

tusharbhatt7:
@mariobehling @hpdang, tagging you for review as this relates to a priority/urgent issue.

Saksham-Sirohi:
Closing this as the issue is resolved earlier

PR #1700, opened 2026-01-07
fix(cfp): resolve cfp flow

Sak1012:
This PR fixes the cfp flow and solves #1688:
- When there is no name in the account, the user’s submitted name will be automatically picked up and updated.
- When the user has a name in the account, the field will automatically select that name.
- If the user attempts to submit a proposal with a different name, a warning will be displayed indicating a mismatch with the account name and the submission doesn't go through.

### DIFF
--- a/app/eventyay/cfp/flow.py
+++ b/app/eventyay/cfp/flow.py
@@ -295,9 +295,16 @@ class FormFlowStep(TemplateFlowStep):
 
         # For "submit" and "draft" actions, validate as before
         if not form.is_valid():
-            non_field_errors = form.non_field_errors()
-            if non_field_errors:
-                messages.error(self.request, '\n\n'.join(non_field_errors))
+            warning_messages = getattr(form, 'warning_messages', None) or []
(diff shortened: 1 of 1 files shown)


norbusan:
### Code review

Found 1 issue:

1. Field-specific validation errors are silently dropped and won't be displayed to users

The PR changes error handling in flow.py from displaying all form errors to only showing non-field errors. The new name validation in profile.py adds an error to the 'fullname' field using `self.add_error('fullname', ...)`, but the updated error handling only calls `form.non_field_errors()`, which excludes field-specific errors. This means users won't see the name mismatch validation error or any other field-level errors, making the form appear to fail silently.

https://github.com/fossasia/eventyay/blob/d7ef66428dfe1a9581aa76c8d76e8e60025b008f/app/eventyay/cfp/flow.py#L290-L296

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1724, opened 2026-01-08
Make navbar sticky in events public page

### LINKED ISSUE #1706: Make Public Page Top Navigation Menu Sticky
On public event pages, the top navigation menu scrolls away with the page. This makes navigation less accessible, especially on long pages such as the start page, schedule, or sessions lists.

## Expected Behavior

* The top menu on public pages should be sticky and remain fixed at the top while scrolling
* Page content below the menu must remain scrollable and should not be hidden behind the sticky header
* The layout should account for the header height so that the content starts below the sticky menu

Image

## Reference

Take inspiration from fossasia.org, where the top navigation remains visible while the page content scrolls underneath it.

## Scope

* Public facing pages only, for example:

  * Event landing page
  * Schedule
  * Sessions
  * Speakers
  * Tickets pages

## Requirements

* Sticky behavior works across common screen sizes (desktop and mobile)
* No overlap between sticky menu and page content (proper spacing or offset)
* No layout shift or jitter while scrolling
* Dropdown menus and header interactions still work correctly when sticky

## Acceptance Criteria

* Top menu remains visible at the top during scroll on public pages
* Content scrolls normally below the menu without being obscured
* Works on mobile and desktop
* No regressions to navigation interactions, dropdowns, or page layout

ishanmitra:
Fixes #1706 

# Preview

Demo shows Tickets page and Call for Speakers page with sticky navbar tabs implemented. The dropdown menu stays in place, has no jitter, and is responsive.

Note: Video has been updated as per today's commit:

https://github.com/user-attachments/assets/66065d3c-09df-443d-b5d7-5ab0afe29c5e

## Summary by Sourcery

Make the event public page navbar a separate sticky element at the top of the page for improved navigation on tickets and call-for-speakers views.

Enhancements:
- Move header navigation markup into a dedicated <nav> element with its own wrapper to separate it from the main header content.
- Apply sticky positioning and layering styles to the header navbar so tabs and user menu remain visible while scrolling on public pages.

### DIFF
--- a/app/eventyay/common/templates/common/base_public.html
+++ b/app/eventyay/common/templates/common/base_public.html
@@ -130,8 +130,6 @@
                         </div>
                     </div>
                 </div>
-            </header>
-            <nav class="header-navbar">
                 <div class="header-wrapper">
                     <div id="header-tabs">
                         {% block header_tabs %}{% endblock header_tabs %}
@@ -142,7 +140,7 @@
                         </div>
                     </div>
                 </div>
-            </nav>
+            </header>
             <div class="card" id="main-card">
                 {% block alerts %}
                     {% if messages %}
--- a/app/eventyay/static/cfp/css/_layout.css
+++ b/app/eventyay/static/cfp/css/_layout.css
@@ -28,6 +28,9 @@ h1 a {
 }
 
 header {
+  position: sticky;
+  top: -180px;
+
   z-index: 1000;
   height: 220px;
   display: flex;
@@ -39,12 +42,6 @@ header {
     max-height: 140px;
     width: auto;
   }
-}
-
-nav.header-navbar {
-  position: sticky;
-  top: 0;
-  z-index: 1000;
 
   .header-wrapper {
     width: 100%;
@@ -332,6 +329,7 @@ footer {
     }
 
     header {
+      top: -124px;
       padding: 0;
       #event-logo {
         max-height: 120px;
--- a/app/eventyay/static/pretixpresale/scss/custom.css
+++ b/app/eventyay/static/pretixpresale/scss/custom.css
@@ -28,6 +28,16 @@ header {
     word-break: normal;
 }
 
+header .header-wrapper {
+    width: 100%;
+    max-width: calc(100vw - 8px);
+    word-break: keep-all;
+    flex-wrap: nowrap;
+    background: var(--color-bg);
+    border-radius: var(--size-border-radius) var(--size-border-radius) 0 0;
+    margin-bottom: 0;
+}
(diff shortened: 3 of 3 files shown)


Sak1012:
Thank you for the PR, please remove the unnecessary changes made to rename the tags and just add the css to make it sticky.

ishanmitra:
> Thank you for the PR, please remove the unnecessary changes made to rename the tags and just add the css to make it sticky.

@Sak1012 The navigation bar has been migrated from its original `<header>` parent to a `<nav>` just below it, as a pure CSS implementation of sticky did not work:
1. Simply adding the sticky property to the header-bar did not work at all,
2. Using the sticky property on the header made the entire top portion: Event Name, Event Date, and the `header-bar`, stick to the top while scrolling down.

I had originally tried that minimal approach, as you have suggested. All of the migrated divs have the same classes. The only difference is the addition of the nav with class `header-navbar`.

app/eventyay/common/templates/common/base_public.html
```diff
@@ -130,17 +130,19 @@
...
+            <nav class="header-navbar">
                 <div class="header-wrapper">
                     <div id="header-tabs">
                         {% block header_tabs %}{% endblock header_tabs %}
-                    	<div class="header-row-right">
-                    	    {% block header_right %}
-                    	    	{% include "common/fragment_header_user.html" with current_event=current_event %}
-                    	    {% endblock header_right %}
-						</div>
+                        <div class="header-row-right">
```
Saksham-Sirohi, line 136: Get rid of these unnecessary whitespace changes. Also, I don't think we need changes to this file at all; only position sticky seems to be the necessary change here please make sure you are achieving this with minimal code change maybe if needed use a separate CSS class

ishanmitra: Yeah, I can fix the whitespace changes. I'd like to confirm whether I am required to keep the sticky tab within the header section itself. In that case, I can work on a different solution.

Sak1012: Yes please go ahead keep the sticky tab within the header section.

ishanmitra: Sounds great, thanks for the clarity.

ishanmitra:
@Sak1012 Removed the unnecessary code changes, solution is purely CSS.

Saksham-Sirohi:
Please add a video showcasing solution

ishanmitra:
> Please add a video showcasing solution

@Saksham-Sirohi Please check the updated video in the PR description.

PR #1766, opened 2026-01-12
Fix: Anonymize user and unlink social accounts on deletion

### LINKED ISSUE #1393: Accounts created via MediaWiki SSO cannot be deleted (accounts instantly recreated)
User accounts created through **MediaWiki SSO** cannot be fully deleted.
When an account deletion is requested:

* The user’s data and proposals are deleted as expected.
* However, the **account itself is immediately recreated**, containing only:

  * The email
  * The account name
* This makes MediaWiki SSO accounts effectively **undeletable**.

This behavior prevents proper account removal and violates expected user control over their identity.

**Expected Behavior**

* Deleting a MediaWiki SSO account should permanently delete the account.
* After deletion, the account must **not** be automatically recreated.
* If the user logs in again via MediaWiki SSO, a new account may be created normally, but deletion should not trigger auto-recreation.

**Current Behavior**

* Deleting an SSO account removes data, but the account reappears instantly with minimal fields restored.

**Tasks**

* Fix deletion logic so SSO accounts remain deleted until a user voluntarily logs in again.
* Prevent automatic recreation during or immediately after deletion.
* Verify correct deletion behavior for:

  * MediaWiki SSO accounts
  * Accounts linked to multiple authentication methods
* Ensure compliance with expected account removal workflows.

Abhijeet212004:
Fixes #1393

Hi team,

This PR solves the issue where accounts created via MediaWiki SSO were not fully deleting. Currently, when a user requests deletion, the account can be recreated upon the next login because the link to the external provider remains active.

The changes in this PR update the 'soft_delete' method to:

1.  Anonymize User Data: Scrambles the email, full name, and username to ensure the account is truly deactivated.
2.  Unlink Social Accounts: Explicitly deletes the SocialAccount entry so that future logins from MediaWiki are treated as new users rather than resurrecting the old account.

I have added a test case 'app/tests/base/test_issue_1393.py' and verified the fix locally using the Dockerized PostgreSQL database.

Could you please review this? Happy to make any changes if needed.

Thanks!

## Summary by Sourcery

Anonymize user accounts and remove linked social logins when a user is soft-deleted to prevent unintended account resurrection.

Bug Fixes:
- Ensure soft-deleted users have their personal identifiers anonymized to prevent re-login with previous credentials.
- Remove associated social login records on soft delete so external SSO providers no longer resurrect deleted accounts.

Tests:
- Add regression test verifying user anonymization and social account removal when performing a soft delete.

### DIFF
--- a/app/eventyay/base/models/auth.py
+++ b/app/eventyay/base/models/auth.py
@@ -949,17 +949,15 @@ the eventyay team"""
         self.show_publicly = False
         self.profile = {}
 
-        # Anonymize user data to prevent re-login and invalidate session
+        # Anonymize user data to prevent re-login via reused credentials;
+        # session/token invalidation is handled by the authentication/session layer
         self.email = f"deleted_{uuid.uuid4()}@localhost"
-        self.fullname = "Deleted User"
+        self.fullname = str(_("Deleted User"))
         self.wikimedia_username = None
         
         # Delete social accounts to prevent automatic re-association
-        try:
-            from allauth.socialaccount.models import SocialAccount
+        if hasattr(self, "socialaccount_set"):
             self.socialaccount_set.all().delete()
-        except ImportError:
-            pass
 
         self.save()
 
--- a/app/tests/base/test_issue_1393.py
+++ b/app/tests/base/test_issue_1393.py
@@ -29,7 +29,7 @@ def test_soft_delete_anonymization():
     # Verify Anonymization
     assert user.deleted is True
     assert "deleted_" in user.email
-    assert user.fullname == "Deleted User"
+    assert user.fullname == "Deleted User" # This should still pass if locale is EN
     assert user.wikimedia_username is None
     
     # Verify SocialAccount deletion


mariobehling:
Thank you for working on this and for taking the initiative to address the issue. The direction of anonymising users and unlinking social accounts is generally aligned with what we want to achieve, and the fact that you added a test shows the right intent.

However, we are going to close this PR.

In its current form, the PR places too much effort on the reviewer to reason about correctness, edge cases, and long-term maintainability. While parts of the solution move in the right direction, the implementation does not fully or cleanly address the problem end to end, and would require significant refactoring, redesign, and additional testing during review.

In particular:

* The logic is tightly coupled and partially defensive (for example broad exception handling), which makes it hard to verify correctness and can hide real issues.
* The anonymisation approach and values are hard-coded and not clearly defined as part of a consistent deletion strategy.
* Important edge cases are not covered, such as session or token invalidation and ensuring deleted users cannot authenticate again through any provider.
* The test coverage is incomplete and does not fully demonstrate that the original issue is resolved in all relevant scenarios.

What is good:

* The PR correctly identifies the core problem.
* It attempts to solve it at the model level.
* A test was added, which is the right direction.

Next steps:
Please submit a **new PR** that fully and cleanly addresses the issue, rather than iterating on this one. The new PR should:

* Clearly define the intended deletion and anonymisation behaviour.
* Avoid broad exception suppression and incomplete safeguards.
* Cover all relevant edge cases, including authentication, sessions, and social account re-linking.
* Include comprehensive tests that demonstrate the issue is fully resolved.
* Be scoped and structured in a way that allows reviewers to validate correctness without reverse-engineering the implementation.

PR #1794, opened 2026-01-14
Fix: Return 400 error for duplicate check-ins (Fixes #809)

### LINKED ISSUE #809: Clear error message from Check-In API
When the user is checked in there are error messages (e.g. the user is already checked in), it does not prevent them from checking in again. Often the reason why the checkin has failed is not mentioned. Goal: We need displays of error. Also need to take care of things like “User cannot be checked in. User is already checked in.”

Rachit7168:
**Fixes:** #809

**Summary:**
This PR fixes a bug in the `_redeem_process` function where the API was returning a **201 Created (Success)** status even when a check-in failed (e.g., "User already checked in").

**Changes Made:**
* **Modified `app/eventyay/api/views/checkin.py`**:
    * Updated the `except CheckInError` block in `_redeem_process`.
    * **Commented** `Checkin.objects.create(...)` so invalid check-ins are not recorded as valid.
    * Changed the response status code from `201` to `400 Bad Request`.
    * Changed the response body status from `'redeemed'` to `'error'`.
    * Replaced the hardcoded string `'Already checked in'` with the dynamic error code `e.code`.

## Summary by Sourcery

Handle duplicate check-in attempts as client errors and avoid logging them as successful check-ins.

Bug Fixes:
- Return HTTP 400 with an error status and dynamic error code when a duplicate or invalid check-in is attempted instead of incorrectly returning 201 created.
- Prevent duplicate check-in attempts from being recorded as successful check-ins in the database.

Note: This is my second contribution, so I don't have much experience yet. If I made a mistake or have done something wrong, please feel free to correct me—I am eager to learn!

### DIFF
--- a/app/eventyay/api/views/checkin.py
+++ b/app/eventyay/api/views/checkin.py
@@ -688,10 +688,6 @@ def _redeem_process(
                     user=user,
                     auth=auth,
                 )
-                # Checkin.objects.create(    # commented out to avoid double logging
-                #     position=op,
-                #     **common_checkin_args,
-                # )
 
             serializer_context = _setup_context(request, expand, op.order.event, pdf_data, user, auth)
             position_data = CheckinListOrderPositionSerializer(op, context=serializer_context).data
@@ -700,13 +696,13 @@ def _redeem_process(
 
             return Response(
                 {
-                    'status': 'error', # Correct: 'error'
-                    'reason': e.code,  # Correct: dynamic code
+                    'status': 'error', 
+                    'reason': e.code,  
                     'require_attention': op.require_checkin_attention,
                     'position': position_data,
                     'list': MiniCheckinListSerializer(list_by_event[op.order.event_id]).data,
                 },
-                status=400,   # Correct: 400 Bad Request
+                status=400,   
             )
         else:
             serializer_context = _setup_context(request, expand, op.order.event, pdf_data, user, auth)


mariobehling (changes requested):
Were you able to get it running with those code changes? Please provide a small video of the process.

Rachit7168:
Hi @mariobehling ,

Yes, I have verified the changes locally. Attached is a video demonstration of the fix in action.

In the video, you can see the following process:
I attempt a check-in for a ticket that has already been redeemed.
The API now correctly returns an HTTP 400 Bad Request status.
The response body explicitly shows status: "error" and the reason already_redeemed.
The check-in entry is no longer created in the database for this failed attempt, preventing duplicate logs.

https://github.com/user-attachments/assets/279d006f-729c-480e-80d1-41aabb15369c

Since this is my second contribution, I want to ensure I am following the project's standards correctly. Please let me know if I have missed anything or if there is a better way to structure this response. I am eager to learn and happy to make further adjustments if needed.

As this is my second contribution, I want to be transparent that I have been using an AI mentor(Gemini) to help me navigate the codebase, understand the Docker environment, and ensure I am following open-source best practices. This has been a great learning tool for me to understand the technical requirements of this project.

Thank you for your guidance!

Rachit7168:
Hi @mariobehling ,

I wanted to get more information about this particular problem as I have been working on it for past 3 days and I can't figure out the exact problem nor the solution. I have tried Gemini , and chatgpt as well. If I can get more information , It would be a huge help , So I can restart fixing this from that start in the right direction.

Sak1012:
Thank you for the contribution, but this issue is related to the eventyay-checkin app and the error already works in this system.

PR #1796, opened 2026-01-14
Return human-readable error messages for Check-In API

Sushil2k4:
This PR addresses issue #809 by preventing duplicate Checkin creation and returning proper error responses when a user is already checked in.

The API now returns HTTP 400 with both a machine-readable error code and a human-readable message (e.g. “User cannot be checked in. User is already checked in.”) and fixes the audit log `reason_explanation`.


mariobehling:
What is going with the PR title? Please fix it.

Sushil2k4:
Thanks for pointing this out.
I’m cleaning up the PR now by removing the unrelated EventMessage changes and fixing the CheckInError attribute usage.
Will push an update shortly.

PR #1822, opened 2026-01-15
Fix duplicate check-in and return human-readable error

### LINKED ISSUE #809: Clear error message from Check-In API

Sushil2k4:
Fixes #809

When a ticket is already checked in, the API was still returning HTTP 201 and creating duplicate `Checkin` rows.  
This PR fixes that behavior and makes the error response human-readable.

### What this PR does

- Prevents duplicate `Checkin` creation when a ticket is already checked in  
- Returns HTTP 400 instead of 201 for failed check-ins  
- Returns both a machine-readable error code and a human-readable message  
- Logs the correct `reason_explanation` instead of the broken `unkown` value  

### Example response

```json
{
  "status": "error",
  "code": "already_checked_in",
  "reason": "User cannot be checked in. User is already checked in."
}

## Summary by Sourcery

Handle already-checked-in tickets as explicit errors and clean up CI workflows.

Bug Fixes:
- Prevent creation of duplicate check-in records and return an error response when a ticket is already checked in.
- Return a human-readable error message and proper error code in the check-in response and logs instead of an incorrect placeholder value.

CI:
- Remove obsolete GitHub Actions workflows for Docker, docs, strings, style checks, and tests.

### DIFF
--- a/.github/workflows/docker-pr-deploy.yml
+++ b/.github/workflows/docker-pr-deploy.yml
@@ -1,0 +1,34 @@
+name: Docker-pr-deploy
+on:
+  workflow_dispatch:
+    inputs:
+      repository: 
+        description: "Repository name with owner. For example, actions/checkout"
+      ref: 
+        description: "The branch, tag or SHA to checkout."
+
+jobs:
+  push_to_registry:
+    name: Push Docker image to Docker hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ inputs.repository }}
+          ref: ${{ inputs.ref }}
+      
+      - name: Log in to Docker Hub
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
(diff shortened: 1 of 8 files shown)


mariobehling (changes requested):
Please also remove the codecov reference in the Readme.

! Disregard the above. It was made wrongly here !

mariobehling (changes requested):
What is the reason for removing all the yml files in the Github workflows folder? Do not do this!

Sushil2k4:
Restored the GitHub Actions workflows.
The deletion was unintentional while rebasing.
The PR now only contains the Check-In error handling fix.

sneharani1994:
Hey @mariobehling
I am new to this I would like to work on this issue as my first contribution.
Can you assign me.

Sak1012:
Thank you for the contribution. Closing this PR as the intent of this issue needs to be implemented in the [checkin-app](https://github.com/fossasia/eventyay-checkin).

Issue #1200, opened 2025-11-06
Meta: One-Button Deployment Options for Eventyay


mariobehling:
This issue tracks all tasks related to implementing and documenting **one-button deployment options** for Eventyay.
The goal is to enable contributors, partners, and communities to deploy a full Eventyay stack (API, Frontend, Workers, Postgres, Redis, Storage, TURN) with minimal setup through hosted platforms supporting automated deployments.

### 🎯 Objectives

* Provide ready-to-use deployment templates (`render.yaml`, `fly.toml`, `railway.json`, `app.yaml`, etc.) for supported platforms.
* Simplify setup for community events by providing one-click “Deploy” buttons.
* Standardize environment variables and documentation across platforms.
* Exclude **Heroku** and **Koyeb** (limited scaling and pricing flexibility).

#### 1. Render

* [ ] Create `render.yaml` blueprint with services:

  * Web (Django + Gunicorn)
  * Frontend (static build)
  * Worker (Celery)
  * Scheduler (Celery Beat)
  * Postgres + Redis (managed)
* [ ] Configure object storage (S3-compatible) and SMTP.
* [ ] Document setup and variables in `docs/deployment/render.md`.

#### 2. Fly.io

* [ ] Add `fly.toml` and multi-app setup for web, worker, and scheduler.
* [ ] Include Fly Postgres and Redis via Upstash or managed add-ons.
* [ ] Document TURN setup (coturn instance).
* [ ] Write guide in `docs/deployment/flyio.md`.

#### 3. Railway

* [ ] Create `railway.json` for one-click deployment.
* [ ] Define variables for backend, worker, and frontend.
* [ ] Link managed Postgres and Redis add-ons.
* [ ] Add guide `docs/deployment/railway.md`.

#### 4. DigitalOcean App Platform

* [ ] Provide `app.yaml` spec with web and worker components.
* [ ] Use DO Managed Postgres, Redis, and Spaces.
* [ ] Add sample Droplet TURN setup.
* [ ] Document in `docs/deployment/digitalocean.md`.

#### 5. Google Cloud Run

* [ ] Define `cloudrun.yaml` for container deploys.
* [ ] Connect Cloud SQL, Memorystore, and Cloud Storage.
* [ ] Add TURN setup notes using Compute Engine.
* [ ] Document in `docs/deployment/gcp.md`.

#### 6. AWS Elastic Beanstalk / ECS

[...]

PR #1869, opened 2026-01-16
Fix: Add missing whitenoise dependency

Abhijeet212004:
**Description**
Adds `whitenoise` to [pyproject.toml](cci:7://file:///Users/abhijeet/Documents/project3/eventyay/app/pyproject.toml:0:0-0:0) and updates `uv.lock`. 

**Relation to #1200**
This fixes the CI build failure blocking the deployment work for #1200. Without this dependency, collectstatic fails, preventing any production Docker image from building.

**Changes**
- Added `whitenoise>=6.6.0` to [app/pyproject.toml](cci:7://file:///Users/abhijeet/Documents/project3/eventyay/app/pyproject.toml:0:0-0:0)
- Updated [app/uv.lock](cci:7://file:///Users/abhijeet/Documents/project3/eventyay/app/uv.lock:0:0-0:0) via uv lock

## Summary by Sourcery

Build:
- Add whitenoise>=6.11.0 to the application dependencies and update uv.lock accordingly.

### DIFF
--- a/app/pyproject.toml
+++ b/app/pyproject.toml
@@ -210,6 +210,7 @@ dependencies = [
     "webauthn>=2.2.0",
     "webencodings>=0.5.1",
     "websockets>=15.0.1",
+    "whitenoise>=6.11.0",
     "xmlschema>=3.4.3",
     "yarl>=1.22.0",
     "zipp>=3.21.0",
--- a/app/uv.lock
+++ b/app/uv.lock
@@ -1438,6 +1438,7 @@ dependencies = [
     { name = "webauthn" },
     { name = "webencodings" },
     { name = "websockets" },
+    { name = "whitenoise" },
     { name = "xmlschema" },
     { name = "yarl" },
     { name = "zipp" },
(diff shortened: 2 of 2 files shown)


mariobehling:
Issue https://github.com/fossasia/eventyay/issues/1200 is not about failing of collectstatic. Please open an issue and provide additional information, e.g. error logs to showcase the need for this additional dependency.

mariobehling:
Thank you for the PR.

After reviewing this change, we cannot accept it and are closing the PR. Adding the `whitenoise` dependency is only justified if it is explicitly required by our Django settings, either via `WhiteNoiseMiddleware` or a WhiteNoise-based static files storage backend. At the moment, there is no such reference in the codebase.

The PR description mentions that `collectstatic` fails without this dependency, but no error logs or traceback have been provided to show that WhiteNoise is actually being imported or required. Without a concrete failure and a clear link to a specific settings entry, adding a new dependency increases maintenance and attack surface without demonstrated benefit.

If you encounter a `collectstatic` or startup failure related to static files in the future, please open an issue or PR that includes:

* the full error output or traceback
* the exact settings line that requires the dependency
* an explanation of why the existing static file setup is insufficient

For now, since there is no clear technical requirement for WhiteNoise in the current configuration, we are closing this PR.

PR #1982, opened 2026-01-21
Add migration for locale choices update

ArnavBallinCode:
added missing migration

## Summary by Sourcery

Enhancements:
- Synchronize event and user locale field definitions with the updated list of supported language choices via a new migration.


Saksham-Sirohi:
i think these are already done by @Sak1012

PR #2015, opened 2026-01-22
feat: implement primary email preference support

### LINKED ISSUE #1997: Check if email sending respect primary address
The #1722 enable users to add more email addresses and choose one as "primary". But we are not sure if other parts of the app respect this "primary" preference or not, like if the system is about to send email to speakers, it should pick the "primary" address to send.
We need to check and modify if needed.

---
### Update:

- After checking more on this, I found that, our login methods are not aligned with _django-allauth_, they don't respect user primary email. For the password method, we can enable _django-allauth_ [backend](https://docs.allauth.org/en/dev/installation/quickstart.html) to make it respect primary email address. But the 2FA, JWT methods are implemented with custom code, we can not change them with just one setting.

- Another difficult situation is that `Order` is bound with `email`:

  https://github.com/fossasia/eventyay/blob/519bd8e91cd94bfd398b63fbb7abbe3198e30422/app/eventyay/base/models/orders.py#L176

  It means that, if user creates an `Order`, then change primary email address, he/she may not able to see the orders bound with previous email address.
  There are also some places where we filter the `Event` based on the `Order`, it means that the `Event` bound with old orders could be out-of-sight after changing primary email address.
  https://github.com/fossasia/eventyay/blob/519bd8e91cd94bfd398b63fbb7abbe3198e30422/app/eventyay/eventyay_common/forms/filters.py#L22

sarafarajnasardi:
Fixes #1997 
implement primary email preference support
- Add get_primary_email() helper method to User model
- Update email sending code to use primary email (7 locations)
- Respect django-allauth EmailAddress primary field
- Backward compatible fallback to user.email

## Summary by Sourcery

Add primary email preference support and update email usage across the app to respect it.

New Features:
- Introduce a User.get_primary_email() helper that returns the primary email from django-allauth or falls back to the user.email field.

Enhancements:
- Update mail sending, notifications, and export functionality to consistently use the user's primary email instead of the raw user.email value.

### DIFF
--- a/app/eventyay/base/tests/test_user_primary_email.py
+++ b/app/eventyay/base/tests/test_user_primary_email.py
@@ -221,6 +221,8 @@ class GetPrimaryEmailTestCase(TestCase):
             from allauth.account.models import EmailAddress
             EmailAddress.objects.all().delete()
         except ImportError:
+            # allauth is optional; if it's not installed, there are no EmailAddress
+            # records to clean up and we can safely ignore this.
             pass
         
         User.objects.all().delete()


sarafarajnasardi:
@mariobehling I have reviewed and implemented all changes suggested by Copilot.

Sak1012:
Thank you for the PR, draft PR's older than a week are closed by the team, please open a new PR once ready.

Issue #2042, opened 2026-01-23
Preselect custom field target based on entry point in speaker forms (Speaker, Proposal, Reviewer)


mariobehling:
In the **talk component**, organisers can configure three distinct data collection areas for the Call for Speakers:

* **Speaker information**
* **Proposal (session) information**
* **Reviewer information**

Each of these sections provides a **“+ New custom field”** button. However, all three buttons currently lead to the **same custom field creation form**, with no preselection applied. The organiser must manually choose the correct target via a dropdown on the creation form.

This behavior is unintuitive and error prone, especially in larger CfPs with many custom fields.

### Current Behavior

* Speaker Information, Proposal Information, and Reviewer Information each have a “+ New custom field” button
* Clicking any of these buttons opens the same custom field creation form
* The **target dropdown** (per speaker / per proposal / for reviewers) is not preselected
* Organisers must manually choose the correct option, even though the context is already known

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/149cb558-ea60-422a-8d0d-8d21c55b8c0f" />
<img width="2490" height="1142" alt="Image" src="https://github.com/user-attachments/assets/f65f34f1-77db-422c-b8f2-e109328a173f" />

### Expected Behavior

* The custom field creation form should **preselect the correct target** automatically based on which button was used:

  * Clicking **+ New custom field** under *Speaker Information* → target preselected to **per speaker**
  * Clicking **+ New custom field** under *Proposal Information* → target preselected to **per proposal**
  * Clicking **+ New custom field** under *Reviewer Information* → target preselected to **for reviewers**
* The dropdown should remain editable in case the organiser wants to change it

### Proposed Changes

* Pass context (speaker / proposal / reviewer) from the originating button to the custom field creation view
* Use this context to preselect the corresponding option in the target dropdown
* Keep existing form logic and validation unchanged

### Scope of Work

* Identify all entry points that link to the custom field creation form in the talk component
* Extend routing or query parameters to include the target context
* Apply default selection logic in the form view
* Ensure no regression for direct access to the custom field creation page

### Acceptance Criteria

* Custom field target is automatically preselected based on where the user clicks “+ New custom field”
* Dropdown remains editable
* No changes to existing custom fields or stored data
* No regressions in speaker, proposal, or reviewer form configuration

### Notes

This is a UX improvement that reduces friction and prevents accidental misconfiguration of custom fields in the Call for Speakers setup.

[...]

PR #2045, opened 2026-01-23
Preselect custom field target based on entry point in speaker forms

SxxAq:
Fixed custom field target preselection in the CfP form configuration. The buttons for creating new custom fields under "Speaker Information", "Proposal Information", and "Reviewer Information" now properly preselect the corresponding target dropdown value.

### Changes
- Modified `QuestionView.get_form_kwargs()` to pass initial values (including target parameter) to the form
- Fixed `get_initial()` to extract target from URL query parameters

### Before
- Organizers had to manually select the target dropdown when creating custom fields
- No indication of which section's button was clicked

### After
- Target field is automatically preselected based on the entry point:
  - Speaker Information button → "per speaker" preselected
  - Proposal Information button → "per proposal" preselected  
  - Reviewer Information button → "for reviewers" preselected
- Dropdown remains editable if organizer needs to change it
- Direct form access without target parameter still works

Fixes **#2042**

## Summary by Sourcery

Bug Fixes:
- Ensure the CfP custom question form receives the correct initial target value from URL query parameters so the appropriate target option is preselected.

### DIFF
--- a/app/eventyay/orga/views/cfp.py
+++ b/app/eventyay/orga/views/cfp.py
@@ -266,20 +266,10 @@ class QuestionView(OrderActionMixin, OrgaCRUDView):
     context_object_name = 'question'
     detail_is_update = False
 
-    def get_initial(self):
-        try:
-            initial = super().get_initial()
-        except AttributeError:
-            initial = {}
-        if 'target' in self.request.GET:
-            initial['target'] = self.request.GET['target']
-        return initial
-
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
-        initial = kwargs.get("initial", {}) or {}
-        initial.update(self.get_initial())
-        kwargs["initial"] = initial
+        if target := self.request.GET.get('target'):
+            kwargs.setdefault('initial', {})['target'] = target
         return kwargs
 
     def get_queryset(self):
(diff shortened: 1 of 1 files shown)


SxxAq:
@mariobehling 

https://github.com/user-attachments/assets/4eaaaa0e-14b3-4af9-a721-46cd875c802c

app/eventyay/orga/views/cfp.py
```diff
@@ -267,11 +267,21 @@ class QuestionView(OrderActionMixin, OrgaCRUDView):
...
             initial['target'] = self.request.GET['target']
         return initial
 
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        initial = kwargs.get("initial", {}) or {}
+        initial.update(self.get_initial())
+        kwargs["initial"] = initial
+        return kwargs
+
```
norbusan, line 274: Isn't that a bit complicated? What about dropping get_initial and do:
```
  def get_form_kwargs(self):                                                                                                    
      kwargs = super().get_form_kwargs()                                                                                        
      if target := self.request.GET.get('target'):                                                                              
          kwargs.setdefault('initial', {})['target'] = target                                                                   
      return kwargs                                                                                                             
```
instead?

The same pattern is in AccessCodeView line ~685

Also, in the original code, why two times `{}`?
```
initial = kwargs.get("initial", {}) or {}
```
The `get` already defaults to `{}`.

SxxAq: You're  right! That's much cleaner. The separate get_initial() method was unnecessary and the or {} was redundant.
Applied your suggestion to both QuestionView and AccessCodeView for consistency. Thanks for catching that!

ArnavBallinCode:
Thanks for the pr, This issue has been solved in https://github.com/fossasia/eventyay/issues/2092

PR #2095, opened 2026-01-25
feat: refactor question toggle handlers 

### LINKED ISSUE #2091: Custom Questions Cannot Be Deactivated
Custom questions cannot be deactivated after they are created.
When attempting to deactivate a custom question, the system displays the message “Custom fields cannot be deactivated, unlike system fields.”

This limits flexibility when managing registration forms, especially when a question is no longer needed but should be preserved for reference or historical data.

Image

**Steps to Reproduce**
1. Go to Tickets component → Order Form → Attendee Data
2. Create a new Custom Question
3. Save the changes
4. Attempt to deactivate the custom question
5. Observe the displayed message: “Custom fields cannot be deactivated, unlike system fields.”

**Expected Behavior**

Custom questions should be able to be deactivated, similar to system fields.
Users should have the option to disable a custom question without deleting it, allowing better flexibility in managing registration forms.

sarafarajnasardi:
Fixes #2091 

### Description
This PR adds the ability to **deactivate custom questions** in the Eventyay platform.  
Previously, once a custom question was created, it could not be hidden without deletion. With this change, organizers can toggle a question’s **Active** status to prevent it from appearing in future orders while still preserving historical responses.

---

#### 🗄️ Database
- Added an `active` boolean field to the `Question` model  
  - Default value: `true`

#### ⚙️ Backend
- Updated `QuestionToggle` view to handle the `active` field via AJAX.
- Modified `OrderQuestionsViewMixin` and `AbstractPosition.cache_answers` to exclude inactive questions during:
  - Checkout
  - Order processing

#### 🎨 Frontend
- Added an **Active** toggle switch in the `orderforms.html` template.
- Updated `orderFormToggles.js` to handle interactions for the `active` toggle.

## Summary by Sourcery

Add support for deactivating custom questions so inactive questions are hidden from user flows while preserving existing data.

New Features:
- Introduce an 'active' flag on questions that can be toggled from the order forms UI via AJAX.

Enhancements:
- Exclude inactive questions from order position answer caching and question selection in order-related views.

Chores:
- Add a database migration to persist the new question active state.

### DIFF
--- a/app/eventyay/static/pretixcontrol/js/orderFormToggles.js
+++ b/app/eventyay/static/pretixcontrol/js/orderFormToggles.js
@@ -1,320 +1,87 @@
-/**
- * Order Form Toggle Handlers
- */
+/*global $*/
 
-const FIELD_STATES = {
+const REQUIRED_STATES = {
     OPTIONAL: 'optional',
     REQUIRED: 'required',
     DO_NOT_ASK: 'do_not_ask'
 };
 
-const VALID_STATES = Object.values(FIELD_STATES);
+const REQUIRED_STATES_ARRAY = Object.values(REQUIRED_STATES);
 
-class CSRFTokenManager {
-    static #cachedToken = null;
-
-    static getToken() {
-        if (this.#cachedToken) {
-            return this.#cachedToken;
-        }
-
-        const cookieNames = ['pretix_csrftoken', 'csrftoken', 'eventyay_csrftoken'];
-        for (const name of cookieNames) {
-            const token = this.#getCookieValue(name);
-            if (token) {
-                this.#cachedToken = token;
-                return token;
-            }
-        }
-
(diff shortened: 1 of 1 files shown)


MukundC25:
@sarafarajnasardi address reviews and add video/screenshots of the changes, also rename the title correctly

sarafarajnasardi:
@ArnavBallinCode Thanks for the review. I investigated the failure and realized I made a mistake by relying on the static ``settings.py`` (which defines ``eventyay_csrftoken``) instead of verifying the actual runtime environment.
Upon debugging the live browser session, I found the app is actually serving ``pretix_csrftoken`` (likely due to legacy middleware).

I have refactored the implementation to be much more robust.

https://github.com/user-attachments/assets/6663bc3d-67ae-44ba-8d52-80b223968a97

ArnavBallinCode:
Thank you for the pr, but this issue has already been solved in another pr solving multiple issues on the same page ( #2100 ).

PR #2098, opened 2026-01-25
chore: give return message on no email selection

### LINKED ISSUE #1998: Remove edge case bug causing TemplateDoesNotExist error in Manage Email Addresses page
**Describe the bug**
In the Manage Email Addresses page, when there is only one email address, it is selected by default as shown here.

Image

However, when there are more than one email address tied to that account, the radio button is not selected for anyone, however the three buttons for Make Primary, Re-Send Verification and Remove still remain active.

Image

Clicking on these buttons lead to this particular error.

Image

*IMPORTANT NOTE:* This bug only happens if there is no explicitly set primary address. When there is just one address, the sole email is treated as the primary address. After selecting one of the emails as Make Primary, this issue vanishes.

Image

**Expected behavior**
If radio button is unselected, the button should remain disabled.
OR
Rework the primary email detection such that this particular bug is not encountered.

**Additional context**
I found a temporary fix to this edge case bug, making it difficult to properly record this bug. I will need to clone the repo again, after which I will record a video to properly highlight this issue.

Saksham-Sirohi:
Fixes #1998

## Summary by Sourcery

Bug Fixes:
- Prevent email management actions from proceeding when no email address is selected, showing an appropriate error message instead.

### DIFF
--- a/app/eventyay/base/forms/questions.py
+++ b/app/eventyay/base/forms/questions.py
@@ -501,15 +501,24 @@ class BaseQuestionsForm(forms.Form):
             )
             add_fields['state'].widget.is_required = True
 
-        field_positions = list(
-            [
-                (
-                    n,
-                    event.settings.system_question_order.get(n if n != 'state' else 'country', 0),
-                )
-                for n in add_fields.keys()
-            ]
-        )
+        # Build field positions using saved order from system_question_order
+        system_question_order = event.settings.system_question_order or {}
+        
+        field_positions = []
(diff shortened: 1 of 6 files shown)


Saksham-Sirohi:
Screenshot 2026-01-25 at 11 21 24 PM
this is what you'll now get if no email address is selected

Saksham-Sirohi:
Screenshot 2026-01-26 at 3 15 07 PM
@mariobehling fixed and addressed reviews

PR #2120, opened 2026-01-26
chore: yes/no rename to confirmation

### LINKED ISSUE #2118: Rename “Yes / No” field type to “Confirmation” in talk component
The current **Yes / No** field type is misleading. Functionally, this field does not represent a true binary choice. When selected, it requires the user to explicitly confirm via a mandatory checkbox and, in some cases, a confirmation dialog before proceeding.

This behaviour is closer to a confirmation or acknowledgement mechanism rather than a yes or no question. The current naming causes confusion for organisers configuring forms in the talk component and does not accurately reflect how the field behaves for end users.

Image

## Expected behaviour

• The field type is renamed from **Yes / No** to **Confirmation**
• Naming reflects the actual semantics and user interaction
• Organisers clearly understand that this field requires explicit confirmation

## Current behaviour

• Field type is labelled **Yes / No**
• Field enforces mandatory confirmation
• Naming suggests a choice where none actually exists

## Scope

• Rename **Yes / No** to **Confirmation** in all relevant dropdowns
• Update labels, helper texts, and descriptions where applicable
• Ensure consistency across organiser UI and any related documentation
• No change to underlying behaviour or validation logic

## Acceptance criteria

• All organiser facing references to **Yes / No** are replaced with **Confirmation**
• No residual **Yes / No** wording remains in the UI where this field type is used
• Existing forms continue to function without changes
• Behaviour remains identical, only naming and wording are updated

Saksham-Sirohi:
Fixes #2118


Saksham-Sirohi:
Screenshot 2026-01-26 at 3 47 52 PM
@mariobehling

PR #2165, opened 2026-01-29
bug:active column bug

Saksham-Sirohi:
orders page in control area causes an error 500 this pr is meant to be a fix for that

## Summary by Sourcery

Chores:
- Reformat the Question.active BooleanField definition in the migration for consistency.


Saksham-Sirohi:
hi @Sak1012 recreating migration does not work for us we get the same error injecting raw sql seem to be working to solve it but i dont think that is the best or ideal approach for us

PR #2195, opened 2026-02-01
Re-enable organizer events API endpoint using ReadOnlyModelViewSet

Anurag-1401:
### What does this PR do?
Re-enables the `/api/v1/organizers/{organizer}/events/` API endpoint, which was previously disabled due to Pretix-specific dependencies.

The endpoint now correctly returns a list of events for a given organizer using a read-only API.

### Related issue
Fixes #2160

### Why is this needed?
The organizer events endpoint is currently registered in routing but does not return any data because `EventViewSet` was stubbed out when Pretix dependencies were removed.

This results in:
- Broken API behavior
- Inconsistent public API contract
- Downstream integrations being unable to fetch organizer events

### How does this work?
- Replaces the disabled `ModelViewSet` with a `ReadOnlyModelViewSet`
- Removes all Pretix-specific dependencies
- Uses the existing ENext `EventSerializer`
- Filters events by the organizer slug from the URL
- Explicitly restricts access to authenticated users
- Ensures only read operations (`list`, `retrieve`) are exposed

### Safety and scope
- No write operations are enabled
- No database schema changes
- No changes to existing event data
- Scope is limited to restoring read-only access

### How was this tested?
- Requested `/api/v1/organizers/{organizer}/events/` with an authenticated user
- Verified a list of events is returned for the given organizer
- Confirmed unauthenticated access is rejected
- Verified existing event detail endpoints remain unaffected

## Summary by Sourcery

Restore a read-only organizer events API endpoint that returns events scoped to a given organizer using the existing ENext serializer and standard authentication.

New Features:
- Expose a read-only `/api/v1/organizers/{organizer}/events/` endpoint that lists events for a specific organizer to authenticated users.

Enhancements:
- Remove legacy Pretix-dependent `EventViewSet` stub and replace it with a minimal, framework-native implementation based on `ReadOnlyModelViewSet`.

### DIFF
--- a/app/eventyay/api/views/event.py
+++ b/app/eventyay/api/views/event.py
@@ -20,6 +20,8 @@ from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_POST
 from django_filters.rest_framework import DjangoFilterBackend, FilterSet
 from django_scopes import scopes_disabled
+from eventyay.models import Event, Organizer
+from eventyay.api.serializers.event import EventSerializer
 from rest_framework import exceptions, filters, serializers, views, viewsets
 from rest_framework.authentication import get_authorization_header
 from rest_framework.decorators import api_view, permission_classes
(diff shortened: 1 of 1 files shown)


Saksham-Sirohi:
closing as the functionality in this pull request is not desired

PR #2190, opened 2026-01-31
Fix duplicate i18n strings and restore site-wide translations

abhivyakti2:
Fixes #2187 

This PR resolves translation issues where multiple UI strings were falling back
to English despite a different language being selected.

Changes:
- Removed duplicate msgid entries across locale .po files
- Merged duplicates while preserving references and translations
- Cleaned up fuzzy/duplicate entries causing compilemessages warnings
- compilemessages now runs without warnings

Result:
- Translations now work consistently site-wide
- Only strings genuinely missing from locale files fall back to English
image

Although reported for the sidebar (#2187), the root cause affected translations
across the entire site, and this PR fixes it globally.

## Summary by Sourcery

Fix translation handling in the organizer UI and clean up locale files to restore consistent site-wide i18n behavior.

Bug Fixes:
- Ensure sidebar labels like General, Speakers, Sessions, and Schedule use proper Django translation tags so they respect the selected language.
- Resolve locale .po issues so translations no longer fall back to English when valid translations exist.

Enhancements:
- Align organizer sidebar templates with the standard Django i18n mechanism for commonly used navigation labels.

Chores:
- Deduplicate and tidy translation entries across all locale .po files to remove warnings and keep compilemessages output clean.

PR #2205, opened 2026-02-02
Fix untranslated sidebar navigation items in organizer area

### LINKED ISSUE #2187: Missing Translations in  Sidebar Navigation
## Description
The sidebar navigation menu , some of the items are not being translated and appear only in English, even when users have selected a different language. This is inconsistent in the website where navigation items are properly translated.

Image

Image

Anurag-1401:
### What does this change do?

Fixes #2187 

This PR fixes several sidebar navigation items in the organizer interface that
were not being translated and always appeared in English, even when a different
language was selected.

The main navigation was already properly translated, but the sidebar contained
hardcoded or untranslated labels, leading to an inconsistent multilingual
experience.

### What was the problem?

- Sidebar navigation showed English-only labels
- Language switcher worked correctly elsewhere
- Issue was reproducible across multiple languages
- Inconsistency between top navigation and sidebar

### How is it fixed?

- Ensured all sidebar labels are wrapped with Django i18n translation tags
- Reused existing translation keys where available
- Kept behavior unchanged while fixing localization consistency

### Screenshots / Visual reference

Before: Sidebar items remained in English after language switch  
After: Sidebar navigation follows the selected language consistently

### Related issue

Fixes #2187

## Summary by Sourcery

Localize organizer sidebar navigation labels and remove obsolete translation template.

Bug Fixes:
- Ensure organizer sidebar items like Speakers, Sessions, and Schedule use Django translation tags so they follow the selected language.

Chores:
- Remove the outdated django.pot translation template file from the locale directory.

### DIFF
--- a/app/eventyay/orga/templates/orga/base.html
+++ b/app/eventyay/orga/templates/orga/base.html
@@ -283,8 +283,7 @@
                                 <span class="has-children">
                                     <a class="nav-link nav-link-inner" href="{{ request.event.orga_urls.speakers }}">
                                         <i class="fa fa-address-card-o"></i>
-                                        <span class="sidebar-text">{% translate "Speakers" %}</span>
-                                    </a>
+                                        <span class="sidebar-text">{{ phrases.schedule.speakers }}</span>                                    </a>
                                     <a class="arrow nav-link" data-toggle="collapse" data-target="#collapseSpeakers" aria-controls="collapseSpeakers">
                                         <i class="fa fa-angle-down"></i>
                                     </a>
@@ -310,8 +309,7 @@
                             <span class="has-children">
                                 <a class="nav-link nav-link-inner" href="{{ request.event.orga_urls.submissions }}">
                                     <i class="fa fa-sticky-note-o"></i>
-                                    <span class="sidebar-text">{% translate "Sessions" %}</span>
-                                </a>
+                                    <span class="sidebar-text">{{ phrases.schedule.sessions }}</span>                                </a>
                                 <a class="arrow nav-link" data-toggle="collapse" data-target="#collapseSubmissions" aria-controls="collapseSubmissions">
                                     <i class="fa fa-angle-down"></i>
                                 </a>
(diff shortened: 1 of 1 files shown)


Anurag-1401:
Thanks for the detailed review — I’ve addressed all the points raised.

Reverted the sidebar labels to reuse existing phrases.schedule.* keys (speakers, sessions, schedule) to stay consistent with the main navigation and avoid introducing duplicate translation msgids.

Replaced {% translate %} with Django’s correct {% trans %} template tag to ensure proper rendering and avoid runtime issues.

No changes were made to localization infrastructure as part of this fix; the sidebar changes are now fully scoped to template-level adjustments only.

Please let me know if you’d like me to align any additional sidebar labels or navigation elements as a follow-up.

abhivyakti2:
Just to clarify one detail around Django i18n: the django.pot file is only a template used by translators and tooling (e.g. as the source for generating or updating .po files). It is not used at runtime by Django when rendering translations.

Translations start working only after the language-specific .po files are compiled into .mo files via compilemessages. Because of that, removing django.pot doesn’t fully address the issue and isn’t required for translations to work; it can actually be useful to keep it for translators and future message updates.

I worked on this same issue earlier and submitted a PR #2190 a few days ago that fixes it by compiling the message files, which is the step that actually resolves the problem in practice. 

Appreciate you digging into this, it’s an easy part of the workflow to get mixed up.

Anurag-1401:
Thanks for confirming.

I agree that keeping django.pot is the right approach since it’s useful for translators and future message updates, even though it isn’t used at runtime. I’ll avoid touching it going forward and keep changes tightly scoped to the actual issue being addressed.

Also noted regarding the importance of compiling .po --> .mo files — that helps explain why your earlier PR #2190 resolves the problem in practice. Appreciate the clarification and the context.

Thanks again for the review and guidance.

Sak1012:
Thank you for the contribution, this PR doesn't solve the issue, please take a look at new issues for future contributions.

PR #2259, opened 2026-02-05
Add global keyboard shortcut to focus search input

### LINKED ISSUE #2256: Keyboard shortcut for search
## Describe the enhancement

The search feature on `dev.eventyay.com` currently lacks a standard keyboard shortcut for quick access.

Modern web platforms commonly support a global search shortcut:

- **Ctrl + K** (Windows/Linux)
- **Cmd + K** (macOS)

Adding this will improve accessibility, usability, and align Eventyay with industry-standard UX patterns.

---

## Expected behavior

- Pressing **Ctrl + K** or **Cmd + K** from anywhere on the start page should:
  - Automatically focus the search input field
  - Allow users to immediately start typing their query

---

## Why this matters

This shortcut is widely adopted across major platforms like Discord, Notion, Slack, etc and users increasingly expect it as a default productivity feature.

---

## Additional context

This issue is separate from the main bug where the search box is non-functional, and focuses only on keyboard accessibility and UX enhancement.

abhivyakti2:
Fixes #2256 
### Description
Adds a global keyboard shortcut (Ctrl + K / Cmd + K) to focus the search input on the start page, improving keyboard accessibility and aligning with common UX patterns. This change is independent of the existing search functionality bug.

https://github.com/user-attachments/assets/e0576914-a99d-44aa-a1dd-265405ede4dd

## Summary by Sourcery

New Features:
- Introduce a Ctrl+K / Cmd+K shortcut on the start page to focus the search input field.


ArnavBallinCode (changes requested):
Please move the this to a separate js file

abhivyakti2:
Hi @ArnavBallinCode , While debugging this, I found the start page JS wasn’t running due to a syntax error in the compressed bundle. It was caused by an `export` in `dropdown.js`, which is included via `{% compress js %}` and executed as a classic script, blocking all following JS (including this feature).
I’m checking whether `initDropdowns` is used elsewhere to decide the right fix, should this be handled in this PR or separately?

app/eventyay/static/pretixpresale/js/startpage.js
```diff
@@ -177,9 +177,29 @@
...
 
+  function initKeyboardShortcuts() {
+    document.addEventListener('keydown', function (event) {
+      if (event.target.matches('input, textarea, [contenteditable]')) {
+        return;
+      }
+      if (
+        !event.repeat &&
+        (event.ctrlKey || event.metaKey) &&
+        event.key.toLowerCase() === 'k'
```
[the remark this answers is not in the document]
ArnavBallinCode: please address this

ArnavBallinCode (changes requested):

app/eventyay/static/pretixpresale/js/startpage.js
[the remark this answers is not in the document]
abhivyakti2: Fixed. Shortcut now triggers only on Ctrl/Cmd+K. Thanks!

app/eventyay/static/common/js/dropdown.js
```diff
@@ -69,4 +69,3 @@ if (document.readyState === 'loading') {
     initDropdowns();
 }
 
-export { initDropdowns };
```
Saksham-Sirohi, line 116: please explain rationale behind this change

abhivyakti2: Thanks for flagging this again. While debugging, I found the start page JS was breaking because `dropdown.js` is bundled via `{% compress js %}` and runs as a classic script, having an `export` there causes a syntax error and blocks all subsequent JS. The dropdown logic is self-initialized on load, so it doesn’t need to be exported.

Saksham-Sirohi: Django Compressor is configured with an esbuild-based precompiler specifically for module-type scripts
Esbuild handles export statements natively and would bundle/strip them in the output
The <script type="module"> attribute is respected by the compressor pipeline; however, this is a welcome change.

abhivyakti2: Thanks for explaining. I was seeing a runtime “Unexpected token export” error on the page, and removing the export resolved it.

PR #2261, opened 2026-02-05
Fix AttributeError: 'NoneType' object has no attribute 'question' in speaker profile form

### LINKED ISSUE #2260: AttributeError: 'NoneType' object has no attribute 'question' when accessing speaker profile page
### **Description**
An `AttributeError: 'NoneType' object has no attribute 'question'` is raised when accessing the speaker profile submission page at /{organizer}/{slug}/submit/{proposal}/profile/. The error occurs in `eventyay/common/forms/mixins.py` at line 200 in the `inject_questions_into_fields` method.
### **Expected behavior**
The speaker profile page should load correctly without errors, displaying the form with all configured questions.
### **Screenshots**

Image

abhivyakti2:
Fixes #2260 
### Summary
Fixes an `AttributeError` on the speaker profile submission page caused by missing null handling when creating form fields for questions with unsupported variants.

### Changes Made
**File:** `eventyay/common/forms/mixins.py`

- Added a null check after calling `get_field()` to safely handle unsupported question variants.
  - `get_field()` may return `None` when a question’s variant type is not supported.
  - Previously, the code attempted to access `field.question` on a `None` value, causing a crash.
- Added warning-level logging when a question field is skipped.
  - Logs include the question ID and variant type for easier debugging.
  - Helps identify unsupported variants without interrupting page rendering.

### Root Cause
The `get_field()` method returns `None` (line 444) when a question’s variant does not match any supported types (BOOLEAN, NUMBER, STRING, URL, TEXT, FILE, CHOICES, MULTIPLE, DATE, DATETIME).  
`inject_questions_into_fields()` did not account for this case and attempted to access attributes on a `None` value, resulting in an `AttributeError`.

## Summary by Sourcery

Handle unsupported question variants safely when injecting dynamic question fields into speaker-related forms.

Bug Fixes:
- Prevent AttributeError when creating form fields for questions whose variants are not supported by get_field().

Enhancements:
- Log a warning and skip questions when form field creation fails due to unsupported variants, aiding debugging without breaking form rendering.

### DIFF
--- a/app/eventyay/common/forms/mixins.py
+++ b/app/eventyay/common/forms/mixins.py
@@ -182,7 +182,7 @@ class QuestionFieldsMixin:
         for question in questions.prefetch_related('options'):
             initial_object = None
             initial = question.default_answer
-            
+
             if target_object:
                 answer = answers_by_question.get(question.id)
                 if answer:
@@ -197,6 +197,13 @@ class QuestionFieldsMixin:
                 initial_object=initial_object,
                 readonly=readonly,
             )
+            if field is None:
+                logger.warning(
+                    "Could not create form field for question %s (variant: %s). Skipping.",
+                    question.pk,
+                    question.variant,
+                )
+                continue
             field.question = question
             field.answer = initial_object
             self.fields[f'question_{question.pk}'] = field


hongquan (changes requested):
This doesn't look like a fix.
You have to investigate why `field` became `None`? What's wrong that made it become `None`.

abhivyakti2:
Thanks for the feedback. I’ll dig deeper into why `get_field()` returns `None` for these variants and address the underlying issue rather than just handling the result.

mariobehling:
Thanks a lot for taking the time to investigate this and for submitting the PR.

We are closing it for now because the current change only adds a defensive `None` check and skips questions when `get_field()` cannot build a field. While this avoids the immediate crash, it does not address why `get_field()` returns `None` in the first place, and it can lead to a silent failure where configured questions simply do not appear in the form. That makes the issue harder to detect and debug for organizers and admins.

A more sustainable fix would tackle the root cause. For example, this could mean adding support for the missing variant in `get_field()`, preventing unsupported variants from being created or stored through validation or migration, or surfacing a clear, admin facing error that points to the specific question that needs correction.

One additional note for future contributions: we value human led PRs where the author has a clear understanding of the problem and the solution. AI tools are fine as assistance, but every line should be reviewed and understood. Being able to reproduce the issue locally, explain the failure mode, identify the exact variant involved, and justify the chosen fix and tests makes reviews much smoother and helps keep the codebase maintainable.

If you would like to continue working on this, please feel free to open a new PR that includes reproduction steps, the exact variant that triggers the `None` case, a brief root cause analysis, and a fix that either supports the variant or blocks it explicitly with validation and tests. We are happy to review that.

PR #2283, opened 2026-02-06
fix : Backfill NULL TalkQuestion.question values to harden migration

### LINKED ISSUE #2279: Follow up: Backfill NULL TalkQuestion.question values to harden migration
PR #2273 changed `TalkQuestion.question` from nullable to required and renamed the label to “Custom question”. The PR is merged and working correctly.

However, the database migration does not explicitly backfill existing NULL values. The schema change relies on an implicit default during the ALTER step, which can lead to inconsistent behavior across environments.

Specifically:
• The field was previously nullable and is now non nullable
• Existing rows with `question IS NULL` are not explicitly handled
• Migration behavior may differ depending on database engine, version, or restore state
• Deployments using older database snapshots may fail unexpectedly

This is a schema hardening issue, not a functional regression.

## Possible solution

Add a follow up migration that explicitly backfills existing NULL values before enforcing the non nullable constraint.

### Migration steps

1. Add a `RunPython` step to update legacy rows
2. Backfill all `TalkQuestion` records where `question IS NULL`
3. Keep the existing non nullable constraint unchanged

### Example logic

```
TalkQuestion.objects.filter(question__isnull=True).update(question='')
```

## Scope

• Database migration only
• No UI changes
• No API changes
• No behavioral changes
• No data loss

ViRUS-0-0:
Fixes #2279 

- Added a backfill_question function to migration `0013_alter_talkquestion_question.py`.
- Inserted a `migrations.RunPython` step to update existing NULL values in the question field of `TalkQuestion` to empty strings ('').
- This ensures the non-nullable constraint is applied safely across all environments.

## Summary by Sourcery

Bug Fixes:
- Populate empty strings for existing TalkQuestion records with NULL question values during migration to prevent failures when applying the non-null constraint.

### DIFF
--- a/app/eventyay/cfp/templates/cfp/event/user_submission_edit.html
+++ b/app/eventyay/cfp/templates/cfp/event/user_submission_edit.html
@@ -66,10 +66,10 @@
         </h5>
         <form method="post" enctype="multipart/form-data" novalidate>
             {% include "common/forms/errors.html" %}
-            {% include "common/forms/errors.html" with form=qform %}
+
             {% csrf_token %}
             {{ form }}
-            {{ qform }}
+
             {% if can_edit or form.instance.resources.count %}
                 {% include "cfp/includes/submission_resources_form.html" %}
             {% endif %}
--- a/app/eventyay/cfp/views/user.py
+++ b/app/eventyay/cfp/views/user.py
@@ -334,17 +334,7 @@ class SubmissionsEditView(LoggedInEventPageMixin, SubmissionViewMixin, UpdateVie
 
         return True
 
-    @context
-    @cached_property
-    def qform(self):
-        return TalkQuestionsForm(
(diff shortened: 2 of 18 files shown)


mariobehling:
Thank you. Please address AI reviews.


Small process note.

We have automatic Copilot PR reviews enabled on this repository. These reviews are only triggered if the contributor has GitHub Copilot enabled and an active license on their own account.

Please enable Copilot in your GitHub settings if you have access. In many regions, free licenses are available through educational institutions or developer programs. Enabling Copilot helps us speed up the auto review process and reduces manual review overhead for the core team.

hongquan (approved):
It is Ok, but will be better if using:

```py
TalkQuestion.objects.filter(question=None).update(question='')
```

(more intuitive).

PR #2349, opened 2026-02-10
ci: load CI config from dedicated TOML and simplify services

Rachit7168:
Introduce a dedicated CI configuration file (eventyay.ci.toml) and update the GitHub Actions workflow to load configuration from it instead of duplicating environment variables.

This simplifies the CI workflow, keeps runs deterministic, and aligns CI with the project’s TOML-first configuration approach. Verbose service health checks and manual wait loops are removed where not strictly required. The CI test scope (tests/stable) remains unchanged and is now explicitly intentional.

## Summary by Sourcery

Load Django CI settings from a dedicated TOML config file and simplify the GitHub Actions service setup.

New Features:
- Introduce an eventyay.ci.toml configuration file describing CI-specific app, database, Redis, and environment settings.

CI:
- Update video-backend CI workflow to use EVY_CONFIG_FILE pointing to a shared eventyay.ci.toml for checks, migrations, and tests.
- Remove custom health-check options and manual wait logic for Postgres and Redis services in CI jobs.
- Drop unused Redis and Postgres client tooling from the CI runner package installation step.


hongquan:
Why do we suddenly have `EVY_CONFIG_FILE`? It is not recognized.

Rachit7168:
Good catch — thanks for pointing this out.

This is not a clean PR , There were 3 options I am testing all 3 option and only Final PR oif good workig one which is better

EVY_CONFIG_FILE is newly introduced in this PR to explicitly load a CI-specific TOML config, but the backend smoke test revealed that the variable is not currently being picked up in that path.

I’m updating the workflow to define EVY_CONFIG_FILE at the job level and verifying that the backend config loader explicitly reads this variable during server startup. This should make the CI behavior explicit and fix the healthcheck failure.

Will push a follow-up commit shortly to address this.

PR #2358, opened 2026-02-10
ignore this request

Anurag-1401:
## Summary by Sourcery

Adjust CI test workflow and improve color picker synchronization in the editor UI.

Enhancements:
- Synchronize the text color input and color picker UI when selecting text elements in the editor so the current color is accurately reflected.

CI:
- Update GitHub Actions test workflow to install system packages with non-interactive flags, run only the stable subset of tests, and clarify this behavior in comments.

Chores:
- Revise CI environment variables to use generic app configuration values instead of direct database connection settings.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,48 +20,64 @@ jobs:
       matrix:
         python-version: [3.12]
         database: [postgres]
+
     steps:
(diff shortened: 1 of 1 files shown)


mariobehling:
Seems to be a PR made as an error.

Issue #2316, opened 2026-02-08
Add AI-assisted bulk translations for French, Arabic, and Spanish


mariobehling:
The eventyay platform supports multiple languages, but the translations for **French**, **Arabic**, and **Spanish** are currently incomplete. This results in fallback English strings appearing in the UI.

Manually completing these translations is time-consuming and requires language skills that most developers do not have. To improve coverage efficiently, AI can be used to generate initial translations for missing strings.

## Objective

Use AI tools (such as Codex or Claude) to **automatically generate translations for missing keys** in French, Arabic, and Spanish locale files. The generated translations will later be reviewed and refined via Weblate.

### 1. Identify missing translation keys

• Use English as the source language
• Compare English locale files with
– French
– Spanish
– Arabic
• Extract only missing or empty keys
• Do not modify existing translations

### 2. Generate translations using AI

• Use an AI model to translate missing strings only
• Preserve
– JSON structure
– keys and nesting
– placeholders such as `{name}`, `{count}`, `%s`
• Do not translate keys

Example prompt:

```
Translate the following UI string into Spanish.
Preserve placeholders and meaning.
Text: "Buy ticket"
```

### 3. Populate locale files

• Insert AI-generated translations into
– `fr` locale files
– `es` locale files
– `ar` locale files
• Keep formatting and structure unchanged

### 4. Submit a pull request

The PR should include:
• Updated locale files for French, Spanish, and Arabic
• A short summary stating
– Number of keys added per language
• A note that translations were AI-generated and will be reviewed in Weblate

## Acceptance criteria

• All missing keys for French, Spanish, and Arabic are populated
• Existing translations are untouched
• Placeholders and formatting are preserved
• Application builds and runs without i18n errors
• Translations are ready for follow-up review in Weblate

## Notes

This issue is intentionally scoped to **translation generation only**.
Language quality review and refinements will be handled separately in Weblate.

mariobehling:
We were able to integrate weblate with OpenAI. So, lets try this alternative approach first. Closing this for now.

PR #2366, opened 2026-02-11
Expand and refine Arabic translations in the main Django .po localization file

ArnavBallinCode:
Arabic phase 2 of https://github.com/fossasia/eventyay/issues/2316

To streamline Arabic .po translation work, I built a small helper script that programmatically scans Django gettext files and extracts only untranslated entries (msgstr ""), along with their source references (#: file:line). This file is passed to ghcp and then another script which appends them back

## Summary by Sourcery

Enhancements:
- Expand and refine Arabic translations in the main Django .po localization file.


mariobehling:
We were able to integrate weblate with OpenAI. So, lets try this alternative approach first. Closing this for now.

PR #2444, opened 2026-02-16
feat: add product validity fields and check-in time validation

Vijaysathvik:
- Introduced new fields to the Product model: `checkin_text`, `valid_from`, `valid_until`, and `validity_type` to manage ticket validity.
- Implemented validation logic in the check-in service to ensure tickets are only valid within specified time frames.
- Updated the check-in interface to reflect these changes, including dynamic visibility of validity date fields based on the selected validity type.
- Enhanced user feedback for invalid check-in attempts due to time constraints.

### DIFF
--- a/app/eventyay/base/models/product.py
+++ b/app/eventyay/base/models/product.py
@@ -1,987 +1,1743 @@
-from decimal import Decimal
-from urllib.parse import urlencode
+import sys
+import uuid
+from collections import Counter, OrderedDict
+from datetime import date, datetime, time
+from decimal import Decimal, DecimalException
+from typing import Tuple
 
-from django import forms
+import dateutil.parser
+import pytz
+from django.conf import settings
 from django.core.exceptions import ValidationError
-from django.db.models import Max
-from django.forms.formsets import DELETION_FIELD_NAME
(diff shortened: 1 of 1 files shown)


mariobehling:
Thank you for the PR and the effort.

We are going to close this PR for now, mainly for process and scope reasons.

Why we are closing it

1. No linked issue and unclear scope
   This PR introduces new Product fields, a migration, check in validation logic, and frontend changes. That is out or scope as it is a meaningful feature change and would also need an issue first to agree on requirements, data model, and UI behavior. Right now there is no linked issue describing the problem, expected behavior, and edge cases, so reviewers have no shared baseline for what should be accepted.

2. Too many changes bundled together
   The PR mixes feature work with a larger refactor, including restructuring around products and introducing ProductCategory. Even if each part is valid, bundling them makes review and rollback difficult. This should be split into small, reviewable PRs after scope is agreed in an issue.

3. Review tooling also indicates risk
   Automated review tooling flags the PR as very large, and the diff view shows warnings about hidden or bidirectional Unicode text in at least one file. That is a red flag and needs careful manual verification, which we cannot justify without an agreed issue and a smaller change set.

PR #2454, opened 2026-02-17
update in the dev branch

Anurag-1401:
## Summary by Sourcery

Restrict CI to running a stable subset of tests while improving test workflow reliability and fix UI color picker synchronization in the editor.

Bug Fixes:
- Ensure the editor color input and color picker stay in sync when selecting text elements so the displayed color matches the actual object color.

Enhancements:
- Clarify and rename CI workflow steps for PostgreSQL, Django checks, asset compilation, and test execution, and improve dependency installation robustness.

CI:
- Limit CI pytest execution to the stable test subset, document the reduced scope, and adjust environment variables used during CI runs.


mariobehling:
Thank you for the PR and for taking the initiative to improve CI reliability.

We are going to close this PR for now.

Main reasons

1. Scope is unclear and not linked to an issue
   This PR changes CI behavior (including test scope) and also modifies the JS editor. These are unrelated areas and without a linked issue it is not clear what problem we are solving, what the expected outcome is, and what tradeoffs are acceptable.

2. Mixed changes make review and rollback risky
   CI workflow changes and UI editor changes should not be bundled. Even if both are valid, combining them makes it hard to review and harder to revert safely if one part is wrong.

3. CI changes look incorrect or unsafe for our setup
   From the diff, the workflow introduces several red flags that would require a deeper CI design discussion, for example:

* duplicated steps and conflicting step definitions
* working directory changes to `./src` which does not match our standard project layout
* forcing the CI suite to run only `tests/stable` in this PR, which is a major policy change and needs maintainer agreement first
* warning about hidden or bidirectional Unicode characters in the workflow file, which must be cleaned up before anything touches CI

4. Process guidance for new contributors
   We prefer new contributors work on existing issues first and avoid starting new DevOps and CI policy changes without maintainer alignment. CI and deployment are high impact and require coordination with the core team.

Thanks again for the effort and interest in helping improve the project.

Issue #2436, opened 2026-02-16
Improve Language Display, Sorting Strategy, and Formality Handling in Eventyay


mariobehling:
Eventyay displays languages in multiple areas in the organiser and attendee areas:

• UI language selector
• Event language selection
• Session submission language
• Profile language preferences

Currently there is no clearly defined strategy for:

• Display format of language names
• Sorting behavior
• Handling multi script languages such as Chinese
• Handling languages with formal and informal variants
• Whether sorting adapts to the current UI language

This leads to inconsistent UX and ambiguity in international contexts.

## Problem

Several languages supported by Eventyay include formality variants. Examples:

• German informal (du)
• German formal (Sie)
• Spanish informal
• French formal
• Portuguese informal

These are not separate languages. They are stylistic variants of the same locale.

If treated as separate languages in lists, this creates:

• UX clutter
• Confusion
• Incorrect alphabetical grouping
• Duplicate language entries

We need a consistent and technically correct model.

## Proposed Solution

Adopt international i18n best practices aligned with CLDR and ICU standards.

### 1. Display Format

Use dual labels:

Translated Name (Native Name)

Example in English UI:

German (Deutsch)
Spanish (Español)
Chinese (Simplified) (简体中文)
Chinese (Traditional) (繁體中文)

Example in German UI:

Deutsch (German)
Spanisch (Español)
Chinesisch (Vereinfacht) (简体中文)
Chinesisch (Traditionell) (繁體中文)

### 2. Sorting Strategy

Sort alphabetically by the translated name according to the current UI locale.

Sorting must:

• Follow current UI locale
• Use locale aware collation
• Ignore accent differences at primary sorting level
• Dynamically update when UI language changes

### 3. Chinese Handling

Chinese must be explicitly split:

• Chinese (Simplified) – zh-Hans
• Chinese (Traditional) – zh-Hant

Each includes native script in label.

These are separate locale codes and should appear as distinct entries.

### 4. Formal and Informal Variants

Formality must not be treated as a separate language.

Instead:

Model formality as a variant of a base locale.

Example:

Base locale:
de

Variants:
de informal
de formal

Recommended Technical Model:

Use BCP 47 extension keys or internal variant keys:

de
de-x-formal
de-x-informal

or

de@formal
de@informal

Recommended UX Behavior:

In language selector:

Group by base language.

Display:

German (Deutsch)
↳ Formal
↳ Informal

or

German (Deutsch)
Formal
Informal

In English UI:

[...]

PR #2460, opened 2026-02-19
fix(locale): update context, Chinese name and add `compilemessages`

Sak1012:
This PR further resolves issue #2436 and adds back `compilemessages` step.

image
image

## Summary by Sourcery

Align locale handling and language metadata with updated language ordering and Chinese naming, and ensure compiled translation messages are built in production.

Bug Fixes:
- Use centralized language choice ordering to populate supported languages in the locale context so UI language lists reflect the configured language priorities.

Enhancements:
- Update Simplified and Traditional Chinese display names to the new naming convention in language settings.
- Simplify locale language option construction by removing custom sorting logic in favor of the shared language ordering helper.

Build:
- Run Django translation message compilation during the production image build so localized strings are available at runtime.

### DIFF
--- a/app/eventyay/base/models/orders.py
+++ b/app/eventyay/base/models/orders.py
@@ -1364,16 +1364,22 @@ class AbstractPosition(models.Model):
 
         # We need to clone our question objects, otherwise we will override the cached
         # answers of other products in the same cart if the question objects have been
-        # selected via prefetch_related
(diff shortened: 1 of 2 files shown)


Sak1012:
> Why are all the other changes here, too? What is the need for them?

This PR also resolves #2436 and some changes are related to that.

PR #2530, opened 2026-02-23
Prevent password validation when changing full name

### LINKED ISSUE #2529: Changing name throwing the wrong error
## Changing Full Name Throws Incorrect Password Error

**Describe the bug**  
When attempting to change only the **Full Name** in Account → General Settings, the form throws an error stating:

> "Please enter your current password if you want to change your e-mail address or password."

No email or password changes were made — only the full name field was modified. Despite this, the system prevents saving and displays:

> "Your changes could not be saved. See below for details."

This behavior is incorrect because updating the full name should not require entering the current password.

**Expected behavior**  
Users should be able to update their Full Name without being required to enter their current password, since no sensitive login credentials are being modified.

**Screenshots**  
Image

**Additional context**  
- No changes made to email or password fields.
- Error appears immediately after attempting to save updated full name.

prayag78:
fixes #2529

https://github.com/user-attachments/assets/41003c43-a1d0-4ee3-a64a-76c77da50995

### Changes
- Allow changing account full name without requiring current password.
- Require current password only when password is being changed.

## Summary by Sourcery

Allow updating profile details without triggering unnecessary password or email validation.

Bug Fixes:
- Require the current password only when the account password is being changed, not when updating other profile fields like full name.
- Avoid incorrect email-change messaging by using the persisted user email when constructing the success message.

### DIFF
--- a/app/eventyay/base/forms/user.py
+++ b/app/eventyay/base/forms/user.py
@@ -18,7 +18,7 @@ class UserSettingsForm(forms.ModelForm):
         'duplicate_identifier': _(
             'There already is an account associated with this e-mail address. Please choose a different one.'
         ),
-        'pw_current': _('Please enter your current password if you want to change your e-mail address or password.'),
+        'pw_current': _('Please enter your current password if you want to change your password.'),
         'pw_current_wrong': _('The current password you entered was not correct.'),
         'pw_mismatch': _('Please enter the same password twice'),
         'rate_limit': _('For security reasons, please wait 5 minutes before you try again.'),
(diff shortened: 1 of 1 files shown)


Sak1012:
The PR actually bypasses password validation required to change E-mail but the title says `Prevent password validation when changing full name` which is wrong, the password validation is required when the user changes their email, it must not be removed.

prayag78:
@Sak1012 @ArnavBallinCode 

The email field is never rendered as an `<input>` in the template, so the browser never sends it on POST. Since the field wasn't disabled, Django normalized the missing value to "", causing ["" != user.email] -> True, triggering the password error even on a fullname-only change.

1. If we move `email.disabled = True` for all users(outside the 'native') Django then uses the model instance value instead of absent POST data, fixing the false positive.
2. And a null-safe comparison in `clean()`
```
email_changed = (email or '').strip().lower() != (self.user.email or '').strip().lower()
        if not self.requires_password_reset and (password1 or email_changed) and not old_pw:
```

The email-change password guard is fully preserved.

PR #2532, opened 2026-02-23
Improve translation handling

norbusan:
Change in cfp/flow.py

change the condition so any locale value that equals the English string (a dead giveaway of a failed gettext lookup) gets re-evaluated from the current .mo file.

Change in cfp/forms/cfp.py

get_editor_config writes every field's label into the database JSON — even fields the admin never touched. When those labels come back as stale English-fallback strings, _update_cfp_texts silently overrides the field's working gettext_lazy with them.

Fix done: in _update_cfp_texts (cfp/forms/cfp.py), don't apply the stored label if every translation in it is identical to the English value (i.e., it carries no useful i18n information)


mariobehling:
This PR fixes the stale fallback override issue, but there is one important scenario where the current heuristic can contradict an organiser’s explicit configuration.

The logic assumes that:

* If a locale value equals the English string, it is likely a failed gettext lookup.
* If all translations in the stored label are identical to the English value, the stored label carries no useful i18n information and should be ignored.

This assumption is not always valid.

There are legitimate cases where an organiser intentionally wants identical labels across languages. Examples:

* Brand terms such as “Gravatar”.
* Technical terms that should not be translated.
* Event specific naming conventions, e.g. Mozilla events using consistent terminology across locales.
* Organisers explicitly choosing to keep the English wording in all enabled languages.

In these cases:

* All locale values being identical to English does not mean “stale fallback”.
* It can represent an intentional, explicit configuration.
* The current condition would ignore that stored configuration and fall back to gettext via the .mo file.
* This effectively overrides the organiser’s choice.

The core issue is that equality to English is used as a proxy for “not customised”, which is not semantically equivalent to “admin never touched this field”.

To avoid contradicting valid organiser configurations, the decision whether to apply stored labels should be based on explicit edit intent rather than string equality. For example:

* Only persist labels for fields that were actually edited.
* Track whether a field or locale was modified by the organiser.
* Compare against the original default at save time instead of inferring intent at render time.

Without such differentiation, this PR risks fixing the fallback regression while introducing a new regression in legitimate custom naming scenarios.

norbusan:
> The core issue is that equality to English is used as a proxy for “not customised”, which is not semantically equivalent to “admin never touched this field”.

Completely agreed. This is what I addressed with the second commit.

I need to update the description.

Now the behavior is: If the English translation is different from the **default** translation, or there is any other translation different, then use the DB entries, otherwise allow updates via po/mo.

Issue #2462, opened 2026-02-19
Top navigation bar inconsistent and broken across multiple pages


Saksham-Sirohi:
**Describe the bug**
The top navigation bar appears broken and inconsistent across multiple pages, including the start pages. Elements do not align or display correctly, and the UI does not appear uniform across the site, leading to a confusing user experience. Some elements overlap, and spacing is off, as visible in the screenshots below.

**Expected behavior**
The top navigation bar should be visually consistent and properly functional on all pages, matching the site's UI guidelines with correct alignment and spacing of elements.

Additionally, please ensure that fixes do NOT break the UI appearance or layout on desktop devices. The desktop UI should remain unchanged and continue to match existing design guidelines.

**Screenshots**
<img alt="Top nav bar misalignment and inconsistency on event start page" width="1206" src="https://github.com/user-attachments/assets/aff02dd6-3671-4377-9aec-031ada70cd19" />
<img alt="Top nav bar issue in side menu context" width="1206" src="https://github.com/user-attachments/assets/044ca05f-dca7-4387-ae71-7a48e63a9adf" />
<img alt="UI inconsistency - nav bar and content overlap" width="1206" src="https://github.com/user-attachments/assets/8b899b2d-5085-4a5f-9403-6abf4297eff2" />

**Additional context**
Refer to the attached screenshots for different states and contexts where the navigation bar breaks or looks inconsistent.

prachi01645:
Hi, I would like to work on this issue.
I am setting up the project locally and will share updates soon.

amreshanand:
@Saksham-Sirohi  I have resolved this issue. Please look through it

Revanth-coder1:
Hi! I would like to work on this issue. Could you please assign it to me?

Saksham-Sirohi:
Parts of this are fixed, the problem with the navbar and items in the navbar still remains more prominent on the dev.eventyay.com homepage

justkrypticAsh:
Hi @Saksham-Sirohi, I noticed that parts of this issue have already been fixed, but the navbar inconsistency is still visible, especially on the dev.eventyay.com homepage.
I would like to work on fixing the remaining issues and ensuring consistency across all pages without affecting the desktop UI. Please let me know if I can proceed or if there are any specific guidelines I should follow.

Saksham-Sirohi:
[...]

PR #2546, opened 2026-02-24
Fix sidebar height so all items are accessible 

### LINKED ISSUE #2539: Sidebar navigation items are cut off and unreachable when menu sections are expanded
### Description:

Expanding multiple sections in the left sidebar causes the bottom navigation items (e.g. "Message center") to go off-screen. There is no way to scroll within the sidebar to reach them the only workaround is to collapse other expanded sections.

### Expected behavior: 

The sidebar should be scrollable so all navigation items remain accessible regardless of how many sections are expanded.

Image

Image

Aqil-Ahmad:
fixes #2539 

image

## Summary by Sourcery

Adjust sidebar layout to account for a 50px offset from the viewport height so the entire sidebar content remains accessible.

### DIFF
--- a/app/eventyay/static/orga/css/_layout.css
+++ b/app/eventyay/static/orga/css/_layout.css
@@ -182,7 +182,7 @@ aside.sidebar {
   background-color: var(--color-grey-ultralight);
   box-shadow: var(--shadow-lighter);
   width: 45px;
-  height: calc(100vh - 50px);
+  height: calc(100vh - var(--navbar-height));
   display: flex;
   font-size: 14px;
   flex-grow: 0;
@@ -190,7 +190,7 @@ aside.sidebar {
   flex-wrap: nowrap;
   z-index: 150;
   padding-left: 0;
-  margin-top: 50px;
+  margin-top: var(--navbar-height);
   margin-bottom: 0;
   overflow-y: auto;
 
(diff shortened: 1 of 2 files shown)


Aqil-Ahmad:
> Please consider sourcery feedback comment.

the hardcoded 50px is consistent with how it's used throughout the rest of the file

app/eventyay/static/orga/css/_layout.css
```diff
@@ -182,7 +182,7 @@ aside.sidebar {
   background-color: var(--color-grey-ultralight);
   box-shadow: var(--shadow-lighter);
   width: 45px;
-  height: 100vh;
+  height: calc(100vh - 50px);
```
ArnavBallinCode: using `top: 50px` would be a cleaner fix and will work.

Aqil-Ahmad: this won't work as sidebar doesn't have position: fixed

ArnavBallinCode: yes, correct . please add a screenshot with mobile view and then good to go.

Aqil-Ahmad:
here you go @ArnavBallinCode 

image

Aqil-Ahmad:
> This approach is not correct, as it is causing problems on the mobile view

hey @Saksham-Sirohi i tested it and i don't see any issues caused by the changes in the sidebar. if you are referring to the white text, that's unrelated and its the navbar issue( documented here https://github.com/fossasia/eventyay/issues/2462) which appears both with these changes and without.

ArnavBallinCode (approved):
LGTM

Saksham-Sirohi:
Screenshot 2026-03-18 at 4 32 53 PM

Issue #2560, opened 2026-02-25
Sidebar collapse toggles break due to duplicate id targets in nav include


SxxAq:
## Summary
The orga sidebar has collapsible sections where clicking the arrow rotates the icon, but the submenu does not open (or opens the wrong section). This is a frontend code bug caused by duplicate collapse target IDs in repeated template includes.

## Affected area
- Orga sidebar navigation (Talks / CFP / related sidebar sections with arrow toggles)

## Current behavior
- Clicking a sidebar arrow updates arrow state (`aria-expanded` / icon rotation).
- Submenu does not expand in the clicked section, or an unrelated section is affected.

## Expected behavior
- Each sidebar arrow should expand/collapse its own submenu reliably.
- No cross-section interference.

## Steps to reproduce
1. Open orga interface with sidebar groups that include collapsible children.
2. Click an arrow on a collapsible sidebar item.
3. Observe arrow state changes.
4. Observe submenu does not open (or incorrect panel behavior).

### 1) Duplicate IDs generated by shared include
In [sidebar_nav.html](/home/shxdow/Oss/eventyay/app/eventyay/orga/templates/orga/includes/sidebar_nav.html:13), collapse IDs are generated with:
- `id="collapseSidebar{{ forloop.counter }}"`
- `data-target="#collapseSidebar{{ forloop.counter }}"`

This partial is included multiple times from [base.html](/home/shxdow/Oss/eventyay/app/eventyay/orga/templates/orga/base.html:436), [base.html](/home/shxdow/Oss/eventyay/app/eventyay/orga/templates/orga/base.html:459), and [base.html](/home/shxdow/Oss/eventyay/app/eventyay/orga/templates/orga/base.html:474).  
Because each loop restarts `forloop.counter`, duplicate IDs are produced in the same DOM.

### 2) Collapse JS only selects first matching element
In [collapse.js](/home/shxdow/Oss/eventyay/app/eventyay/static/common/js/collapse.js:24):
- `document.querySelector(element.getAttribute('data-target'))`

With duplicate IDs, `querySelector` binds to the first match, so the clicked controller may toggle a different element than intended.

### 3) `aria-controls` mismatch
In [sidebar_nav.html](/home/shxdow/Oss/eventyay/app/eventyay/orga/templates/orga/includes/sidebar_nav.html:13), `aria-controls` is set to:
- `collapse{{ forloop.counter }}`
while the actual target ID uses:
[...]

PR #2561, opened 2026-02-25
Fix orga sidebar collapse

SxxAq:
This PR fixes the orga sidebar collapse regression where clicking a Talks/CFP sidebar arrow rotated the icon (`aria-expanded` changed) but the submenu did not open, or the wrong submenu was toggled.

### Root cause
The bug came from multiple interacting issues:
1. Sidebar panels were rendered with legacy `collapse in...` classes, while current collapse logic toggles `show`.  
2. Reused sidebar include blocks generated non-unique collapse IDs (`forloop.counter` resets per include context).  
3. Collapse target resolution used a global `document.querySelector(data-target)`, which can select the first matching element when IDs collide.  
4. `aria-controls` in the include didn’t match the actual panel ID pattern.

### What this PR changes
- Removes legacy `in` from sidebar collapse containers in orga templates so visibility is controlled consistently by `show`.
- Namespaces collapse IDs for included sidebar nav blocks by passing a `nav_prefix` from each include site.
- Updates include markup so `data-target`, `id`, and `aria-controls` all reference the same namespaced identifier.
- Hardens collapse controller logic in `common/js/collapse.js`:
  - prefers resolving target inside the same `.nav-fold` (local scope first),
  - falls back to global selector only when needed,
  - prevents default click navigation behavior,
  - initializes aria state for both open and closed panels.

### Files updated
- `app/eventyay/orga/templates/orga/base.html`
- `app/eventyay/orga/templates/orga/includes/sidebar_nav.html`
- `app/eventyay/static/common/js/collapse.js`

### Result
- Sidebar arrows now reliably toggle their intended submenu.
- No cross-panel toggling due to duplicated/ambiguous targets.
- Collapse state and accessibility attributes remain synchronized with actual panel visibility.

## Summary by Sourcery

Fix sidebar collapse behavior in the organizer interface so that each arrow reliably toggles its own submenu and ARIA state matches visibility.

Bug Fixes:
- Resolve regression where clicking Talks/CFP sidebar arrows rotated the icon but failed to open the correct submenu.
- Prevent cross-targeting of collapse panels when sidebar includes reuse the same collapse IDs.

Enhancements:
- Standardize collapse containers to rely on the Bootstrap-style `show` class instead of legacy `in` for visibility control.
- Namespace dynamic sidebar collapse IDs via a per-include prefix to keep targets unique across event, organizer, and global nav blocks.
- Scope collapse target lookup to the local nav section before falling back to a global query, and prevent default click navigation while initializing correct ARIA state.

- Closes: #2560

### DIFF
--- a/app/eventyay/orga/templates/orga/base.html
+++ b/app/eventyay/orga/templates/orga/base.html
@@ -232,7 +232,7 @@
                                         <i class="fa fa-angle-down"></i>
                                     </a>
                                 </span>
-                                <div class="collapse{% if "settings." in url_name %} show{% endif %}" aria-expand="true" id="collapseSettings">
+                                <div class="collapse{% if "settings." in url_name %} show{% endif %}" id="collapseSettings">
                                     <a href="{{ request.event.orga_urls.settings }}" class="nav-link nav-link-second-level{% if "settings.event." in url_name %} active{% endif %}">
                                         <span>{{ phrases.base.general }}</span>
                                     </a>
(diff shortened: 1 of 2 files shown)


SxxAq:
https://github.com/user-attachments/assets/dd6a1582-8cad-4083-ae50-e109f99906d0

Saksham-Sirohi:
Closing as the issue is already resolved with other issues in #2565

PR #2582, opened 2026-02-26
Fix pytest collection by removing pretalx imports from talk conftestFix talk test pretalx imports

### LINKED ISSUE #2292: Talk test suite fails due to unresolved pretalx imports
## Description

The `tests/talk` test suite fails during collection due to stale imports from the `pretalx` package (e.g. `pretalx.event.models`, `pretalx.common.models.settings`) even though the project is running against `eventyay` equivalents.

This results in `ModuleNotFoundError: No module named 'pretalx'` during pytest collection, preventing any of the affected tests from running.

## Expected behavior

Pytest should be able to collect and run `tests/talk` without failing on imports where eventyay already provides an equivalent implementation.

## Additional context

Eventyay is a fork of pretalx, and several test files still reference pretalx modules that have since been migrated or renamed under `eventyay.base.*`.

> Note: This issue focuses on resolving pretalx imports in the talk test suite. Other remaining pretalx references elsewhere in the codebase can be addressed incrementally as follow-up refactors, once clear Eventyay equivalents or intended behavior are defined.

sathwikre:
This PR removes stale Pretalx-based fixtures from app/tests/talk/conftest.py.

Pretalx is not a dependency of Eventyay, and these imports caused pytest to fail
during test collection.

Fixes #2292


mariobehling:
Thanks a lot for working on this and for tackling the issue around the `pretalx.*` imports. The underlying problem you are addressing is definitely valid and important.

After reviewing the changes, we think the current approach goes a bit too far by removing the shared talk fixtures entirely. These fixtures are still used across the test suite, so this would lead to new test collection errors instead of resolving the original issue.

Because of this, we are closing this PR for now.

A better approach would be to:

* migrate the existing fixtures from `pretalx.*` to the current Eventyay equivalents, or
* move them into a shared fixture module and update the affected tests in the same PR

This way, we fix the import issue while keeping the test suite working.

We would be very happy to review a new PR following this approach. Thanks again for your contribution and effort on this.

PR #2617, opened 2026-02-27
fix: add type hints and https fallback to get_protocol

### LINKED ISSUE #2616: Fix: Handle empty URL schemes and add type hints to `get_protocol`
## Summary

While reviewing `app/eventyay/api/utils.py`, I noticed that the `get_protocol` function lacks type annotations and doesn't handle cases where a schemeless URL or `None` is passed.

Currently, `urlparse` returns an empty string for `parsed.scheme` in these edge cases. This causes downstream callers (like in event creation) to generate and save malformed domains (e.g., `://host/path`) silently, without raising any errors.

## Proposed Fix
I'd like to update this utility by:
1. Adding strict `str` type hints to the function signature.
2. Implementing a safe fallback (like returning `'https'`) when the parsed scheme is empty to prevent the silent failure.

DarkBird10020:
While reviewing the API utilities, I noticed that `get_protocol()` in `app/eventyay/api/utils.py` lacked type annotations and returned an empty string when a URL had no scheme. 

**The Problem:**
This behavior caused `CreateEventView.post()` to silently save malformed domains (e.g., `://host/path`) to the event object without any error handling.

**Problem Fixed**
- Added `str` type hints to the function signature for better type safety.
- Added a guard clause to return `'https'` as a safe default when `parsed.scheme` is empty.
- Ensured valid schemes like `http` are still preserved for local development.

Fixes #2616

## Summary by Sourcery

Ensure URL protocol parsing returns a lowercase scheme with a safe HTTPS default when missing.

Bug Fixes:
- Prevent malformed URLs from propagating by defaulting to HTTPS when no scheme is present in get_protocol().

Enhancements:
- Add explicit string type hints to get_protocol() for improved type safety.


norbusan:
@DarkBird10020 please see the suggestion by copilot, which is valid. Your code is as is broken.

ArnavBallinCode:
Closing this pr as it has been inactive since over a month , please open a fresh pr / look into open issues for further contributions

PR #2626, opened 2026-02-28
Fix regexp with backtrack ability that run on user supplied data

norbusan:
- RE_CHROMIUM_VERSION
  change Chrom[^ /]+[ /] to Chrom(?:e|ium)[ /]
  the + quantifier would scan all chars before failing O(n), only match exact browser names

- RE_IOS_VERSION
  change  .+; CPU .*OS to [^;]+; CPU [^)]*OS and trailing .* to [^)]*
  .+ eats everyting, including ; same with .*OS, multiple paths for matching. restrict matches

- RE_MAC_VERSION
  change \(Macintosh;.*Mac OS X to \(Macintosh;[^)]*Mac OS X and trailing .*\) to [^)]*\)
  as above, .* matches too much

- RE_SAFARI
  change Version/.* to Version/[^ ]*
  dont scoop up everything

## Summary by Sourcery

Bug Fixes:
- Refine Chromium, iOS, macOS, and Safari User-Agent regex patterns to prevent pathological backtracking on crafted input.


Sak1012:
@norbusan CodeQL has 3 alerts marked as **3 high severity security vulnerabilities** please take a look.

norbusan:
Yup, that is tricky due to the apple UA string format. 
Maybe we should use something else. I need to think

Saksham-Sirohi (changes requested):
@norbusan any updates on this pr if this has already been completed, maybe we can close this

PR #2632, opened 2026-02-28
Fix account settings form password validation bug

AgnusSerena:
Remove email field from UserSettingsForm since email is managed through a separate 'Manage email addresses' page. This field was in the form but not rendered in the template, causing form validation issues.

The bug occurred because:
- Email field was in form.fields but not in POST data
- Clean method incorrectly checked if email was being changed
- This caused false positives requiring password for fullname-only changes

Changes:
- Remove email from UserSettingsForm.Meta.fields
- Remove email field validation logic from form's __init__ and clean methods
- Update GeneralSettingsView to get email from user object instead of form
- Update error message to only mention password changes (not email)

This fixes the issue where users couldn't update just their fullname without being prompted for their current password.

## Summary by Sourcery

Fix account general settings so that only password changes require current password verification and email is no longer handled in this form.

Bug Fixes:
- Allow updating fullname and other non-sensitive settings without incorrectly requiring the current password.
- Ensure security notices use the user’s stored email address since email is managed separately from the general settings form.

Enhancements:
- Remove the email field and related validation from the general account settings form to reflect that email is managed on a dedicated page.
- Clarify the current password help text to refer only to password changes.

Tests:
- Add integration tests for the general account settings form and view, covering fullname-only updates and password change flows without an email field.

### DIFF
--- a/app/eventyay/common/templates/common/avatar.html
+++ b/app/eventyay/common/templates/common/avatar.html
@@ -10,6 +10,11 @@
             <div>
                 <div class="avatar-upload hide-label">
                     {{ form.avatar.as_field_group }}
+                    {% if form.avatar.help_text %}
+                        <small class="form-text text-muted d-block mt-2">
+                            {{ form.avatar.help_text|safe }}
+                        </small>
+                    {% endif %}
                 </div>
                 {{ form.get_gravatar.as_field_group }}
             </div>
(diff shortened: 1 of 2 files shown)


Saksham-Sirohi:
Closing as the issue is already resolved. Thank you for your contribution

PR #2698, opened 2026-03-07
fix: replace generic Exception with PermissionDenied in check_access_permissions

MEGHANA660:
## Problem
In `check_access_permissions()` in `organizer.py`, two places 
raised a generic `Exception` when permission checks failed.

## Fix
Replaced both `raise Exception(...)` with `raise PermissionDenied(...)` 
from `django.core.exceptions` — the semantically correct exception 
for permission failures in Django.

## Changes
- `app/eventyay/base/models/organizer.py`: Replaced 2 generic 
  exceptions with PermissionDenied and added import

## Summary by Sourcery

Ensure BigBlueButton room creation uses a valid active server and raise appropriate errors for missing servers and permission issues.

Bug Fixes:
- Handle cases where no active BigBlueButton server is available by logging the failure and raising a specific error instead of using an inactive server.
- Raise Django's PermissionDenied in organizer access checks instead of a generic Exception when permission requirements are not met.

Enhancements:
- Refine BigBlueButton call server selection logic to reuse or create calls only when a valid active server is available, improving robustness of room creation.

### DIFF
--- a/app/eventyay/base/models/organizer.py
+++ b/app/eventyay/base/models/organizer.py
@@ -14,6 +14,7 @@ from django.utils.functional import cached_property
 from django.utils.timezone import get_current_timezone, make_aware, now
 from django.utils.translation import gettext_lazy as _
 from django_scopes import scope
+from django.core.exceptions import PermissionDenied
 from rules.contrib.models import RulesModelBase, RulesModelMixin
 
 from eventyay.base.models.base import LoggedModel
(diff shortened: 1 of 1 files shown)


mariobehling:
Thanks for your contribution.

We are closing this PR for now because it combines multiple unrelated changes, which makes it difficult to review and assess properly. In addition, there is no linked issue that clearly describes the problem and intended outcome.

To help move this forward, we suggest:

* opening an issue first with a clear description of the problem and expected behavior
* creating a focused PR that addresses one concern at a time

We would be happy to review a new PR once it is well scoped and linked to an issue. This helps keep the codebase maintainable and the review process efficient for everyone.

Thanks again for your effort and looking forward to your updated contribution.

PR #2699, opened 2026-03-07
fix: add exception handling for missing user in update_user

MEGHANA660:
## Problem
In `update_user()` in `services/user.py`, the `.get()` call had 
a TODO comment noting missing exception handling. If a user ID 
does not exist, it raises an unhandled `User.DoesNotExist` error.

## Fix
Wrapped the `.get()` call in a try/except block. Returns `None` 
with a warning log when the user is not found, instead of crashing.

## Changes
- `app/eventyay/base/services/user.py`: Added try/except around 
  User.objects.get() in update_user()

## Summary by Sourcery

Bug Fixes:
- Return None and log a warning when update_user is called with a non-existent user instead of raising User.DoesNotExist.

### DIFF
--- a/app/eventyay/base/services/user.py
+++ b/app/eventyay/base/services/user.py
@@ -243,12 +243,15 @@ def create_user(
 def update_user(
     event_id, id, *, traits=None, data=None, is_admin=False, serialize=True
 ):
-    # TODO: Exception handling
-    user = (
-        User.objects.select_related("event")
-        .select_for_update()
-        .get(id=id, event_id=event_id)
-    )
+    try:
+        user = (
+            User.objects.select_related("event")
+            .select_for_update()
+            .get(id=id, event_id=event_id)
+        )
+    except User.DoesNotExist:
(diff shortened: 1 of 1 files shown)


mariobehling:
Thanks for your contribution.

We are closing this PR for now as there is no linked issue describing the problem being addressed. Changes like this should be driven by a clearly defined issue with a concrete use case and expected behavior.

In addition, this PR changes the behavior of `update_user()` by swallowing `User.DoesNotExist` and returning `None`, which alters the contract of the function without justification or corresponding updates in calling code or tests. This makes the change difficult to evaluate and potentially unsafe.

Please open an issue first describing the problem you are trying to solve. Then create a new, focused PR that clearly addresses that issue, includes proper validation of the behavior, and ensures consistency across all affected code paths.

Also, please ensure Copilot is enabled on your account so automated reviews can run, helping to streamline the review process.

Thanks again, and we are happy to review a revised PR once it is properly scoped and justified.

PR #2700, opened 2026-03-07
Start generating the badge already during checkin

norbusan:


Sak1012:
The initial issue was checkin app polling the system from mande which did not check for the blob response, it has been now solved in https://github.com/fossasia/eventyay-checkin/pull/60.

PR #2703, opened 2026-03-07
fix: Allow identical ticket IDs across different events

### LINKED ISSUE #575: Allow tickets with the same ID in different events.

OsauravO:
This fixes #575 
## Problem

`OrderPosition.pseudonymization_id` has a global `unique=True` constraint at the database level. This means no two ticket positions anywhere on the platform can share the same pseudonymization ID, even if they belong to completely different events.

This causes problems when:

- **Importing tickets** from external systems — IDs can collide with unrelated events
- **Migrating or cloning events** — regeneration is forced even when IDs would be perfectly valid in the new event context
- **Scaling the platform** — as the number of positions grows, the global collision-avoidance loop becomes less efficient

In practice, `pseudonymization_id` is only ever looked up within a single event's context (lead scanning, search, export, PDF generation). Global uniqueness provides no functional benefit and introduces unnecessary cross-event coupling.

## Solution

Scope `pseudonymization_id` uniqueness to the event level, matching how `secret` (the ticket barcode) is already handled.

### Changes

`**app/eventyay/base/models/orders.py**`

- Remove `unique=True` from the `pseudonymization_id` field definition (keep `db_index=True` for query performance).
- Update `assign_pseudonymization_id()` to check for collisions within the event (`order__event=self.order.event`) instead of globally. Also remove the redundant nested `scopes_disabled()` context manager since the method is already decorated with `@scopes_disabled()`.

`**app/eventyay/base/services/orderimport.py**`

- Remove the early `position.assign_pseudonymization_id()` call at line 120. This call happened before `position.order` was set, which would crash now that the method needs `self.order.event`. The removal is safe because `OrderPosition.save()` already calls `assign_pseudonymization_id()` as a fallback when the field is empty, and at that point `position.order` is properly assigned.

`**app/eventyay/base/migrations/0020_scope_pseudonymization_id_uniqueness_to_event.py**`

- New migration that drops the global unique constraint from `pseudonymization_id`.

### What does NOT change

- `**secret field**` — already event-scoped in all three validation paths (`save()`, CSV import, API serializer). No changes needed.
- **Checkin logic** — already queries within event-scoped checkin lists.
- **PDF/QR generation, export, search** — all already filter by event context.
- **Behavior within a single event** — uniqueness is still enforced per-event by the application logic.

## Analysis of all creation paths

| Path                        | Event lock           | Transaction                  | Race-safe                                                                                                                         |
| --------------------------- | -------------------- | ---------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
| Web checkout (cart → order) | Yes (`event.lock()`) | Yes                          | Yes                                                                                                                               |
| CSV import                  | Yes (`event.lock()`) | Yes                          | Yes                                                                                                                               |
| API order creation          | No                   | Yes (`transaction.atomic()`) | Auto-generated IDs: yes (high entropy). Explicit IDs: same TOCTOU risk as `secret` — pre-existing, not introduced by this change. |
| Secret regeneration         | No                   | Yes (`@transaction.atomic`)  | Yes (operates on existing positions)                                                                                              |

## Summary by Sourcery

Scope ticket pseudonymization IDs to be unique per event instead of globally, and align generation/import behavior with this event-level uniqueness.

Bug Fixes:
- Allow identical pseudonymization IDs to be reused across different events without database constraint violations.
- Prevent crashes during order import caused by assigning pseudonymization IDs before the order (and its event) is attached.

Enhancements:
- Restrict pseudonymization ID collision checks to the current event when generating new IDs for order positions.

Chores:
- Add a database migration to drop the global unique constraint from the pseudonymization_id field on order positions.


OsauravO:
@mariobehling can you please review this

hongquan (changes requested):
Removing `unique=True` alone is not enough. The requirement is:

>  Ticket IDs are unique per event

You need to add [`Unique`](https://docs.djangoproject.com/en/6.0/ref/models/constraints/#uniqueconstraint) constraint in the `Meta` class to enforce this.

OsauravO:
@hongquan 
Thanks for the review. I understand that this alone is insufficient as we need a db level `UniqueConstraint` scoped to the event.
Since `UniqueConstraint` can only reference columns on the same table (not joined fields like `order__event`), the only way to enforce per-event uniqueness at the DB level is to add a concrete `event` FK directly on `OrderPosition`. This follows the existing pattern which is in CartPosition (the sibling model) already has a direct `event` FK. The migration would add the field as nullable, backfill from `Order.event_id`, then make it non-null and add the constraint.

Should I proceed with this approach?

hongquan:
Postpone it. We need to make clear what "Ticket ID" is. We don't have a dedicated field for "Ticket ID" and there are many fields that look like can be used for "Ticket ID".

The `OrderPosition.pseudonymization_id` is generated in our system when importing, it is not the incoming data:

https://github.com/fossasia/eventyay/blob/d16b93249ec9116a796c2fdd74c47bd9e3faa300/app/eventyay/base/services/orderimport.py#L120

so it doesn't look like "Ticket ID".

ArnavBallinCode (commented):
Thank you for your contributions, closing this pr as doesnt seem to be active for about 2 months, feel free to re-open if you plan to proceed with it

PR #2710, opened 2026-03-08
Allow duplicate order codes across different events

### LINKED ISSUE #575: Allow tickets with the same ID in different events.

SergioChan:
## Summary
- scope manual `code` uniqueness validation in `OrderCreateSerializer` to the current event instead of organizer-wide scope
- keep same-event duplicate prevention unchanged
- aligns order import/API behavior with issue #575 expectation for independent events

## Testing
- `python -m py_compile app/eventyay/api/serializers/order.py`
- Could not run project tests in this environment because Django dependencies are not installed (`ModuleNotFoundError: No module named 'django'` when invoking pytest)

## Related
Fixes #575

### DIFF
--- a/app/eventyay/api/serializers/order.py
+++ b/app/eventyay/api/serializers/order.py
@@ -1109,8 +1109,11 @@ class OrderCreateSerializer(I18nAwareModelSerializer):
         return channel
 
     def validate_code(self, code):
-        if code and Order.objects.filter(event=self.context['event'], code=code).exists():
-            raise ValidationError('This order code is already in use.')
+        event = self.context.get('event')
+        if event is None:
+            raise ValidationError('Event context is required for order code validation.')
+        if code and Order.objects.filter(event=event, code=code).exists():
+            raise ValidationError('This order code is already in use for this event.')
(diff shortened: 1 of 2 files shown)


SergioChan:
Addressed the review suggestions in the latest commit:\n\n- Updated the duplicate-code error to clarify event scope (`This order code is already in use for this event.`).\n- Switched to `self.context.get("event")` with an explicit validation error when event context is missing, to avoid an opaque KeyError path.\n- Added coverage for cross-event reuse of the same order code (same code accepted in a different event).\n\nValidation:\n- `python -m compileall -q app/eventyay/api/serializers/order.py app/tests/tickets/api/test_orders.py`\n- Full pytest run wasn’t possible in this environment because Django dependencies are not installed locally.\n

ArnavBallinCode (commented):
please attach screenshots / screen-recording demonstrating your implementation.

ArnavBallinCode:
Closing pr as the pr description shows the user is not able to run the system / test the implementation created by him. 
Feel free to re open the pr if otherwise. thanks

PR #2735, opened 2026-03-09
fix: block poll updates after close or archive state

MEGHANA660:
## Problem
`update_poll()` had a TODO comment questioning whether updates 
should be blocked after a poll is closed or archived.

## Fix
Added a state check at the start of `update_poll()`. Raises 
ValueError if poll is in CLOSED or ARCHIVED state, preventing 
unintended modifications to finalized polls.

## Changes
- `app/eventyay/base/services/poll.py`: Added state guard 
  in update_poll()

## Summary by Sourcery

Bug Fixes:
- Prevent updates to polls that are in CLOSED or ARCHIVED state by raising an error instead of allowing modifications.

### DIFF
--- a/app/eventyay/base/services/poll.py
+++ b/app/eventyay/base/services/poll.py
@@ -62,10 +62,12 @@ def get_polls(room, moderator=False, early_results=False, for_user=None, **kwarg
     ]
 
 
+# AFTER:
 @database_sync_to_async
 def update_poll(**kwargs):
-    # TODO: do we want to block updates after close/archive?
     poll = Poll.objects.get(pk=kwargs["id"], room=kwargs["room"])
+    if poll.state in (Poll.States.CLOSED, Poll.States.ARCHIVED):
+        raise ValueError("Cannot update a poll that is closed or archived.")
     options = kwargs.pop("options", None)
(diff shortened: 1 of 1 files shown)


mariobehling:
Thanks for the contribution.

We are closing this PR for now because there is no linked issue describing the problem, use case, and expected behavior. This repository expects changes to be driven by a clearly defined issue first. 

I reviewed the actual diff. This PR adds a guard in `update_poll()` to block updates when a poll is in `CLOSED` or `ARCHIVED` state and raises a generic `ValueError` with a hard-coded message. The PR description says this was based on an internal TODO comment, but there is no issue discussing whether this is the intended product behavior, how callers should handle the exception, or whether this should use an existing domain-specific error flow instead. 

The automated reviews also raised the same implementation concern: a generic `ValueError` is probably not the right contract here, and a user-facing hard-coded error message may not fit the project’s normal error handling and translation approach.

Please open an issue first and describe:

* the exact problem
* where the current behavior is wrong
* the expected behavior after a poll is closed or archived
* how the API or UI should surface that error

After that, please open a new, focused PR linked to the issue. That will make the change much easier to review and evaluate.

PR #2742, opened 2026-03-10
Fix: incorrect queryset assignment and invalid field lookup in product_meta_value

### LINKED ISSUE #2741: Bug: Incorrect queryset assignment and invalid field lookup in `product_meta_values`
## Description

In the `product_meta_values` view, there is a bug in the access restriction logic when `all_access` is `False`.

### Problematic Code

```python
if not all_access:
    defaults = matches.filter(
        event__id__in=request.user.teams.filter(can_change_items=True)
        .values_list('limit_events__id', flat=True)
    )
```

#### 1. Invalid field lookup

`ProductMetaValue` does not have a direct `event` field.

Model structure:

```
ProductMetaValue
 └── product (FK)
      └── event (FK)
```

Correct lookup should be:

```
product__event__id
```

#### 2. Incorrect queryset assignment

`defaults` is originally a queryset of `ProductMetaProperty`:

```python
defaults = ProductMetaProperty.objects.filter(...)
```

But it is reassigned to:

```
matches.filter(...)
```

which returns a queryset of `ProductMetaValue`. This changes the expected queryset type and may break later logic.

### Evidence

Manual verification in Django shell:

```python
ProductMetaValue.objects.filter(event__id=1)
```

Result:

```
FieldError: Cannot resolve keyword 'event'
```

Correct query:

```python
ProductMetaValue.objects.filter(product__event__id=1)
```

### Expected Fix

The filtering should be applied separately:

```python
if not all_access:
    defaults = defaults.filter(
        event__id__in=request.user.teams.filter(can_change_items=True)
        .values_list('limit_events__id', flat=True)
    )

    matches = matches.filter(
        product__event__id__in=request.user.teams.filter(can_change_items=True)
        .values_list('limit_events__id', flat=True)
    )
```

### Location

`eventyay/control/views/select2.py`
Function: `product_meta_values`

---

keshviagrawal:
Found a bug in product_meta_values where non-admin users were getting a 500 error.

The main issues were:
The filter was using event__id on a model that doesn't have it (needs to go through the product).
defaults was being overwritten by a different queryset type, which broke things later in the function.

I've separated the filters and used the correct product__event__id path. Also cleaned it up by putting the team's event IDs into a single variable to avoid hitting the DB twice.

Additionally, I added organizer-level filtering (organizer=organizer) to ensure that users only access product meta values belonging to their respective organizer.

Previously, team members associated with multiple organizers could potentially access data across events from different organizers when filtering by limit_events. This change improves data isolation and security alongside fixing the 500 error.
Fixes #2741

## Summary by Sourcery

Fix access-controlled product meta value lookups to avoid server errors for non-admin users.

Bug Fixes:
- Correct the queryset filters in product_meta_values to use the proper event relationship and avoid invalid field lookups.
- Ensure the defaults queryset is filtered correctly without being overwritten by an incompatible queryset type for non-admin users.

Enhancements:
- Reuse a single computed list of user-accessible event IDs when filtering product meta values to reduce duplicate database queries.


Rachit7168:
Hi @keshviagrawa

 include the related issue number in the PR description and not in the PR title itslef Maintainers usually recommend linking the issue,Just a small suggestion to align with the contribution guidelines

keshviagrawal:
Thanks for the suggestion @Rachit7168! I've updated the title to follow the contribution guidelines.

shivam-pawar-7217:
Good fix on the `product__event__id__in field path`  that was definitely causing the FieldError on non-admin users.

One thing worth highlighting explicitly: the new code adds `organizer=organizer` to the team filter, which wasn't in the original. Without it, a user who's a team member across multiple organizers could see product meta values for events from other organizers when filtering by` limit_events`. The new scope fix is correct  just worth documenting in the PR description since it's a data isolation improvement on top of the 500 fix.

keshviagrawal:
Thanks @shivam-pawar-7217 for pointing that out! I've clarified the organizer-level filtering and its impact on data isolation here.

mariobehling (changes requested):
Please provide a screencast showing that the fix works on the system itself. Thanks.

keshviagrawal:
https://github.com/user-attachments/assets/bc91e37e-c397-4c89-acdb-eeeb9501f06c


Added a screencast demonstrating the fix.

The video shows:

Non-admin user accessing the product page
Page loads successfully without any 500 error
Network requests returning 200 status

app/eventyay/static/jsi18n/en/djangojs.js
```diff
@@ -93,10 +93,10 @@
     "FIRST_DAY_OF_WEEK": 0,
     "MONTH_DAY_FORMAT": "F j",
     "NUMBER_GROUPING": 3,
-    "SHORT_DATETIME_FORMAT": "m/d/Y P",
-    "SHORT_DATE_FORMAT": "m/d/Y",
+    "SHORT_DATETIME_FORMAT": "Y-m-d H:i",
+    "SHORT_DATE_FORMAT": "Y-m-d",
```
Rachit7168, line 97: Are these changes needed here for the particular issue ?

Rachit7168 (commented):
Please resolve below reviews

Rachit7168:
Thanks for your contribution. Looking forward to future contributions
Closing as already resolved via #3438

PR #2748, opened 2026-03-13
fix: add timeout to all outbound HTTP requests to prevent worker hangs

### LINKED ISSUE #2747: Missing timeout on outbound HTTP requests can hang workers indefinitely
I was going through the codebase and noticed that several `requests.post()` and `requests.get()` calls don't specify a `timeout` parameter. Without it, these calls will
  block forever if the remote server accepts the connection but never responds.

  The most concerning one is in `api/webhooks.py`, the `target_url` is set by organizers, so a slow or malicious endpoint could tie up a Celery worker permanently.

  **Affected locations**
```

  - `app/eventyay/api/webhooks.py` (line ~316) — webhook delivery, user-provided URL
  - `app/eventyay/base/services/update_check.py` (line ~59) — periodic update check
  - `app/eventyay/base/services/mail.py` (line ~669) — mail-related HTTP call
  - `app/eventyay/features/analytics/graphs/utils.py` (line ~104)
  - `app/eventyay/features/social/utils.py` (line ~27)
```

 **Example**

  ```python
  **current code in webhooks.py**
  resp = requests.post(webhook.target_url, json=payload, allow_redirects=False)

  **should be**
  resp = requests.post(webhook.target_url, json=payload, allow_redirects=False, timeout=30)

**Why this matters**

  - A hanging request holds a Celery worker (or web thread) hostage with no way to recover
  - The webhook case is the worst since target_url comes from user input
  - If enough workers get stuck, the whole task queue stops processing
  - Python's requests library docs explicitly recommend always setting a timeout

CallMeAtif:
Fixes #2747

Several requests.get() and requests.post() calls were missing the timeout parameter, which means they could block indefinitely if a remote server accepts the connection but never responds.

This is especially critical for webhook delivery where the target URL is user-provided, a malicious or slow endpoint could tie up Celery workers permanently.

Added appropriate timeouts (15-30s depending on use case) across 14 files including webhooks, OAuth flows, geocoding, update checks, mail services, and external storage.

## Changes

  Added explicit timeouts (15–30s depending on use case) to all outbound HTTP calls across 14 files:

  - `api/webhooks.py` — webhook delivery (user-provided URL)
  - `base/services/update_check.py` — periodic update check
  - `base/services/mail.py` — mail image fetching
  - `agenda/views/talk.py` — ticket check API call
  - `control/views/geo.py` — geocoding (OpenCage, MapQuest)
  - `core/management/commands/bbb_update_cost.py` — BBB server polling
  - `eventyay_common/tasks.py` — internal webhook + video world creation
  - `orga/tasks.py` — schedule sync to tickets component
  - `features/analytics/graphs/utils.py` — schedule widget fetch
  - `features/social/views/linkedin.py` — LinkedIn OAuth
  - `features/social/views/twitter.py` — Twitter/X OAuth
  - `features/social/utils.py` — avatar download
  - `features/importers/conftool.py` — ConfTool schedule/poster import
  - `features/integrations/platforms/storage/nanocdn.py` — CDN read/write/delete

  ## Timeout values

  - **30s** for webhooks (user-provided URLs, needs tolerance for slow servers), file uploads, and large data imports
  - **15s** for internal API calls, OAuth flows, and image downloads

## Summary by Sourcery

Add explicit timeouts to all affected outbound HTTP requests to prevent indefinite blocking when remote services are slow or unresponsive.

Bug Fixes:
- Prevent potential worker hangs and indefinite blocking by adding timeouts to webhook deliveries, schedule syncs, update checks, and other HTTP-dependent operations.

Enhancements:
- Apply consistent 15s/30s timeout policy across external integrations including geocoding services, social OAuth flows, avatar and image fetching, analytics schedule retrieval, BBB server polling, ConfTool imports, and NanoCDN storage operations.

### DIFF
--- a/app/eventyay/agenda/views/talk.py
+++ b/app/eventyay/agenda/views/talk.py
@@ -653,7 +653,7 @@ def check_user_owning_ticket(user: User, event: Event) -> TicketCheckResult:
     api_url = urljoin(base_url, f'api/v1/{organizer_slug}/{event_slug}/ticket-check')
     logger.info('To call API %s', api_url)
     # In development, we disable the SSL verification.
-    response = requests.post(api_url, json=check_payload, verify=(not settings.DEBUG))
+    response = requests.post(api_url, json=check_payload, verify=(not settings.DEBUG), timeout=15)
 
     if response.status_code != HTTPStatus.OK:
         logger.debug('Response from eventyay-ticket: %s', response.text)
--- a/app/eventyay/api/webhooks.py
+++ b/app/eventyay/api/webhooks.py
@@ -313,7 +313,7 @@ def send_webhook(self, logentry_id: int, action_type: str, webhook_id: int):
 
         try:
             try:
-                resp = requests.post(webhook.target_url, json=payload, allow_redirects=False)
+                resp = requests.post(webhook.target_url, json=payload, allow_redirects=False, timeout=30)
                 WebHookCall.objects.create(
                     webhook=webhook,
                     action_type=logentry.action_type,
--- a/app/eventyay/base/services/mail.py
+++ b/app/eventyay/base/services/mail.py
@@ -666,7 +666,7 @@ def convert_image_to_cid(image_src, cid_id, verify_ssl=True):
             path = urlparse(image_src).path
             guess_subtype = os.path.splitext(path)[1][1:]
 
-            response = requests.get(image_src, verify=verify_ssl)
+            response = requests.get(image_src, verify=verify_ssl, timeout=15)
             mime_image = MIMEImage(response.content, _subtype=guess_subtype)
 
         mime_image.add_header('Content-ID', '<%s>' % cid_id)
--- a/app/eventyay/base/services/update_check.py
+++ b/app/eventyay/base/services/update_check.py
@@ -56,7 +56,7 @@ def update_check():
         'plugins': [{'name': p.module, 'version': p.version} for p in get_all_plugins()],
     }
     try:
-        r = requests.post('https://eventyay.org/.update_check/', json=check_payload)
+        r = requests.post('https://eventyay.org/.update_check/', json=check_payload, timeout=15)
         gs.settings.set('update_check_last', now())
         if r.status_code != 200:
             gs.settings.set('update_check_result', {'error': 'http_error'})
--- a/app/eventyay/control/views/geo.py
+++ b/app/eventyay/control/views/geo.py
@@ -38,7 +38,8 @@ class GeoCodeView(LoginRequiredMixin, View):
         gs = GlobalSettingsObject()
 
         r = requests.get(
-            'https://api.opencagedata.com/geocode/v1/json?q={}&key={}'.format(quote(q), gs.settings.opencagedata_apikey)
+            'https://api.opencagedata.com/geocode/v1/json?q={}&key={}'.format(quote(q), gs.settings.opencagedata_apikey),
+            timeout=15,
         )
         r.raise_for_status()
         d = r.json()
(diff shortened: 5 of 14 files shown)


shivam-pawar-7217:
Reviewed the diff across all 14 files - good coverage overall, nice to see this finally addressed.

A couple of things I noticed:

In` eventyay_common/tasks.py`, the first POST (webhook-style call) gets timeout=15, but `api/webhooks.py` uses timeout=30 for webhook delivery. If both are sending to `external/user-provided` URLs, they probably should use the same value? Or is the 15s one hitting an internal endpoint?

Also in `nanocdn.py` , `_read `and the upload POST both get timeout=30 which makes sense for potentially large files, but delete gets `timeout=15.` Deletes are usually just a header exchange, so 15s is probably fine, but just flagging it in case it was unintentional.

One general idea worth considering for a follow-up: defining these as named constants somewhere `(e.g. TIMEOUT_DEFAULT = 15, TIMEOUT_WEBHOOK = 30) `would make the intent clearer and easier to adjust globally later without hunting across 14 files. Not blocking, just a thought.

Otherwise looks solid - the 15/30s split makes sense for the use cases.

CallMeAtif:
> Reviewed the diff across all 14 files - good coverage overall, nice to see this finally addressed.
> 
> A couple of things I noticed:
> 
> In` eventyay_common/tasks.py`, the first POST (webhook-style call) gets timeout=15, but `api/webhooks.py` uses timeout=30 for webhook delivery. If both are sending to `external/user-provided` URLs, they probably should use the same value? Or is the 15s one hitting an internal endpoint?
> 
> Also in `nanocdn.py` , `_read `and the upload POST both get timeout=30 which makes sense for potentially large files, but delete gets `timeout=15.` Deletes are usually just a header exchange, so 15s is probably fine, but just flagging it in case it was unintentional.
> 
> One general idea worth considering for a follow-up: defining these as named constants somewhere `(e.g. TIMEOUT_DEFAULT = 15, TIMEOUT_WEBHOOK = 30) `would make the intent clearer and easier to adjust globally later without hunting across 14 files. Not blocking, just a thought.
> 
> Otherwise looks solid - the 15/30s split makes sense for the use cases.

Hey, thanks for going through all 14 files.                                                                                                                                
                                                                                                                                                                               
Yeah the one in eventyay_common/tasks.py is actually hitting an internal URL(reverse('eventyay_common:webhook.team')), not user-provided, so I kept it at 15s. Happy to bump it to 30 if you think it's better to keep all webhook-like calls consistent though.                                                                                         
                                                                                                                                                                               
 The nanocdn delete at 15s was intentional, figured a delete is just a quick header round-trip so it doesn't need the same tolerance as a file upload. Let me know if you'd rather keep it uniform at 30 across the whole storage backend.
                                                                                                                                                                               
 And yes the named constants idea makes sense, I'll save that for a follow-up to keep this one focused.

shivam-pawar-7217:
Makes sense on both counts , internal URL for the team webhook explains the 15s, no need to change it. And yeah, a delete being just a header round-trip is a fair distinction from a file upload. I'd leave them as-is.

Sounds good on the constants as a follow-up. Clean PR overall, good fix!

CallMeAtif:
@Saksham-Sirohi maybe review this change and if possible give some feedback.

Sak1012:
Thanks for the effort. Closing this PR as it addresses the symptom in a way that hurts maintainability more than it helps. The current change mostly inlines timeout=15 and timeout=30 across many unrelated call sites, which spreads magic numbers throughout the codebase without defining any shared policy for outbound HTTP requests. That makes future tuning and auditing harder, and it leaves timeout behaviour encoded in each file.

PR #2760, opened 2026-03-14
refactor: replace jQuery with Fetch API

### LINKED ISSUE #2759: Remove jQuery dependency from polling scripts
## Summary

Three small polling scripts under `static/pretixbase/js/` and `static/pretixcontrol/js/ui/` still use jQuery (`$.get`, `$.getJSON`) for simple HTTP requests that can be replaced with the native `fetch()` API. This is inconsistent with the project's coding guidelines.

## Problem

- `app/eventyay/static/pretixbase/js/reloadpending.js` uses `$.get()` (8 lines)
- `app/eventyay/static/pretixbase/js/ajaxpending.js` uses `$.getJSON()` (10 lines)
- `app/eventyay/static/pretixcontrol/js/ui/devices.js` uses `$.getJSON()` (14 lines)

Two of the three templates that use these scripts (`pretixbase/cachedfiles/pending.html` and `pretixpresale/waiting.html`) are standalone pages that load the **entire jQuery 3.7.1 library (~90KB)** only for these trivial polling calls. No other jQuery usage exists on those pages.

OsauravO:
## Summary

Fixes #2759

- Replace `$.get()` and `$.getJSON()` with native `fetch()` in three small polling scripts (`reloadpending.js`, `ajaxpending.js`, `devices.js`)
- Remove jQuery `<script>` include from two standalone pages (`pending.html`, `waiting.html`) that loaded the full ~90KB library solely for these trivial calls
- Add `.catch()` error handling that was absent in the original jQuery versions

## Changes

| File | What changed |
|------|-------------|
| `static/pretixbase/js/reloadpending.js` | `$.get()` → `fetch()` + `.text()` |
| `static/pretixbase/js/ajaxpending.js` | `$.getJSON()` → `fetch()` + `.json()` |
| `static/pretixcontrol/js/ui/devices.js` | `$.getJSON()` → `fetch()` + `.json()`, `$(fn)` → `DOMContentLoaded` |
| `pretixbase/cachedfiles/pending.html` | Removed jQuery script tag and `{% compress js %}` wrapper |
| `pretixpresale/waiting.html` | Removed jQuery script tag and `{% compress js %}` wrapper |

## Why

These three scripts are the only jQuery consumers on their respective pages. The project guidelines require new and updated JS to avoid jQuery and use native APIs. This change aligns the polling scripts with that policy.

## Test plan

- [ ] Visit a cached file download page (triggers `pending.html`) — verify the page auto-reloads when the file is ready
- [ ] Trigger an async presale action that shows the waiting page (`waiting.html`) — verify it redirects on completion
- [ ] Open the device connect page in the organizer panel — verify it reloads once the device initializes
- [ ] Confirm no jQuery errors in the browser console on any of the three pages

## Summary by Sourcery

Replace jQuery-based polling on pending and device pages with native Fetch-based implementations and remove unnecessary jQuery includes from related templates.

Bug Fixes:
- Add explicit error handling for polling requests on pending and device pages to avoid silent failures.

Enhancements:
- Migrate three small polling scripts from $.get/$.getJSON to the Fetch API for JSON/text responses.
- Initialize device polling using DOMContentLoaded instead of jQuery DOM ready helpers.
- Simplify pending and waiting templates by loading only the specific polling scripts without the global jQuery bundle.


OsauravO:
@Saksham-Sirohi 
These changes refactor background  polling scripts (`reloadpending.js, ajaxpending.js, devices.js`) from jQuery `$.get/$.getJSON `to native fetch(). They have no visible UI,   they poll a URL in the background and trigger a page reload or redirect when the server responds.

Sak1012:
Thank you for the effort. Closing this PR because the current `fetch()` migration does not fully preserve the previous behavior. In particular, `fetch()` does not reject on HTTP error status codes, so the new code should explicitly check `response.ok` before calling `.json()` / `.text()`. Without that, non-2xx responses may be handled differently from the previous jQuery-based implementation. Also, removing the `{% compress js %}` wrapper from the affected templates changes asset-processing behavior and would need separate justification.

OsauravO:
@Sak1012 Thanks for the review, both points make sense but I thnik are easy to fix.

For the `fetch() `part, you  are right, it won’t reject on HTTP errors like jQuery. I will add a `(!response.ok) throw new Error(response.status`) check before parsing so non-2xx responses go through the error flow as expected.

For `{% compress js %}`, I removed it along with jQuery, but I should’ve kept it around the remaining `<script>` to preserve the asset pipeline. I will restore that.

I can push these changes and create a fresh PR.

PR #2764, opened 2026-03-14
fix: narrow broad Exception in banktransfer CSV hint import

MEGHANA660:
## Problem
`process_csv_hint()` in `banktransfer/views.py` had a broad 
`except Exception` with a TODO to narrow it down.

## Fix
Replaced with `except (ValueError, TypeError)` which covers 
the realistic failure cases when saving CSV hint settings.
Also removed the redundant `pass` statement.

## Changes
- `app/eventyay/plugins/banktransfer/views.py`: Narrowed 
  broad exception in process_csv_hint()

## Summary by Sourcery

Narrow exception handling for CSV hint processing in the bank transfer plugin to handle only expected errors when saving settings.

Bug Fixes:
- Restrict CSV hint settings import error handling to ValueError and TypeError instead of catching all exceptions.

Enhancements:
- Remove redundant code in CSV hint processing after tightening exception handling.

### DIFF
--- a/app/eventyay/plugins/banktransfer/views.py
+++ b/app/eventyay/plugins/banktransfer/views.py
@@ -500,7 +500,7 @@ class ImportView(ListView):
         try:
             o.settings.set('banktransfer_csvhint', hint)
         except (ValueError, TypeError) as e:
-            logger.error('Import using stored hint failed: ' + str(e))
+            logger.exception('Import using stored hint failed: ' + str(e))
         else:
             parsed, __ = csvimport.parse(data, hint)
             return self.start_processing(parsed)


Rachit7168:
Link the issue this PR solve in the description

mariobehling:
Thanks for your contribution and for looking into this cleanup.

We are closing this PR for now because there is no linked issue describing a concrete bug, use case, or expected behavior.

I reviewed the actual diff. This PR only changes the exception handling in `process_csv_hint()` in `app/eventyay/plugins/banktransfer/views.py` by replacing a broad `except Exception` with `except (ValueError, TypeError)` and switching the logging call to `logger.exception(...)`.

While narrowing broad exception handling can be a good improvement in principle, in this case there is no issue documenting:

* what concrete failure is being solved
* whether `ValueError` and `TypeError` are really the only expected exceptions here
* what behavior should happen if other import related failures occur

Without that context, this becomes a speculative cleanup rather than an issue-driven fix, and it is difficult to review safely. 

Please feel welcome to open an issue first describing the exact problem and expected outcome. After that, a new focused PR linked to the issue would be much easier to review and merge.

Thanks again for the effort and looking forward to a follow-up contribution.

PR #2773, opened 2026-03-15
feat: Add public event listing REST API endpoint

### LINKED ISSUE #2772: Bug: Public events visible on start page but inaccessible via REST API for unauthenticated users
## Problem

Public event data is visible to unauthenticated visitors on the HTML start page but is completely inaccessible via the REST API.

**Start page (works):** `GET /` returns 200 and renders public events to anyone.

**API (blocked):** `GET /api/v1/organizers/{organizer}/events/` returns 403 for unauthenticated requests:

```json
{"detail": "Authentication credentials were not provided."}
```

## Reproduction

1. Open an incognito browser window (unauthenticated)
2. Visit the start page — public events are listed normally
3. Visit `/api/v1/organizers/{organizer}/events/` — returns 403 Forbidden

## Screenshots

**Start page (200 OK, events visible):**

Image

**API endpoint (403 Forbidden, same event inaccessible):**

Image

## Root Cause

The `EventViewSet` uses `EventCRUDPermission`, which extends `EventPermission`. The permission check in [`permission.py` line 71](https://github.com/fossasia/eventyay/blob/dev/app/eventyay/api/auth/permission.py#L71) rejects all unauthenticated requests:

```python
def has_permission(self, request, view):
    if not request.user.is_authenticated and not isinstance(request.auth, (Device, TeamAPIToken)):
        return False
```

Even if the request reached `get_queryset()`, it would return an empty set:

```python
# app/eventyay/api/views/event.py line 117-118
else:
    qs = Event.objects.none()
```

Meanwhile, the presale views (`StartPageView`, `EventListMixin`) already serve this exact data to unauthenticated visitors using `Event.objects.filter(live=True, is_public=True)`.

## Impact

* No programmatic access to public event data for external consumers (mobile apps, third-party widgets, partner integrations)
* Blocks SPA migration — a Single Page Application front-end needs a JSON API to fetch and render public events
* Start page search is client-side only — it filters the ~30 events already in the DOM, so events on other pages are unsearchable

OsauravO:
## Summary

Fixes #2772 

Adds a new read-only REST API endpoint at `GET /api/v1/events/` that allowsunauthenticated users to browse public, live events. Previously, the`EventViewSet` at `/api/v1/organizers/{organizer}/events/` returned 403 for
unauthenticated requests, even though the same event data was publicly visibleon the HTML start page.

## Changes

- **`app/eventyay/api/serializers/event.py`** — Added `PublicEventSerializer`with a limited set of public-safe fields (name, slug, organizer, dates,location, currency, geo coordinates, has_subevents). Sensitive fields like plugins, sales_channels, meta_data, seating_plan, and valid_keys are excluded.

- **`app/eventyay/api/views/event.py`** — Added `PublicEventFilter` (search by name, filter by organizer/past/future) and `PublicEventViewSet` (read-only, no authentication required). The queryset filters for `live=True`, `is_public=True`, and excludes test-mode events.

- **`app/eventyay/api/urls.py`** — Registered the new viewset at `/api/v1/events/`.

## API

| Method | Endpoint | Auth | Description |
|--------|----------|------|-------------|
| GET | `/api/v1/events/` | None | List public events (paginated) |
| GET | `/api/v1/events/{slug}/` | None | Single event detail |

**Query parameters:** `?q=` (name search), `?organizer=` (filter by slug),
`?is_future=true/false`, `?is_past=true/false`, `?has_subevents=true/false`,
`?ordering=date_from` / `-date_from` / `name` / `slug`

**Example response:**

```
{
  "count": 1,
  "next": null,
  "previous": null,
  "results": [
    {
      "name": {"en": "Test Conference 2026"},
      "slug": "testevent",
      "organizer": "testorg",
      "date_from": "2026-04-05T15:42:36.239375Z",
      "date_to": "2026-04-07T15:42:36.239382Z",
      "location": null,
      "currency": "USD",
      "has_subevents": false,
      "geo_lat": null,
      "geo_lon": null
    }
  ]
}
```

## Summary by Sourcery

Add a public, read-only REST endpoint to list and retrieve live public events by slug.

New Features:
- Expose a `/api/v1/events/` endpoint that lets unauthenticated users list and view public, live events by slug using a restricted public serializer.

Enhancements:
- Introduce a dedicated `PublicEventSerializer` and filter set to expose only non-sensitive event fields with basic search, filtering, and ordering options via the new endpoint.


Sak1012:
Thanks for the effort. Closing this PR because we do not want to introduce unauthenticated REST endpoints for event data. It is preferred to keep API access under a consistent authenticated model instead of exposing a parallel public API, even for fields that are already visible in the frontend. Public page visibility and public API access are not the same product decision, and the latter carries a larger longterm maintenance and compatibility burden.

OsauravO:
@Sak1012 

Thanks for review, understood, that makes sense. I will keep this in mind for future contributions.

Issue #2808, opened 2026-03-15
Dashboard should display only teams with access to the event, not all organizer teams


mariobehling:
On the **Events and Tickets dashboard**, the **Event organizer section currently lists all teams belonging to the organizer**, even if those teams do not have access or permissions for the specific event.

This creates confusion and exposes irrelevant information to users managing the event.

In the example shown, multiple teams belonging to the organizer are displayed even though not all of them have actual access rights to the event.

<img width="2454" height="868" alt="Image" src="https://github.com/user-attachments/assets/373f8b69-0a6e-463a-a64f-65794dc06d13" />

### Current behavior

* The dashboard shows **all teams associated with the organizer entity**.
* Teams appear in the **Event organizer section** regardless of whether they have access to the event.
* This makes it unclear which teams actually manage the event.

### Expected behavior

* The dashboard should display **only teams that have explicit access or permissions to the current event**.
* Teams that belong to the organizer but are **not assigned to the event** should **not appear**.

### Required changes

* Filter teams shown in the **Event organizer section** based on event-level permissions.
* Only display teams that are:

  * Assigned to the event, or
  * Granted permissions within the event.

### Acceptance criteria

* The dashboard shows only teams that have permissions for the current event.
* Teams belonging to the organizer but not assigned to the event are not displayed.
* Event managers can clearly see which teams are responsible for the event.
* No regression in team permission management or organizer display logic.

v-ju:
Hello @mariobehling , Please check PR #2815

Saksham-Sirohi:
closing as the issue has been resolved

PR #2815, opened 2026-03-16
fix(dashboard): show only event teams on events and organizer page

v-ju:
Addresses #2808

Expected Outcome:
Dashboard must display only teams that have explicit access/permissions to the current event.
Teams belonging to the organizer, but not assigned to the event must not appear.
Fix:

https://github.com/user-attachments/assets/93bc0c51-eed9-466d-a894-205028a564fb

In /views/dashboards.py file , getContextData calls request.event.teams.values_list('id', 'name').
This returns only teams that have access to this event.

Steps to recreate:

1.The bug is caused by calling "request.organizer.teams.values_list('id', 'name')", which returns all teams of the organization.
complete path
image

2.This returned value is looped over the Event Orgainzer section's template.
image

## Summary by Sourcery

Bug Fixes:
- Ensure the dashboard only displays teams that have explicit access to the current event instead of all organizer teams.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -116,7 +116,7 @@ After running ``uv sync``, activate a virtual environment
 
   python manage.py migrate
 
-9. **Create a admin user account** (for accessing the admin panel):
+9. **Create an admin user account** (for accessing the admin panel):
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


Saksham-Sirohi (approved):
LGTM!

mariobehling:
The fix looks correct and aligns well with the expected behavior from #2808. Using `request.event.teams` instead of `request.organizer.teams` is the right approach to ensure only event scoped teams are shown.

PR #2885, opened 2026-03-21
Fix horizontal scrollbar in order form for long questions

### LINKED ISSUE #2882: Fix horizontal scrollbar in order form page for long questions on smaller screens
On the **order form page** of the ticket component, long question texts can cause the right-side content area to overflow horizontally. This results in a horizontal scrollbar, especially on smaller screens such as laptops.

Image

## Steps to Reproduce

* Open the ticket order form page
* Add or view a form with very long question text
* Use a smaller screen size (e.g. laptop resolution)

## Current Behavior

* The content area overflows horizontally
* A horizontal scrollbar appears

## Expected Behavior

* No horizontal scrollbar should appear
* Long text should wrap properly within the available width
* Layout should remain responsive across screen sizes

## Notes

* Likely requires CSS adjustments such as text wrapping or container width constraints
* Ensure consistent behavior across different browsers and screen sizes

Anurag-1401:
Fixes #2882

## Problem
Long question texts caused horizontal overflow and scrollbar on smaller screens.

## Solution
- Added proper text wrapping using:
  - `word-break: break-word`
  - `overflow-wrap: anywhere`
- Prevented container overflow

## Result
- No horizontal scrollbar
- Fully responsive layout
- Long text wraps correctly

## Testing
- Tested on laptop screen sizes
- Verified long text rendering

## Summary by Sourcery

Improve form layout to prevent horizontal overflow on long question text and keep the order form responsive on small screens.

Bug Fixes:
- Fix horizontal scrollbar caused by long question text in order form tables.

Enhancements:
- Ensure table cells, labels, and form groups wrap long text within available width to maintain responsive layout.

### DIFF
--- a/app/eventyay/static/pretixcontrol/scss/_forms.scss
+++ b/app/eventyay/static/pretixcontrol/scss/_forms.scss
@@ -728,3 +728,22 @@ table td > .checkbox input[type="checkbox"] {
     gap: 10px;
   }
 }
+
+
+.question-option-row,
+.table td,
+.table th {
+    white-space: normal !important;
+    word-break: break-word;
+    overflow-wrap: anywhere;
+}
+
+.table-responsive {
+    overflow-x: hidden;
+}
+
+label,
+.form-group {
+    max-width: 100%;
+    word-break: break-word;
+}


mariobehling:
Thanks for working on this. The issue itself is valid, but this PR is not in a state where we can merge it.

The biggest problem is that the change is too broad for such a specific UI issue. In the diff, you are applying wrapping rules to `.table td`, `.table th`, `label`, and `.form-group` globally in `_forms.scss`, which means this does not only affect the order form. It can change tables and form layouts across other parts of the application as well. For a bug that is limited to a specific order form view, the fix needs to be scoped much more narrowly.

There is also a concrete code risk in this line:

`.table-responsive { overflow-x: hidden; }`

That does remove the scrollbar, but it can also hide real overflow and make content inaccessible on small screens instead of fixing the underlying layout problem. In other words, this may visually suppress the symptom while introducing a new usability problem. The existing review comment on the PR already points this out correctly.

A second code concern is that the wrapping rules are quite aggressive. Combining `word-break: break-word` with `overflow-wrap: anywhere` can lead to poor readability and broken text layout, especially when applied globally. This should be tested carefully and used only in the smallest possible scope.

The review material is also incomplete. There is no screencast, no before and after comparison, no screenshots from a smaller laptop viewport, and no evidence that this was tested in a running environment. The PR only says “Tested on laptop screen sizes” and “Verified long text rendering,” which is too vague for a frontend bugfix that changes shared SCSS.

More generally, this is again the kind of PR that appears to have been opened before the work was fully validated. AI assistance is fine, but it does not replace understanding the side effects of a CSS change, testing it in context, and showing reviewers exactly what was verified.

Because of that, we are closing this PR as well.

Please open a new PR once you have a properly scoped fix and the required review material. The new PR should include a screencast, exact test steps, viewport details, and proof that the solution fixes the order form without changing unrelated tables or forms elsewhere in the app.

PR #2893, opened 2026-03-21
fix(ui): prevent text overflow in change history panel

### LINKED ISSUE #2872: Email text overflows in Organizer "Change history" panel UI
The email address in the "Change history" panel inside Organizer settings does not fit properly within the container.

Long text (e.g., admin@example.com) overflows the UI box, breaking layout and reducing readability.

## Steps to Reproduce
1. Go to Organizer settings
2. Scroll to "Change history" panel
3. Observe the displayed email

## Actual Behavior
Email text overflows outside the container and is not properly wrapped.

## Expected Behavior
Email text should wrap properly within the container without breaking layout.

## Environment
- OS: Windows (WSL)
- Browser: Chrome
- Mode: Development (Docker)

## Suggested Fix (optional)
Add CSS like:

word-break: break-word;
overflow-wrap: anywhere;

The photos:-

Image
Image

Image
Image

dsk-dev-ai:
## Problem
Email text in the "Change history" panel overflows outside the container, breaking the layout and reducing readability.

## Solution
Moved overflow handling to the existing `.logentry` CSS class by adding:

CSS: overflow-wrap: anywhere;

This ensures long text (e.g., email addresses) wraps properly within the container.

## Changes
- Removed inline styles from template
- Added overflow handling in `base.css` under `.logentry`

## Result
- Text now wraps correctly
- UI layout remains intact
- Cleaner and reusable implementation

Fixes #2872

Improved Check Screenshots:
solve 4
solve 3
solve 2
solve 1

### DIFF
--- a/app/eventyay/control/templates/pretixcontrol/includes/logs.html
+++ b/app/eventyay/control/templates/pretixcontrol/includes/logs.html
@@ -3,7 +3,7 @@
 
 <ul class="list-group">
     {% for log in obj.top_logentries %}
-        <li class="list-group-item logentry" style="overflow-wrap: anywhere;">
+        <li class="list-group-item logentry">
 
             <p class="meta">
                 <span class="fa fa-clock-o"></span> {{ log.datetime|date:"SHORT_DATETIME_FORMAT" }}
@@ -56,7 +56,7 @@
     {% endfor %}
 
     {% if obj.all_logentries_link and obj.top_logentries_has_more %}
-        <li class="list-group-item logentry" style="overflow-wrap: anywhere;">
+        <li class="list-group-item logentry">
             <a href="{{ obj.all_logentries_link }}">
                 {% trans "View full log" %}
             </a>
--- a/app/eventyay/static/common/css/base.css
+++ b/app/eventyay/static/common/css/base.css
@@ -1138,3 +1138,6 @@ pl-0 {
     margin: 5mm;
   }
 }
+.logentry {
+    overflow-wrap: anywhere;
+}


mariobehling:
Thanks for the contribution. We are closing this PR.

The screenshots clearly show a real UI issue with long log entries causing horizontal overflow. So the problem itself is valid.

However, the implementation in this PR is not in a mergeable state.

**Template issues**

* The changes in `logs.html` introduce unnecessary and potentially incorrect modifications:

  * duplicated tooltip attributes on the staff icon span
  * redundant changes to icon markup
  * an extra closing `</ul>` at the end of the template
* These are not related to the actual issue and indicate that the template was not carefully reviewed after editing.

**CSS approach**

* The fix applies `overflow-wrap: anywhere` to `.logentry`, which is a very broad rule.
* This may visually hide the overflow, but it does not demonstrate that the root cause was properly identified.
* Such a global rule can affect readability and layout in other contexts and should be scoped more precisely.

**Quality of the change**

* Even though the issue is real, the diff mixes the actual fix with unrelated and incorrect template changes.
* This makes the PR unsafe to merge and increases review complexity.

**Conclusion**

We are closing this PR because the implementation is not clean and not sufficiently scoped.

If you want to address this issue properly, please:

* limit changes strictly to what is needed to fix the overflow
* avoid unrelated template modifications
* ensure the template is syntactically correct
* clearly show before and after behavior (which you already started with)
* verify that the fix does not introduce side effects

A clean, minimal, and well-tested fix would be welcome.

Issue #2519, opened 2026-02-23
Restore login required popup when attempting checkout without authentication


mariobehling:
For events that require users to be logged in before purchasing tickets:

* A user can add a ticket to the order.
* When clicking “Check out” while not logged in, nothing happens.
* No popup, message, or redirect is shown.
* The user does not receive any indication why checkout is not proceeding.

Previously, a popup informed users that they must log in before completing the purchase. This popup is no longer appearing.

<img width="2490" height="996" alt="Image" src="https://github.com/user-attachments/assets/0c6b79db-16ca-4014-951a-da5ebe54de0d" />

#### Expected behavior

If an event requires authentication for ticket purchase and a non-authenticated user clicks “Check out”:

* A clear popup or modal should appear.
* The message should inform the user that login is required before checkout.
* The popup should include:

  * A direct “Log in” button.
  * Optionally a “Register” button if registration is enabled.

The system must not fail silently.

#### Proposed implementation

* Reintroduce or fix the frontend login check before initiating checkout.
* On “Check out” click:

  * Verify authentication state.
  * If not authenticated and login is required for the event:

    * Prevent checkout action.
    * Trigger login-required modal.
* Ensure this works consistently across:

  * Public event pages
  * Embedded ticket widgets (if applicable)

#### Acceptance criteria

* Non-logged-in users attempting checkout receive a clear login-required popup.
* Logged-in users can proceed normally.
* No silent failure when clicking “Check out”.
* Behavior matches previous expected UX.

#### Notes

This is a regression in user experience and may negatively impact ticket conversions because users receive no feedback when checkout does not proceed.

ArnavBallinCode:
Issues seems to be back on dev.eventyay.com but works locally.

ArnavBallinCode:
issue is still present on https://dev.eventyay.com , it was working on local before too

Saksham-Sirohi:
@ArnavBallinCode Closing as the issue is resolved; it was not working due to a CSP issue. @its403 implemented it correctly, that is why it worked

Saksham-Sirohi:
please open follow up issue and pr for anything pending

PR #2958, opened 2026-03-23
fix: checkout login popup

its403:
Fixes: #2519


its403:
Screenshot_2026-03-23_at_3 52 31_PM

ArnavBallinCode:
this pr does not solve the issue, it was already working on local, maybe some csp issues causing this issue

PR #2963, opened 2026-03-23
ci: simplify services and align config with TOML defaults

### LINKED ISSUE #2314: Follow up: Rework CI workflow to reduce complexity while keeping deterministic behavior
PR #1710 introduced a GitHub Actions workflow for backend validation and webapp build. The workflow is functional and was merged, but review comments highlighted several structural issues that should be addressed in a follow up to keep CI maintainable and aligned with project conventions.

A key tension discussed in the PR was between:

* keeping CI **self contained and deterministic**, and
* avoiding **unnecessary duplication, verbosity, and fragile setup logic**.

This issue tracks the remaining work needed to resolve that tension cleanly.

### 1) Reduce Docker verbosity for database services

The current workflow uses Docker service containers for PostgreSQL and Redis, including explicit health checks and manual readiness loops, for example:

```yaml
options: >-
  --health-cmd="redis-cli ping"
  --health-interval=10s
  --health-timeout=5s
  --health-retries=5

options: >-
  --health-cmd="pg_isready -U eventyay -d eventyay-db"
  --health-interval=10s
  --health-timeout=5s
  --health-retries=5
```

and additional shell loops such as:

```bash
for _ in {1..30}; do
  if pg_isready -h localhost -p 5432 -U eventyay -d eventyay-db >/dev/null 2>&1; then
    echo "Postgres is up"
    break
  fi
  sleep 2
done
```

While this works, GitHub runners already provide a full Ubuntu environment, and this level of Docker orchestration adds complexity and potential flakiness.

Action items

* Rework the workflow to rely on native services available on the runner where possible.
* Remove custom health check loops unless strictly required.

Acceptance criteria

* No verbose Docker service configuration for Postgres and Redis.
* Simpler and more robust startup logic.

### 2) Align CI configuration with TOML defaults while preserving determinism

The workflow explicitly sets many EVY_ environment variables that already exist as defaults in the project’s TOML configuration, for example:

```yaml
EVY_POSTGRES_USER: eventyay
EVY_POSTGRES_PASSWORD: eventyay
EVY_POSTGRES_DB: eventyay-db
EVY_POSTGRES_HOST: localhost
EVY_POSTGRES_PORT: "5432"
EVY_REDIS_URL: redis://localhost:6379/0
EVY_SECRET_KEY: ci-test-secret-key
EVY_DEBUG: "true"
```

The intent was to make CI runs self contained and deterministic. However, duplicating defaults in the workflow increases maintenance overhead and makes future config changes harder.

Action items
Preserve determinism while reducing duplication by choosing one clear approach:

* Option A: introduce a dedicated CI TOML config file loaded by the workflow.
* Option B: generate a minimal TOML override file during CI that only sets values that must differ.
* Option C: keep a minimal set of EVY_ variables and remove anything already covered by stable defaults.

Acceptance criteria

* CI does not duplicate TOML defaults unnecessarily.
* Deterministic behavior is preserved, including explicit secrets and stable service configuration.

### 3) Clarify and document the test scope decision

The workflow narrows test execution from all tests to a subset (`tests/stable`). This was done because parts of the broader test tree currently contain legacy imports and failing suites.

This decision needs to be explicit and documented so it does not silently become permanent technical debt.

Action items

* Document why `tests/stable` is used in CI.
* Define a follow up path, for example:

  * gradually fixing and re enabling the full test tree, or
  * explicitly quarantining legacy tests with markers or a separate job.

Acceptance criteria

* The reduced test scope is documented and intentional.
* There is a tracked path toward broader coverage over time.

## Outcome

A CI workflow that remains deterministic and reliable, but is simpler, less verbose, and better aligned with the project’s TOML first configuration strategy and long term maintenance goals.

Rachit7168:
Fix : #2314


Rachit7168:
This PR is ready to be reviewed.

@mariobehling , @Saksham-Sirohi

Rachit7168:
@mariobehling , @hongquan ,
Should I add a retry and sleep mechanism as suggested by copilot ?

hongquan:
@Rachit7168 Do you use Ubuntu or any Linux distro for work? If not, I suggest you not to mess with `apt` and `systemd`.

---
The code and the last question shows that the PR author is not familiar with Linux system.

Rachit7168:
> @Rachit7168 Do you use Ubuntu or any Linux distro for work? If not, I suggest you not to mess with `apt` and `systemd`.
> 
> The code and the last question shows that the PR author is not familiar with Linux system.

Yes, I use Ubuntu at work.

Thanks for the feedback — that’s fair.

I realize now that switching to apt/systemd for service setup was the wrong architectural choice for CI, as it reduces predictability compared to service containers. I’ll revert to using Docker services and clean up the workflow accordingly.

Appreciate the guidance.

hongquan:
> apt/systemd for service setup was the wrong architectural choice for CI

It is the correct choice, and better choice than Docker.
The issue is that, you don't understand Linux, apt and systemd to work on this task.

> it reduces predictability

It is just that you cannot predict because you don't understand the system.

hongquan:
Here is the explanation why:

- When you install Postgres, Redis via `apt`, `apt` will start Postgres, Redis automatically (via systemd). If you use Ubuntu in your daily work, you should notice this and should not need the `sudo systemctl start ....`.
- `systemd` can manage the state of services better than Docker, so do don't need the `pg_isready` checking stuff. You should figure out that the AI suggestion for this stuff is non-sense.
- To create Postgres users and DB in non-Docker setup, you don't need the long SQL queries. Your use of long queries prove that you are not familiar with typical Linux (non-Docker), and you didn't bother to read README at all.

You told to use Ubuntu at work but you seem to only use Docker and only copy the tutorial for Docker, without noticing that this task is for non-Docker.

Rachit7168:
@hongquan 
Thanks for the detailed explanation — I understand the issue now.

I overcomplicated the setup by approaching it like a Docker-based workflow instead of relying on systemd behavior on Ubuntu runners. I’ll revisit the CI with that in mind and come back with a cleaner, minimal setup aligned with native service management.

Really, thanks for the guidance, will improve and make a clear PR.

PR #2967, opened 2026-03-23
Fix: correct toggle URL in orderFormToggles.js 

### LINKED ISSUE #2881: Error when changing required or optional status of fields in ticket order forms
When attempting to change a form field from **required to optional** or vice versa directly on the **ticket order forms page**, an error message is shown.

Image

## Steps to Reproduce

* Go to ticket order forms
* Select an existing form field
* Try to change its status from required to optional or optional to required

## Current Behavior

* An error message appears
* The change cannot be applied on this page

## Expected Behavior

* It should be possible to change the required or optional status directly on the ticket order forms page
* Changes should be applied without errors

## Notes

* Currently, this change is only possible on the **custom form page**
* Behavior should be consistent across both interfaces

Shubhamhandigutte:
Fixes #2881

The toggle URL in orderFormToggles.js was incorrectly constructed 
using window.location.pathname, causing the required/optional status 
update to fail. Fixed by using a simple relative URL instead.

## Summary by Sourcery

Bug Fixes:
- Fix construction of the question toggle endpoint URL in orderFormToggles.js by using a relative path instead of deriving it from window.location.pathname.

### Before
567297051-054b3e0d-c200-4553-805b-7f45ea120bb6

 ### After the changes 
Screenshot (60)


Rachit7168:
Can you confirm whether the backend expects the endpoint as /questions/<id>/toggle/?

Right now the change simplifies it to `${questionId}/toggle/`,  which might break in contexts where the base path is not already  `/questions/`.

Also, were you able to reproduce the issue locally before fixing it? 

Provide a short screenshot or recording for it

Rachit7168:
Just a small note don't include the issue number in the title only include in description and provide Screenshots or screenrecordings

mariobehling:
We are closing this PR.

This PR does not meet the minimum requirements for review.

For UI or workflow-related changes, we require clear evidence that the solution works as intended. This includes:

* screenshots showing before and after behavior
* a screencast demonstrating the full user flow
* clear test steps and explanation of what was validated

None of this has been provided here.

Without this information, reviewers would need to reproduce and verify the changes themselves, which shifts the responsibility from the contributor to the reviewer. This is not acceptable.

Please ensure that future PRs are complete, properly tested in a running environment, and include all required documentation and visual proof before submission.

We are happy to review once these requirements are met.

PR #2968, opened 2026-03-23
Fix #200: Use Event name as identifier for payments

### LINKED ISSUE #200: Use Event name as identifier for payments instead of short form
The short form defined in the event basics is currently used to identify events when paying with Stripe and other gateways. Change it in a way that is uses the event name as an identifier instead.

Screenshot from 2024-06-21 20-00-37

In the below example we can see that the short form 2024 was used instead of the event name "Wikimania".

a8019c0d-34dd-4b7f-838b-6cbbd1bc0faf

ANKIEMUNKIE:
…tead of short form

## Summary by Sourcery

Use event names instead of slugs as identifiers in bank transfer payment lookup and keep error responses consistent.

Bug Fixes:
- Resolve order lookup issues in bank transfer processing by matching payments against event names rather than event slugs.
- Ensure the generic error response is always returned when no valid bank transfer action is processed.

### DIFF
--- a/app/eventyay/plugins/banktransfer/views.py
+++ b/app/eventyay/plugins/banktransfer/views.py
@@ -227,7 +227,8 @@ class ActionView(View):
             return JsonResponse({'results': []})
 
         if '-' in u:
-            code = Q(event__name__icontains=u.split('-')[0]) & Q(code__icontains=Order.normalize_code(u.split('-')[1]))
+            parts = u.rsplit('-', 1)
+            code = Q(event__name__icontains=parts[0]) & Q(code__icontains=Order.normalize_code(parts[1]))
         else:
             code = Q(code__icontains=Order.normalize_code(u))
         qs = (


ANKIEMUNKIE:
Hi @mariobehling! I have implemented the change to use event.name as requested in #200. I also pushed an update to use rsplit('-', 1) to safely handle any event names that legitimately contain dashes.

The CI bot raised a good point about uniqueness: unlike slugs, event.name isn't strictly guaranteed to be unique in the database, which could theoretically cause a MultipleObjectsReturned error during lookup if two events share the exact same name. Since this is an edge case, should we filter this further by organizer/date, or fallback to the slug if a duplicate name is detected? Let me know how you'd like to proceed!

mariobehling:
We are closing this PR.

This PR does not address the actual issue.

The issue is about the payment identifier used in **Stripe and other gateways**. It explicitly shows that the short form is being used there and asks for the **event name** to be used instead. This PR changes only `app/eventyay/plugins/banktransfer/views.py`, so it is working in the wrong part of the system and does not solve the reported problem.

In addition, the diff changes identifier logic in a way that is questionable on its own:

* `event__slug__iexact` is changed to `event__name__iexact`
* `event__slug__icontains` is changed to `event__name__icontains`
* displayed bank transfer codes are changed from slug-based values to event name-based values

This is not related to Stripe. It alters bank transfer behavior without explaining why this is correct or safe.

There is also no proof that this was tested end to end in a running system. For payment-related changes, we require clear evidence that the full flow was tested, including checkout, payment processing, and the resulting identifiers. None of that is shown here.

We need contributors to understand the system they are modifying and to validate their changes properly. The responsibility for producing a correct, tested solution lies with the contributor and must not be shifted to reviewers. We do not want reviewer time to be spent figuring out whether a PR works or even targets the correct component.

This PR appears to be a quick, speculative change that targets the wrong area of the codebase.

Do not submit PRs like this again. 

We expect complete, validated, end-to-end tested PRs, not partial or AI-generated guesses.

PR #2994, opened 2026-03-26
fix(organisers): restore organiser deletion

Sak1012:
This PR solves the issues and allows to delete an organiser, which was previously broken.


Sak1012:
Thank you for pointing it out @Saksham-Sirohi, it's now resolved
image

mariobehling:
Ultimately we should have a 30 day period where we keep deleted organisers and events in a bin and could reinstate them if needed, e.g. because an account was taken over by a hacker.

PR #3034, opened 2026-03-29
Fix Stripe webhook billing task dispatch and idempotency

### LINKED ISSUE #3001: Bug: Stripe webhook dispatches billing task with None invoice_id, causing silent Celery failures
### Description
In `stripe_webhook_view`, when a `payment_intent.succeeded` event is received, `invoice_id` is extracted from Stripe's `payment_intent` metadata. However, there is no guard before dispatching the Celery task  if `invoice_id` is absent from the metadata (which is valid for non-billing PaymentIntents or Stripe dashboard test events), the task is dispatched with `invoice_id=None`.

### Affected Code
```

# app/eventyay/api/views/stripe.py
if event.type == 'payment_intent.succeeded':
   invoice_id = event.data.object.get('metadata', {}).get('invoice_id')
    update_billing_invoice_information.delay(invoice_id=invoice_id)  # dispatched even if None

```

### Impact

- **Silent failure:** The `update_billing_invoice_information` Celery task catches None and returns without raising an exception. Consequently, the task shows SUCCESS in monitoring even though nothing was updated, hiding billing processing errors.
- **Wasted resources:** A Celery worker is scheduled and consumed for every non-billing payment_intent.succeeded event, even those that can never result in a billing update.
- **Partial idempotency gap:** The update() call in the task does not filter by current status, meaning rapid Stripe retries can produce multiple overlapping writes, resulting in an incorrect `paid_datetime` value for audit purposes.

### Proposed Fix

- Add a None guard in the view before dispatching the task `(if invoice_id:).`
- Add a `status=STATUS_PENDING` filter in the Celery task's update() call to ensure exactly-once semantics and prevent concurrent workers from overwriting `paid_datetime.`

shivam-pawar-7217:
Fixes #3001 
## Summary

This fixes a Stripe webhook edge case in billing.

Right now, when we receive `payment_intent.succeeded`, we try to read `invoice_id` from metadata and always enqueue the billing task.  
If `invoice_id` is missing (which can happen for non-billing payment intents or dashboard test events), we were still dispatching the Celery task with `None`.

The task then returned without raising, so it looked successful even though nothing was updated.

## Why this matters

- We were creating avoidable Celery jobs for events that cannot map to a billing invoice.
- These jobs looked like successful executions, which hides real billing issues.
- Retry behavior could also overwrite paid timestamps because updates were not restricted to pending invoices.

### 1) Stripe webhook guard
In the webhook view, we now check `invoice_id` before dispatching the task.

- If `invoice_id` is missing: we skip dispatch and log a warning.
- If `invoice_id` is present: we dispatch as before.

### 2) Task validation + idempotency
In `update_billing_invoice_information`:

- Raise `ValueError` when `invoice_id` is missing (`None`/empty).
- Update only invoices in `STATUS_PENDING` to avoid duplicate updates on retries.
- Split no-update logging into two cases:
  - warning: invoice exists but already paid (idempotent retry)
  - error: invoice does not exist (data inconsistency)
- Use timezone-aware datetime for `paid_datetime`.

### 3) Tests
Added/updated tests to cover:

- no task dispatch when `invoice_id` is missing
- task dispatch when `invoice_id` is present
- missing `invoice_id` raises `ValueError`
- pending invoice is marked paid
- already-paid invoice is not updated again
- nonexistent invoice logs error and returns `None`

## Scope

This PR is only about webhook processing correctness and billing task behavior.  
It does **not** change routing/proxy setup, and is separate from webhook endpoint 404 issues.

### DIFF
--- a/app/eventyay/eventyay_common/tasks.py
+++ b/app/eventyay/eventyay_common/tasks.py
@@ -301,13 +301,20 @@ def update_billing_invoice_information(invoice_id: str | None):
         if not invoice_information_updated:
             current_status = BillingInvoice.objects.filter(id=invoice_id).values_list('status', flat=True).first()
             if current_status is not None:
-                logger.warning(
-                    'Invoice %s was not updated because current status is %s '
-                    '(expected %s for payment update). Skipping webhook.',
-                    invoice_id,
-                    current_status,
-                    BillingInvoice.STATUS_PENDING,
-                )
+                if current_status == BillingInvoice.STATUS_PAID:
+                    logger.warning(
+                        'Invoice %s was not updated because it is already marked as paid '
+                        '(idempotency guard). Skipping duplicate webhook.',
+                        invoice_id,
+                    )
+                else:
+                    logger.error(
+                        'Invoice %s was not updated because current status is %s '
+                        '(expected %s for payment update). Possible state inconsistency.',
+                        invoice_id,
+                        current_status,
+                        BillingInvoice.STATUS_PENDING,
+                    )
             else:
                 logger.error(
                     'Invoice %s does not exist in the database. '
--- a/app/tests/billing/test_stripe_webhook_task.py
+++ b/app/tests/billing/test_stripe_webhook_task.py
@@ -51,6 +51,16 @@ def paid_invoice(db, organizer, event):
     )
 
 
+@pytest.fixture
+def expired_invoice(db, organizer, event):
+    """A BillingInvoice in STATUS_EXPIRED (unexpected for payment success)."""
+    return BillingInvoice.objects.create(
+        organizer=organizer, event=event, status=BillingInvoice.STATUS_EXPIRED,
+        amount="100.00", currency="USD", ticket_fee="100.00", final_ticket_fee="100.00",
+        monthly_bill=date.today(), created_by="test", updated_by="test",
+    )
+
+
 def _make_stripe_event(invoice_id=None, event_id="evt_test_123"):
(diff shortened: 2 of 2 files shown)


mariobehling:
We are closing this PR.

**Scope and relevance**

* This PR introduces changes without a clearly defined or validated issue.
* There is no clear evidence that the problem exists in real usage or that users are affected.
* The changes appear to be speculative improvements rather than a fix for a concrete problem.

**Diff quality and approach**

* The changes in this PR are not tightly scoped and do not clearly demonstrate a well-understood root cause.
* It looks like a generic or AI-assisted modification rather than a targeted fix based on actual behavior in the system.
* There is no clear explanation of why the chosen approach is correct or how it integrates with existing logic.

**Missing validation**

* There is no proof that the platform was deployed and tested.
* No end-to-end validation is shown.
* No screenshots, screencast, or reproducible test steps are provided.
* No evidence that edge cases or regressions were considered.

**Contribution quality**

* This PR follows the same pattern we have seen in other submissions: quick, low-confidence changes without proper validation.
* It shifts the responsibility of understanding and testing the change onto reviewers, which is not acceptable.
* Reviewer time should not be spent figuring out whether a speculative change is needed or works.

**Conclusion**

We are closing this PR because:

* it does not address a clearly defined, real issue
* it lacks proper scoping and justification
* it shows no evidence of end-to-end testing
* it does not meet the minimum requirements for review

Please do not submit PRs of this kind.

If you want to contribute, focus on issues opened by the core team.

PR #3182, opened 2026-04-06
Fix general settings validation for missing name scheme fields

### LINKED ISSUE #3181: General Settings save fails because shared form requires fields not present on that page
## Title
General Settings save fails because shared form requires fields not present on that page

## Summary
Saving changes in Control → Event → General Settings can fail with a generic settings validation error, even when editing unrelated fields such as currency, dates, or basic options.

## Steps to Reproduce
1. Open Control Panel for an event.
2. Go to Event Settings → General.
3. Change any basic field (for example currency or timeline field).
4. Click Save.
5. Observe validation failure.

Image

## Actual Behavior
Form submission fails with a generic settings validation error.

## Expected Behavior
General Settings should save successfully when submitted fields are valid.

## Root Cause
The page uses a shared form class that includes required fields for attendee name scheme configuration.
Those fields are not rendered/submitted on General Settings, but were still treated as required during form validation.
As a result, bound form validation fails even for unrelated changes.

ArnavBallinCode:
Fixes #3181 

Screenshot 2026-04-07 at 1 50 36 AM

## Summary by Sourcery

Adjust event settings form to handle missing name scheme fields without invalidating the form

Bug Fixes:
- Preserve existing name scheme and title settings when the corresponding fields are omitted from submitted data on shared settings pages

Enhancements:
- Relax the required constraint for name scheme fields on settings pages that do not render those inputs


Sak1012 (approved):

app/eventyay/control/forms/event.py
```diff
@@ -659,6 +659,19 @@ class EventSettingsForm(SettingsForm):
         'og_image',
     ]
 
+    def _is_submitted(self, field_name):
```
hongquan: Please do not make function / method privater for no reason.

hongquan (changes requested):

MukundC25 (commented):
@ArnavBallinCode Thanks for looking into this error. This fix was indeed required to make pr https://github.com/fossasia/eventyay/pull/2936 work which simplifies the checkout flow because the 'cart' setting under General was also failing. I'll update that pr once this gets merged :)

mariobehling (approved):

ArnavBallinCode (commented):
This pr was reverted because order form settings were getting reset on every save

PR #3257, opened 2026-04-14
fix(zod): update zod schema

Sak1012:
This PR fixes an issue with Zod that prevent's the schedule editor from loading

image

## Summary by Sourcery

Improve schedule editor handling of speaker references and relax CSP for Cloudflare analytics.

Bug Fixes:
- Allow talks to reference speakers by nullable codes or objects and robustly resolve them for display in the schedule editor, preventing crashes when speaker data is incomplete.

Enhancements:
- Generalize Zod schemas and frontend types to support nullable/optional speaker codes while keeping speaker name display consistent.
- Guard speaker links in the schedule editor UI when no speaker code is available, falling back to plain text names.
- Add a helper to resolve session speaker codes into existing speakers, filtering out invalid references.
- Extend content security policy to permit Cloudflare Insights scripts and connections and modernize domain/port string formatting.

### DIFF
--- a/app/eventyay/base/middleware.py
+++ b/app/eventyay/base/middleware.py
@@ -269,6 +269,7 @@ class SecurityMiddleware(MiddlewareMixin):
             'default-src': ['{static}'],
             'script-src': [
                 '{static}',
+                'https://static.cloudflareinsights.com',
                 'https://checkout.stripe.com',
                 'https://js.stripe.com',
                 'http://localhost:8080',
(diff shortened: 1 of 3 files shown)


Saksham-Sirohi (approved):
Works fine on local

Saksham-Sirohi:
This should relax the policy a bit. Real test would be done on Wikimania itself, as everywhere else it was already working, even though it is old, so a lot of stuff behaves differently, that is the pain point

mariobehling (approved):

PR #3261, opened 2026-04-15
fix(time): update cfp form to show correct deadline

Sak1012:
This PR resolves:
- CFP deadlines for non-UTC events were being saved with the wrong local time.
- On save, datetime-local values were parsed under a pytz request timezone, which produced the wrong offset for zones like America/Edmonton.
- This caused values like 05:00 to come back as 06:34 after saving.
- Updated request timezone activation to use zoneinfo.ZoneInfo(...) so datetime parsing matches the event timezone correctly.
- Restricted CFP settings saves to only the fields they own, avoiding accidental overwrites of freshly saved deadline values.
- Other changes are for ruff.

## Summary by Sourcery

Fix CFP configuration handling and timezone activation to preserve correct local deadlines for events.

Bug Fixes:
- Ensure CFP settings save only their own fields to avoid overwriting freshly updated deadline values.
- Use zoneinfo-based timezone activation on requests so CFP datetimes are parsed with the correct local offset for the event timezone.

Enhancements:
- Tighten CfP model save calls with update_fields for more precise persistence of settings and fields.
- Apply minor style and formatting cleanups in CFP forms.

### DIFF
--- a/app/eventyay/base/middleware.py
+++ b/app/eventyay/base/middleware.py
@@ -1,7 +1,7 @@
+import zoneinfo
 from collections import OrderedDict
 from urllib.parse import urlsplit
 
-import pytz
 from django.conf import settings
 from django.http import Http404, HttpRequest, HttpResponse
 from django.middleware.common import CommonMiddleware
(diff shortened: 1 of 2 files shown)


Sak1012:
https://github.com/user-attachments/assets/ca8a7000-fae8-4b29-bef6-2c40b1d1f78d

mariobehling (approved):

Saksham-Sirohi (approved):
LGTM

PR #3289, opened 2026-04-18
fix: guard django_pdb behind DEBUG flag in settings

### LINKED ISSUE #3288: Bug: django_pdb debug tool unconditionally loaded in INSTALLED_APPS
In app/eventyay/config/settings.py line 306, django_pdb is added to INSTALLED_APPS unconditionally, meaning it loads even in production. This is a debug tool that should only be active when DEBUG=True.
Fix: Guard it with if DEBUG similar to how debug_toolbar is handled on line 326.
app/eventyay/config/settings.py

MEGHANA660:
## What this PR does
Removes `django_pdb` from the unconditional `INSTALLED_APPS` list.

## Why
`django_pdb` is a debug tool that was being loaded in ALL environments 
including production. It should only be active when `DEBUG=True`.

The fix is consistent with how `debug_toolbar` is already guarded 
on line 326 of the same file.

## Changes
- Removed bare `'django_pdb'` from `INSTALLED_APPS` (line 306)
- The existing `if DEBUG` guard on line 326 already handles it correctly

Closes #3288

## Summary by Sourcery

Guard debug-only tooling and tighten error handling in bank transfer CSV hint processing.

Bug Fixes:
- Load the django_pdb debug tool only when DEBUG is enabled and the package is available.
- Restrict bank transfer CSV hint setting errors to expected value/type issues and log them with full exception context.

Enhancements:
- Align django_pdb configuration with other optional debug libraries using conditional INSTALLED_APPS registration.

### DIFF
--- a/app/eventyay/plugins/banktransfer/views.py
+++ b/app/eventyay/plugins/banktransfer/views.py
@@ -499,9 +499,8 @@ class ImportView(ListView):
         o = getattr(self.request, 'event', self.request.organizer)
         try:
             o.settings.set('banktransfer_csvhint', hint)
-        except Exception as e:  # TODO: narrow down
+        except (ValueError, TypeError) as e:
             logger.error('Import using stored hint failed: ' + str(e))
-            pass
         else:
             parsed, __ = csvimport.parse(data, hint)
             return self.start_processing(parsed)
--- a/app/eventyay/plugins/banktransfer/views.py
+++ b/app/eventyay/plugins/banktransfer/views.py
@@ -500,7 +500,7 @@ class ImportView(ListView):
         try:
             o.settings.set('banktransfer_csvhint', hint)
         except (ValueError, TypeError) as e:
-            logger.error('Import using stored hint failed: ' + str(e))
+            logger.exception('Import using stored hint failed: ' + str(e))
         else:
             parsed, __ = csvimport.parse(data, hint)
             return self.start_processing(parsed)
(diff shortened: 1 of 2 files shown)


app/eventyay/plugins/banktransfer/views.py
```diff
@@ -499,9 +499,8 @@ def process_csv_hint(self):
         o = getattr(self.request, 'event', self.request.organizer)
         try:
             o.settings.set('banktransfer_csvhint', hint)
-        except Exception as e:  # TODO: narrow down
-            logger.error('Import using stored hint failed: ' + str(e))
-            pass
+        except (ValueError, TypeError) as e:
+            logger.exception('Import using stored hint failed: ' + str(e))
```
SxxAq, line 503: Thanks for the contribution! The django_pdb change looks good and aligns with how other debug-only tools are conditionally loaded.

One small suggestion: the banktransfer exception-handling changes seem unrelated to the PR scope. It may be cleaner to move them into a separate PR.

SxxAq (changes requested):

app/eventyay/plugins/banktransfer/views.py
ViRUS-0-0: Agree with this suggestion. @MEGHANA660 Please do the needful.

Sak1012:
Thank you for the effort, closing this PR as it's been inactive for a while.

PR #3365, opened 2026-04-24
removed outline when tab focuses

### LINKED ISSUE #3361: Unexpected black focus border in tickets component on input/link fields after recent UI change
A recent change in the tickets component introduced a visual regression where input or link fields display a black border when clicked or focused.

This behavior is inconsistent with the existing design system and creates a distracting UI experience.

Image

**Steps to Reproduce:**

* Open any form or page in the tickets component with input or link fields
* Click into a field (e.g. text input or link field)

**Actual Behavior:**

* A black border appears around the focused field
* Styling differs from previous UI behavior

**Expected Behavior:**

* Focus state should follow the design system (e.g. subtle highlight, theme color)
* No harsh or inconsistent black border
* Consistent focus styling across all input fields

**Additional Notes:**

* Likely caused by:

  * default browser focus styles overriding custom styles
  * missing or overridden CSS rules (e.g. `outline`, `box-shadow`)
* Should ensure accessibility is preserved while fixing styling
* Verify across browsers

**Acceptance Criteria:**

* Focus styling is consistent with design system
* No unexpected black borders on focus
* Accessibility (focus visibility) remains intact
* Verified across browsers and form components

v-ju:
Fixes #3361

removes outline of 5px which causes border visibility when tab is focused.

https://github.com/user-attachments/assets/85e2b5aa-c2f5-4d9e-9cb8-b004fbe3a9f0

## Summary by Sourcery

Bug Fixes:
- Remove the 5px WebKit focus outline that caused an unintended visible border on tab-focused elements.


Rachit7168:
Rather than commenting out, remove it if it's not needed. And address Comments

app/eventyay/static/bootstrap/scss/bootstrap/mixins/_tab-focus.scss
```diff
@@ -4,6 +4,6 @@
   // WebKit-specific. Other browsers will keep their default outline style.
   // (Initially tried to also force default via `outline: initial`,
   // but that seems to erroneously remove the outline in Firefox altogether.)
-  outline: 5px auto -webkit-focus-ring-color;
+  // outline: 5px auto -webkit-focus-ring-color;
   outline-offset: -2px;
```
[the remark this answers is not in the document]
v-ju: tab-focus should not have any outline ring, and this behavior aligns to it.

Aqil-Ahmad:
@v-ju  i'm not sure, but i think overriding the mixin is better than modifying the original mixin file, as it will be overwritten again  after bootstrap update

ArnavBallinCode (changes requested):
This change seems to remove the focus outline (focus state) only in this specific mixin, but the issue is present across multiple pages (e.g., presale and others).
Could you please address this globally. 


Image

Image

Image

Image

ArnavBallinCode (changes requested):
Since this issue is global, it would be better to define a centralised focus style in the app-level SCSS  rather than patching individual components, there still exist some places with the same

Aqil-Ahmad:
@v-ju the borders still appear on the login page and public event page also the navigation buttons. please merge this https://github.com/v-ju/eventyay/pull/1 and test the changes then i believe there is nothing more.

image



image

v-ju:
> @v-ju the borders still appear on the login page and public event page also the navigation buttons. please merge this [v-ju#1](https://github.com/v-ju/eventyay/pull/1) and test the changes then i believe there is nothing more.
> 
> image image

Hi @Aqil-Ahmad , as Arnav suggested,  the changes needs to be applied globally, creating a separate css file targetting global elements and its cascaded children would be better approach right ? Please confirm on this.

Aqil-Ahmad:
> Hi @Aqil-Ahmad , as Arnav suggested, the changes needs to be applied globally, creating a separate css file targetting global elements and its cascaded children would be better approach right ? Please confirm on this.

i am not sure about that because it will again be needed to be used in all places

v-ju:
> > Hi @Aqil-Ahmad , as Arnav suggested, the changes needs to be applied globally, creating a separate css file targetting global elements and its cascaded children would be better approach right ? Please confirm on this.
> 
> i am not sure about that because it will again be needed to be used in all places

So for now, should i target individual elements itself ? @Aqil-Ahmad

Aqil-Ahmad:
> So for now, should i target individual elements itself ? @Aqil-Ahmad

i tested locally and i think the the _theme file and the custom css file are the only two where we need the overrides, so if only two files then there is no need to create separate css file and import it there. but lets hear from @ArnavBallinCode before any further changes

ArnavBallinCode (changes requested):
https://github.com/user-attachments/assets/a08cb3ca-765b-489f-8891-c5be65753194

Instead of adding rules to 4 different files, put a single global override in `app/eventyay/static/pretixbase/scss/_theme.scss` (which is already imported by both presale and control). This avoids duplication and ensures all pages are covered, it is not fixed here

Aqil-Ahmad:
> Instead of adding rules to 4 different files, put a single global override in `app/eventyay/static/pretixbase/scss/_theme.scss` (which is already imported by both presale and control). This avoids duplication and ensures all pages are covered, it is not fixed here

@ArnavBallinCode please look at https://github.com/v-ju/eventyay/pull/1 . only overriding in _theme works but not for the event page, look at the screenshots:

image

image
 
so we need to apply overrides in two files _theme.scss and custom.css

Aqil-Ahmad:
@v-ju you can merge my pr and then, i think, this is ready to be merged.

ArnavBallinCode (approved):

Aqil-Ahmad (approved):

Rachit7168 (approved):
LGTM. I viewed and tested your changes, and they look great.

Rachit7168 (changes requested):
Maybe try not to use or avoid the use of !important?

app/eventyay/static/pretixbase/scss/_theme.scss
```diff
@@ -158,6 +158,13 @@
   background: $state-info-border;
   content: "\f0e3";
 }
+
+:where(a, button, summary, [role="button"], .btn, .header-nav):focus,
+:where(a, button, summary, [role="button"], .btn, .header-nav):focus-visible,
+:where(.btn, .header-nav).focus {
+  outline: none !important;
```
Saksham-Sirohi: please avoid use of !important if required you can create a new class also this CSS logic may look complex to beginners consider making it more easy to understand

Saksham-Sirohi (changes requested):

app/eventyay/static/pretixbase/scss/_theme.scss
Saksham-Sirohi: the issue still seem to be prevalent in the talk area

v-ju:
@Aqil-Ahmad , to keep accessibility support, do you think we should make outline none only for focus on the element, not focus-visible too ?

app/eventyay/static/pretixbase/scss/_theme.scss
Aqil-Ahmad: > please avoid use of !important if required you can create a new class also this CSS logic may look complex to beginners consider making it more easy to understand

it won't work without the !important

Aqil-Ahmad: > the issue still seem to be prevalent in the talk area

i think the issue was only in the tickets/common dashboard, talks still looks good @v-ju could you please verify

Aqil-Ahmad:
> @Aqil-Ahmad , to keep accessibility support, do you think we should make outline none only for focus on the element, not focus-visible too ?

yes i think youre right override on focus-visible should be removed @Saksham-Sirohi please confirm

app/eventyay/static/pretixbase/scss/_theme.scss
Saksham-Sirohi: The !important parameter should not be used if you think this is explicitly required maybe change order of writing use new classes or ids there are a lot of ways to doing this

ArnavBallinCode: We already use this in a couple of places, I think it is fine in this case?

Saksham-Sirohi (changes requested):
@v-ju
Yes, I agree. We should avoid overriding `:focus-visible`.

It exists specifically for keyboard accessibility, and removing it globally would make navigation harder for keyboard users and screen reader assisted workflows.

I think the safer approach is:

* keep the visual cleanup only for `:focus`
* preserve `:focus-visible`
* scope overrides only where the unwanted browser outline appears

Regarding `!important`, my concern is mainly long-term maintainability and CSS specificity conflicts. If the existing styling architecture already relies on `!important` in a few shared overrides and there is no cleaner specificity/order-based solution here, or classes require greater change, then using it in this isolated case is acceptable.

After removing the `focus-visible` override, the accessibility concern should also be resolved.

Also, I highly doubt that all these SCSS files are actually used. Please try removing changes from these to see whether they are in active use

CC: @Aqil-Ahmad @ArnavBallinCode @Rachit7168

Aqil-Ahmad:
> Also, I highly doubt that all these SCSS files are actually used. Please try removing changes from these to see whether they are in active use

i think the custom.scss changes aren't needed, @v-ju please remove the focus-visible overrides and custom.scss changes (but do test without these) so it can finally be merged.

v-ju:
@Rachit7168 , Please check current changes.

I have tried to test on all pages, and havent found focus issue. Please let me know with screenshots if you find any discrepancy. I have removed changes from custom.css file as i found the file dint have any effect of the issue.

For ex. a page with id main-container did not have any focus issues withour existence of the special target on #main-container.

Screenshot 2026-05-08 at 6 28 27 PM
Screenshot 2026-05-08 at 6 29 00 PM

Rachit7168 (changes requested):
https://github.com/user-attachments/assets/c5039f43-28d8-4143-b300-8b39425676c9

The changes are looking pretty good , No issue found in talk area as well. Just there are some unnessary changes such as space left , line removed , can we just remove that and then I think it's good to go !

Aqil-Ahmad:
> I have removed changes from custom.css file as i found the file dint have any effect of the issue.

@v-ju custom.css changes are needed. please restore those and remove unrelated changes. try flushing your cache, collect static files and compress, and then you will able to see these issues

image

image

image

v-ju:
Closing as reworked on #3501

PR #3389, opened 2026-04-25
Fix Wikimedia OAuth login for existing email accounts

### LINKED ISSUE #3351: OAuth (Wikimedia) login fails for existing email account; redirects to password login instead
## Title
OAuth (Wikimedia) login fails for existing email account; redirects to password login instead

---

## Describe the bug
Previously, users could log in using Wikimedia OAuth even if an account already existed with the same email (used for email/password login).

Now, when attempting to log in via Wikimedia (with the same email as an existing account), the system does not authenticate via OAuth. Instead, it sends an email stating that an account already exists and instructs the user to log in using email/password.

This creates a broken authentication flow where:
- Same email exists across both login methods
- OAuth login is not linked or recognized
- User is blocked from using Wikimedia login

---

## Steps to reproduce
1. Create or have an existing account using email/password
2. Ensure the same email is associated with a Wikimedia account
3. Attempt to log in via Wikimedia OAuth
4. Observe the response/email

---

## Expected behavior
- Wikimedia OAuth login should authenticate successfully if the email matches an existing account  
- OR the system should automatically link the Wikimedia account to the existing user  
- Users should be able to log in via either method seamlessly (email/password OR Wikimedia)

---

## Actual behavior
- OAuth login fails
- User receives an email stating "Account Already Exists"
- System forces login via email/password instead of allowing Wikimedia login

---

## Screenshots

Image

ArnavBallinCode:
Fixes #3351

- Ensure existing users found via MediaWiki social login always have a verified allauth EmailAddress, so mandatory email verification does not redirect them to /accounts/confirm-email/
- Keep email-auth linking for MediaWiki even when provider app settings disable email_authentication, matching project-level settings
- Add a MediaWiki social signup fallback form to log in and auto-connect existing users instead of sending account-already-exists flow

https://github.com/user-attachments/assets/4968b9b2-8bbb-4df6-8634-2f8a0d5157c6

## Summary by Sourcery

Improve MediaWiki OAuth handling to correctly link and verify existing user accounts based on confirmed email information.

Bug Fixes:
- Ensure MediaWiki OAuth logins for existing users create or update a verified allauth EmailAddress to avoid unnecessary email confirmation flows.
- Align MediaWiki email-based authentication with global EMAIL_AUTHENTICATION settings even when the provider app disables email_authentication.
- Prevent MediaWiki OAuth from linking accounts when the social profile email is unconfirmed.

Enhancements:
- Add adapter logic to resolve a verified MediaWiki email from multiple sources and to fetch it from the MediaWiki profile API when missing.
- Introduce helper utilities to normalize and verify user email addresses during social login.

Tests:
- Add MediaWiki OAuth integration tests covering existing-user linking, profile-email fallback, case-insensitive email verification, and protection against unconfirmed email linking.


ArnavBallinCode:
on further testing this issue on dev.eventyay.com, after logging through mediawiki, it sends a confirmation email and once confirmed, it lets the user sign in from the next time onwards with the same account, need to look into this in more detail. Although the same behaviour is not present on the local testing

ArnavBallinCode:
On dev / prod once logged in through the the wikimedia using the above method, cannot login through username and password again

hongquan:
Seems to be in wrong redirection: we should delete custom Wikimedia login flow, instead of adding a lot of custom code to the account adapter.

ArnavBallinCode:
Closing in favour of #3444

PR #3397, opened 2026-04-26
fix: log.py duplicate import, bare except, and user.name bugs

Manishnemade12:
### Description

- Remove duplicate import of ScopedManager
- Change LogEntry.__str__ user.name to user.email
- Change ActivityLog.__str__ person.name to person.email
- Replace bare except in display_object with except Exception

## Summary by Sourcery

Improve logging models’ string representations and exception handling for safer debugging output.

Bug Fixes:
- Use user email instead of name in LogEntry string representation to avoid relying on a potentially missing name field.
- Use person email instead of name in ActivityLog string representation to avoid relying on a potentially missing name field.
- Replace a bare except in display_object with an explicit Exception handler to prevent unintended exception swallowing.

### DIFF
--- a/app/eventyay/base/models/log.py
+++ b/app/eventyay/base/models/log.py
@@ -153,7 +153,7 @@ class LogEntry(models.Model):
                 return ''
 
             co = self.content_object
-        except Exception:
+        except Exception:Exception
             return ''
         a_map = None
         a_text = None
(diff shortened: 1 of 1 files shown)


Saksham-Sirohi (changes requested):
Please provide screenshots of the change in action and handle CI failures. One thing I see is wrong indentation. I think you are relying too heavily on LLMs, so much so that you are not even able to recognise such small problems. Failure to comply will lead to the closure of the pull request

Issue #3435, opened 2026-04-28
Import legacy Wikimedia events into new Eventyay talk platform


mariobehling:
Several Wikimedia events currently exist on legacy Eventyay instances and need to be imported into the new talk platform.

The goal is to migrate these events to ensure consistency, maintainability, and continued usage on the updated platform.

**Events to Import:**

* WikiIndaba 2024
* Wikimania 2024

  * Source 1: [https://wikimedia-legacy.eventyay.com/talk/orga/event/wikimania2025/](https://wikimedia-legacy.eventyay.com/talk/orga/event/wikimania2025/)

* Wikimania 2025
* Source 2: [https://app.eventyay.com/talk/orga/event/](https://app.eventyay.com/talk/orga/event/)

**Scope of Import:**

The following data should be migrated where applicable:

* Event configuration
* Sessions / proposals
* Speakers
* Tracks and session types
* Review data (if compatible)
* Schedule

**Expected Behavior:**

* Events are fully available in the new platform
* Data integrity is preserved
* Relationships between sessions, speakers, and tracks remain intact
* No duplicate or missing records

**Additional Notes:**

* Legacy data structures may differ and require mapping
* Some components (e.g. reviews) may need transformation or partial migration
* URLs and identifiers should be adjusted to the new platform
* Validation needed after import

**Acceptance Criteria:**

* All listed events are available in the new talk platform
* Sessions, speakers, and tracks are correctly imported
* Schedule (if available) is correctly reflected
* No critical data loss
* Organisers can access and manage imported events
* Verified against legacy data for consistency

PR #3437, opened 2026-04-29
add FILE_UPLOAD_MAX_MEMORY_SIZE

ArnavBallinCode:
part of #3435 

Django side was mostly fine, settings.py already defines [upload_size_csv = 10] (10MB) and [FILE_UPLOAD_DEFAULT_LIMIT = 10MB]. 

However, Django's [DATA_UPLOAD_MAX_MEMORY_SIZE]  was not set, defaulting to 2.5MB, which would also block larger uploads.

post this pr, the increased limit should work once the nginx is reloaded  

https://docs.djangoproject.com/en/6.0/ref/settings/#file-upload-max-memory-size

## Summary by Sourcery

Set a server-side memory limit for handling incoming file uploads.

New Features:
- Configure DATA_UPLOAD_MAX_MEMORY_SIZE to cap in-memory data for uploads in the Django settings.

Deployment:
- Touch nginx deployment config to align with the updated upload size handling.


Sak1012:
Nginx config updated on dev server.

PR #3444, opened 2026-04-30
Use Django-allauth native social login flow 

### LINKED ISSUE #3148: Implement strict email identity model and SSO linking on top of django-allauth integration
The recent PR https://github.com/fossasia/eventyay/pull/2675 for the migration to django-allauth introduces improved authentication and multi-email support. The enforcement of global email uniqueness is handled as part of that work.

This follow up issue focuses on completing the identity model by addressing missing SSO linking behavior, account management visibility, and consistency across authentication flows.

**Current Gaps:**

1. **SSO linking not implemented**

   * No guaranteed behavior when an SSO login email matches an existing account
   * Risk of creating duplicate or disconnected identities across login methods

2. **No “Linked services” UI**

   * Users cannot see which providers (e.g. Google, Wikimedia) are linked
   * No way to manage or extend linked login methods

3. **Unclear SSO behavior across flows**

   * No defined or consistent handling for:

     * first time SSO login
     * repeated SSO login
     * linking additional providers

4. **Registration constraints need verification**

   * Ensure previous restrictions (e.g. disabling registration) are still respected after migration

**Required Implementation**

### 1. SSO Account Linking

**Goal:**

* One user account per identity, regardless of login method

**Implementation:**

* Override or extend:

  * `SOCIALACCOUNT_ADAPTER`

Flow:

* On SSO login:

  * If email matches an existing account:

    * link provider to that account
    * do not create a new account
    * log user in

**User message:**

* “We are linking your login to your existing account. You can now log in using the SSO provider or email.”

* Configure:

  * `SOCIALACCOUNT_EMAIL_AUTHENTICATION = True`
  * implement auto-connect behavior if not sufficient out of the box

### 2. Linked Services UI

Add a new section in account settings:

* Title: “Linked services”
* Display:

  * connected providers (e.g. Google, Wikimedia)

**Features:**

* Show all linked providers
* Allow linking additional providers (if enabled)
* (Optional) allow unlinking where safe

### 3. Consistent SSO Flow Definition

Define and implement consistent behavior for:

* First time SSO login:

  * create account if no match exists

* Existing email match:

  * link and log in

* Additional provider linking:

  * attach provider to existing logged-in account

Ensure no ambiguous or duplicate account creation paths remain.

### 4. Preserve Registration Controls

* Ensure prior constraints (e.g. registration enable or disable flags) still apply
* Reintroduce via adapter or wrapper if lost during migration

**Acceptance Criteria:**

* SSO login links to existing accounts when email matches
* No duplicate accounts created via SSO
* Clear user message shown on account linking
* “Linked services” section is visible and functional
* Users can see and manage connected providers
* Registration constraints remain intact
* End to end tested for:

  * email login
  * SSO login
  * linking additional providers

**Notes:**

* Builds on django-allauth integration
* Focused on identity consistency and user transparency
* Should be implemented in incremental steps (SSO linking first, UI second)

### LINKED ISSUE #3351: OAuth (Wikimedia) login fails for existing email account; redirects to password login instead

ArnavBallinCode:
## Problem

Users created with username/password could not log in via WikiMedia OAuth even when their WikiMedia account used the same email address. They were redirected to `/accounts/3rdparty/signup/` instead of being logged in.

Two root causes:

**1. `OAuthReturnView` bypassed allauth entirely**
The old view called `User.objects.get_or_create(email=...)` directly, creating bare User objects with no `SocialAccount` or `EmailAddress` records. This meant allauth's email-based auto-connect had nothing to match against on subsequent logins.

**2. `MediaWikiProvider` discards email when `confirmed_email=False`**
The upstream provider's `extract_email_addresses()` returns `[]` when the Wikimedia account has `confirmed_email=False`. With no email addresses on the `SocialLogin`, allauth's `_lookup_by_email` never runs, the user is treated as new, and the signup form is shown.

**3. Email may be `null` regardless of `confirmed_email`**
When the OAuth consumer registered on meta.wikimedia.org lacks "View email address" permission, the profile API returns `email=null` even for confirmed accounts. This blocks email-based lookup entirely.

---

### Remove `OAuthReturnView`
Deleted the custom view and its URL. `OAuthLoginView` now simply redirects to allauth's standard `mediawiki_login` URL, letting allauth own the full callback flow.

### `EventyayMediaWikiProvider`
Subclass of `MediaWikiProvider` with two overrides:
- `extract_email_addresses`: when `VERIFIED_EMAIL=True` in settings, returns the email as verified even if `confirmed_email=False`.
- `extract_extra_data`: always stores the raw email in `extra_data` (upstream sets it to `null` when unconfirmed).
- `package = "allauth.socialaccount.providers.mediawiki"` — required so allauth loads the upstream URL conf instead of deriving it from `__module__`, which would cause a `NoReverseMatch`.

### Username fallback in `pre_social_login`
When `email=null` (no email permission on the OAuth consumer), the standard email lookup returns nothing. `CustomSocialAccountAdapter.pre_social_login` now tries `User.objects.get(wikimedia_username=extra_data['username'])` as a fallback. If found, `sociallogin.user` is set directly. `_did_authenticate_by_email` is intentionally left `None` to prevent allauth from calling `wipe_password()` (which would break the user's existing username/password login).

### allauth settings
```python
SOCIALACCOUNT_EMAIL_AUTHENTICATION = True
SOCIALACCOUNT_EMAIL_AUTHENTICATION_AUTO_CONNECT = True
```
These enable allauth's built-in email-based user matching.

> **Note on SCOPE:** WikiMedia's OAuth2 implementation does not accept `openid` or `profile` as valid scope values — passing them returns `invalid_scope`. The default scope (no explicit `SCOPE` key) is sufficient to receive the full profile. Whether `email` is included in the response depends solely on whether the OAuth consumer was registered with "View email address" permission on meta.wikimedia.org — not on the requested scope. The username fallback covers the case where it is not.

### `backfill_emailaddress` management command
Pre-allauth users have no `EmailAddress` row. Without it, `_lookup_by_email` cannot verify the match. This one-time command creates verified primary `EmailAddress` rows for all affected users:
```
python manage.py backfill_emailaddress [--dry-run]
```

---

## OAuth callback flow

```
Click "Login with WikiMedia"
  → OAuthLoginView → /accounts/mediawiki/login/ → meta.wikimedia.org
  → User approves → /accounts/mediawiki/login/callback/
  → complete_login()
      → sociallogin_from_response()
          → EventyayMediaWikiProvider.extract_email_addresses()
              → email present + VERIFIED_EMAIL=True → [EmailAddress(verified=True)]
      → sociallogin.lookup()
          → _lookup_by_socialaccount()  (fast path, returning users)
          → _lookup_by_email()          (matches existing user by verified email)
      → pre_social_login()
          → require_provider_enabled()
          → lookup_by_wikimedia_username()  (fallback when email=null)
      → _authenticate()
          → is_existing=True  → _login() → user logged in ✓
          → is_existing=False → process_signup() (genuinely new user)
```

---

## How to test

**Reproduce the bug (on `dev`):**
1. Check out `dev`
2. Create a user with username/password using email address `X`
3. Configure WikiMedia OAuth in admin settings
4. Log out, click "Login with WikiMedia" using a WikiMedia account that has email `X`
5. → Redirected to signup form (**bug**)

**Verify the fix (on this branch):**
1. Check out `fix/use-allauth-native-social-login`
2. Run: `python manage.py backfill_emailaddress`
3. Set the WikiMedia username on the existing user:
   ```python
   User.objects.filter(email='X').update(wikimedia_username='<wikimedia-username>')
   ```
4. Log out, click "Login with WikiMedia"
5. → Logged in as the existing user, redirected to `/common/` ✓

The username fallback (step 3) is only needed if the OAuth consumer lacks email permission. If it has email permission, step 3 can be skipped and the fix works via email matching alone.

---

## Problem

Users who registered with username/password could not log in via WikiMedia OAuth even when their WikiMedia account shared the same email address. They were redirected to the `/accounts/3rdparty/signup/` form instead of being logged in.

Two root causes:

**1. `email=null` in the WikiMedia API response**

The `/oauth2/resource/profile` endpoint only returns the `email` field when the OAuth consumer was registered with "View email address" permission AND the `email` scope is requested. Without this, the response has `confirmed_email: true` but `email: null`.

Because allauth's `_lookup_by_email()` only runs when `sociallogin.email_addresses` is non-empty, an empty list means the user is treated as new and routed to the signup form.

**2. `confirmed_email=False` support**

The upstream `MediaWikiProvider.extract_email_addresses()` discards the email entirely when `confirmed_email=False`. Many Wikimedia accounts have unconfirmed emails (pending re-confirmation after email change, etc.).

---

### Layer 1 — Request the `email` scope

Added `SCOPE: ['openid', 'profile', 'email']` to `SOCIALACCOUNT_PROVIDERS['mediawiki']`. When the OAuth consumer has email permission, this causes Wikimedia to include the email in the profile response, enabling allauth's standard email-based user lookup.

> **Operator requirement**: the OAuth consumer registered on meta.wikimedia.org must have "View user's email address" permission enabled. Users may need to re-authorise after scope change.

### Layer 2 — `EventyayMediaWikiProvider` (trusts email when `VERIFIED_EMAIL=True`)

When `VERIFIED_EMAIL=True` is set (our default), `extract_email_addresses()` returns the email as verified even if `confirmed_email=False`. This allows allauth's `_lookup_by_email` to match an existing eventyay user by email.

`package = "allauth.socialaccount.providers.mediawiki"` is set explicitly so allauth loads the upstream `mediawiki.urls` URLconf instead of deriving it from `__module__` (which would produce a `NoReverseMatch`).

### Layer 3 — Username fallback in `pre_social_login`

When both layer 1 and 2 fail (email still null after scope fix, e.g. consumer has no email permission), `CustomSocialAccountAdapter.pre_social_login` queries `User.wikimedia_username` against `extra_data['username']` from the profile response.

If a match is found, `sociallogin.user` is set directly. `_did_authenticate_by_email` is intentionally left as `None` to prevent allauth's `_accept_login` from calling `wipe_password()`, which would break the user's existing username/password login.

### Layer 4 — `backfill_emailaddress` management command

Pre-allauth users have no `EmailAddress` row in `account_emailaddress`. Without it, `_lookup_by_email` cannot verify the match even when the email is returned. The management command creates verified primary `EmailAddress` rows for all such users:

```
python manage.py backfill_emailaddress [--dry-run]
```

---

## Lookup flow (OAuth callback)

```
WikiMedia callback → complete_login()
  → sociallogin.lookup()
      → _lookup_by_socialaccount()   # fast path for returning users
      → _lookup_by_email()           # requires email in response + EmailAddress row
  → CustomSocialAccountAdapter.pre_social_login()
      → require_provider_enabled()
      → lookup_by_wikimedia_username()  # fallback when email=null
  → _authenticate()
      → is_existing=True  → _login() → perform_login() → user logged in
      → is_existing=False → process_signup() (new account)
```

fixes #3148
fixes #3351


Saksham-Sirohi (changes requested):
the changes are bit too heavy on comments make sure they are small and concise.

app/eventyay/plugins/socialauth/views.py
```diff
@@ -172,21 +100,20 @@ def __init__(self, *args, **kwargs):
...
+        normalised result back so that ``get_context_data`` and the template
+        see the complete provider list.
+        """
+        raw = self.gs.settings.get('login_providers', as_type=dict)
+        try:
+            validated = LoginProviders.model_validate(raw or {})
+        except ValidationError as e:
+            logger.error('Error while validating login providers: %s', e)
+            validated = LoginProviders()
+        self.gs.settings.set('login_providers', validated.model_dump())
```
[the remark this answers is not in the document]
hongquan: Valid review.

app/eventyay/eventyay_common/adapter.py
```diff
@@ -32,9 +60,17 @@ def post_login(
         request.session[KEY_LAST_FORCE_LOGIN] = int(time.time())
         # LoginForm only includes this field when long sessions are enabled; keep this in sync
         # with base/forms/auth.py.
-        keep_logged_in = settings.EVENTYAY_LONG_SESSIONS and bool(request.POST.get('keep_logged_in'))
+        # For social logins the preferred-provider flag is stored in the session
+        # by OAuthLoginView; fall back to the POST field for normal form login.
+        # Pop the session flag unconditionally so it is always consumed, even
+        # when EVENTYAY_LONG_SESSIONS is disabled (short-circuit would skip it).
+        social_keep = request.session.pop('socialauth_keep_logged_in', False)
```
hongquan: Please define a constant for this key.

app/eventyay/eventyay_common/adapter.py
```diff
@@ -44,6 +80,22 @@ def post_login(
             redirect_url=redirect_url,
         )
 
+        # Set JWT SSO cookie for the Talk module integration.
+        if isinstance(response, HttpResponseRedirect) and request.user.is_authenticated:
```
hongquan, line 77: The Talk module run in the same domain, the same server process, why do we need JWT?

app/eventyay/plugins/socialauth/adapter.py
```diff
@@ -34,3 +38,51 @@ def get_requests_session(self):
...
+        return user
+
+    def pre_social_login(self, request, sociallogin):
+        super().pre_social_login(request, sociallogin)
+        self._require_provider_enabled(request, sociallogin.account.provider)
+        if sociallogin.is_existing:
+            self._sync_wikimedia_username(sociallogin.user, sociallogin)
+
+    @staticmethod
+    def _require_provider_enabled(request, provider):
```
hongquan: Make it a public, standalone function.

app/eventyay/plugins/socialauth/adapter.py
```diff
@@ -34,3 +38,51 @@ def get_requests_session(self):
...
+            or not cfg.get('secret')
+        ):
+            logger.warning('Social login attempt for unavailable provider: %s', provider)
+            messages.error(request, _('This login method is not available.'))
+            raise ImmediateHttpResponse(
+                HttpResponseRedirect(reverse('eventyay_common:auth.login'))
+            )
+
+    @staticmethod
+    def _sync_wikimedia_username(user, sociallogin):
```
hongquan: Make it a public, standalone function.

hongquan (changes requested):
The backfill does not need to be a DB migration, just make it a Django command. Leave a command saying that it is one-time-use, which will be deleted after use.

app/eventyay/eventyay_common/adapter.py
ArnavBallinCode: added comment explaining the same

hongquan (approved):
Approve now, but I still need the answer about why we need JWT for Talk module, which is run from the same Django process, same domain.
In that case, the session is shared and user in Talk module should be automatically treated as "logged-in". Why we need the JWT?

Saksham-Sirohi (approved):
Looks Good to Me rest can be done in a follow up PR

Sak1012 (approved):

Issue #3451, opened 2026-05-03
Bug: Speaker Password Reset Crashes with `NoReverseMatch` (500 Error)


SxxAq:
**Describe the bug**

Clicking "Reset Password" on any speaker in the organizer panel results in a 500 error. The speaker never receives a reset email.

**Steps to Reproduce**

1. Submit a talk proposal as a speaker via the public CfP page (creates a speaker account)
2. Log in as the event organizer
3. Go to **Speakers** → click on the speaker → click **Reset Password** → confirm

**Expected behavior**

The speaker receives a password reset email with a valid link.

**Actual behavior**

```
NoReverseMatch at /orga/event/<slug>/speakers/<code>/reset
Reverse for 'event.recover' with keyword arguments
{'token': '...', 'event': '<slug>'} not found.
1 pattern(s) tried: ['(?P<organizer>[...]+)/(?P<event>[...]+)/reset/(?P<token>[^/]+)']
```

**Impact**

The "Reset Password" feature for speakers is **completely non-functional** for all events. Organizers cannot help speakers who are locked out of their accounts. 

<img width="1073" height="356" alt="Image" src="https://github.com/user-attachments/assets/b3288326-4a14-47e2-a32d-1356ad5f617e" />

PR #3452, opened 2026-05-03
Fix Speaker Password Reset Flow

SxxAq:
Fixes a `NoReverseMatch` error in the organizer panel and a related issue in the password change flow.

### Cause

* The `cfp:event.recover` route requires both `event` and `organizer` slugs, but only `event` was passed
* `change_password()` referenced a non-existent `self.name` attribute

### Fix

* Added missing `organizer` kwarg:

  ```python
  if not orga:
      kwargs['organizer'] = event.organizer.slug
  ```
* Replaced `self.name` with `self.fullname` in `change_password()`

### Testing

✅ Password reset email sent after organizer clicks "Reset Password"
✅ Reset link resolves correctly (no `NoReverseMatch`)
✅ Password change completes without error
✅ Confirmation email sent after password change

Screenshot From 2026-05-03 15-48-16

---

### Summary by Sourcery

**Bug Fixes:**

* Resolve `NoReverseMatch` during speaker password reset by including the organizer slug
* Fix attribute error in password change flow

- **Closes:** #3451


Aqil-Ahmad:
@SxxAq your changes works and fixes the issue, but i think the button is buggy, it shows the message that you can't login after the reset, but the password is never reset and you can still login with same old password and  no email for reset is sent(which should happen).

SxxAq:
@Aqil-Ahmad The reset email is generated and sent correctly. The email requires Celery worker to be running (celery -A eventyay worker) for async delivery. In development, emails are written to eventyay/data/dev-sent-emails/. In production, they're sent via SMTP.

Screenshot From 2026-05-04 00-20-18
Screenshot From 2026-05-04 00-21-03

SxxAq:
After password change:

Screenshot From 2026-05-04 00-29-41
Screenshot From 2026-05-04 00-34-20

Aqil-Ahmad:
> @Aqil-Ahmad The reset email is generated and sent correctly. The email requires Celery worker to be running (celery -A eventyay worker) for async delivery. In development, emails are written to eventyay/data/dev-sent-emails/. In production, they're sent via SMTP.

ohhh!! then its nice, if an email is sent for password reset then its good, i though its just generated and queued but never sent.

Aqil-Ahmad (approved):

Saksham-Sirohi (approved):
LGTM

PR #3453, opened 2026-05-03
chore: prevent error 500 on old events due to scope issue

Saksham-Sirohi:
## Summary by Sourcery

Bug Fixes:
- Prevent 500 errors when displaying log entries for old events by bypassing scoped queries in product variation lookup and order change display.


ArnavBallinCode (approved):
LGTM, will fix https://dev.eventyay.com/orga/event/summit/ but wont be needed for new events 


attaching logs of the issue for the same since the pr does not have an issue 

[logs.txt](https://github.com/user-attachments/files/27337231/logs.txt)

PR #3495, opened 2026-05-08
Fix eye icon to show the actual secret key

Aqil-Ahmad:
resolves part of #609 

image

## Summary by Sourcery

Bug Fixes:
- Fix secret key input so the eye icon reveals the real key value rather than a redacted placeholder.


Aqil-Ahmad:
the secret key is marked required but its not needed for stripe payments

sarafarajnasardi:
@codex review

sarafarajnasardi (changes requested):
Please address the codex review

Saksham-Sirohi (changes requested):
please dread carefully in the current area changes leaves keys open to extensions etc and may lead to potential problems

Aqil-Ahmad:
> please dread carefully in the current area changes leaves keys open to extensions etc and may lead to potential problems

i think then we should remove the eye icon completely so its not visible at all(but that contradicts what the issue asks). other way could be to keep ***** in the form, and when the eye is clicked, make an API call to fetch the real value. but that requires significant changes. @Saksham-Sirohi what do you suggest?

Saksham-Sirohi:
Closing this pr as directed earlier, please make a new one in the correct area

PR #3504, opened 2026-05-10
fix(api): return HTTP 400 on CheckInError in check-in redeem endpoint

### LINKED ISSUE #3503: fix(api): check-in endpoint returns HTTP 201 on `CheckInError` instead of HTTP 400
**Describe the bug**
The check-in redeem endpoint returns **HTTP 201 Created** when a `CheckInError` is raised (e.g. ticket already scanned), instead of **HTTP 400 Bad Request**.

**Affected code:** [`app/eventyay/api/views/checkin.py`, lines 674-710](https://github.com/fossasia/eventyay/blob/dev/app/eventyay/api/views/checkin.py)

```python
except CheckInError as e:
    ...
    return Response(
        {
            'status': 'redeemed',   # misleading - this is an error branch
            'reason': 'Already checked in',
            ...
        },
        status=201,  # wrong: 201 means "Created successfully"
    )
```

A client that checks `response.status_code` to distinguish success from failure will incorrectly treat a failed/duplicate check-in as a successful one, breaking integrations that follow standard HTTP semantics.

**Expected behavior**
The endpoint should return **HTTP 400** with `"status": "error"` when a `CheckInError` occurs - consistent with all other error branches in the same view (`_handle_no_candidates` returns 404, `_handle_ambiguous_candidates` returns 400).

```python
return Response(
    {
        'status': 'error',
        'reason': e.code,
        ...
    },
    status=400,
)
```

**Screenshots**
N/A - this is a code-level API response bug.

**Additional context**
- The `_handle_ambiguous_candidates` helper in the same file correctly returns `status=400` for its error branch, making this an inconsistency within the same function.
- - The `'status': 'redeemed'` field in the error response body is also semantically incorrect; it should be `'status': 'error'` to match the rest of the error responses.
- - - I am happy to submit a PR for this fix.
- - -

maniktyagi04:
Fixes #3503

## What this PR does

The `_redeem_process` function in `api/views/checkin.py` caught `CheckInError` and returned **HTTP 201** with `'status': 'redeemed'` Fixes #3503 
## What this PR does

The `_redeem_process` function in `api/views/checkin.py` caught `CheckInError` and returned **HTTP 201** with `'status': 'redeemed'` - both of which are semantically wrong for an error response. This fix brings the `CheckInError` branch in line with the other error branches in the same function.

## Changes

**`app/eventyay/api/views/checkin.py`**

| Field | Before | After |
|---|---|---|
| HTTP status | `201` | `400` |
| `status` field | `'redeemed'` | `'error'` |
| `reason` field | `'Already checked in'` (hardcoded) | `e.code` (from exception) |
| `reason_explanation` | missing | `None` (matches other error branches) |

## Why this matters

HTTP 201 means "resource created successfully". Returning it on a check-in error means any client that checks `response.status_code` will incorrectly treat a failed or duplicate check-in as a success. The `_handle_ambiguous_candidates` helper in the same file already returns `400` correctly - this was an inconsistency.

## Testing

- Trigger a duplicate check-in (scan the same ticket twice) via the API.
- - First scan -> `HTTP 201`, `"status": "ok"` (unchanged, still correct).
- - Second scan -> now returns `HTTP 400`, `"status": "error"`, `"reason": "already_redeemed"` (was `HTTP 201`, `"status": "redeemed"` before this fix).

## Summary by Sourcery

Bug Fixes:
- Change the check-in redeem endpoint to return HTTP 400 with an error status and structured reason details when a CheckInError occurs instead of incorrectly returning a 201 success response.

### DIFF
--- a/app/eventyay/api/views/checkin.py
+++ b/app/eventyay/api/views/checkin.py
@@ -700,13 +700,14 @@ def _redeem_process(
 
             return Response(
                 {
-                    'status': 'redeemed',
-                    'reason': 'Already checked in',
+                    'status': 'error',
+                    'reason': e.code,
+                    'reason_explanation': None,
                     'require_attention': op.require_checkin_attention,
                     'position': position_data,
                     'list': MiniCheckinListSerializer(list_by_event[op.order.event_id]).data,
                 },
-                status=201,
+                status=400,
             )
         else:
             serializer_context = _setup_context(request, expand, op.order.event, pdf_data, user, auth)


Aqil-Ahmad (approved):

Saksham-Sirohi:
Thank you for your contribution

PR #3540, opened 2026-05-11
fix: correct incorrect HTTP status codes in event and submission views

### LINKED ISSUE #3290: Bug: Wrong HTTP 503 status code used for non-infrastructure errors
## Problem
HTTP 503 (Service Unavailable) is returned in 4 places where it is 
semantically incorrect. 503 means the server is down, but these are 
logic/data errors — not infrastructure failures.

## Affected Files
- api/views/event.py line 473 → changed to 404 (data not found)
- api/views/order.py line 558 → changed to 500 (server error)
- api/views/submission.py line 528 → changed to 500 (server error)
- api/views/submission.py line 581 → changed to 500 (server error)

## Why It Matters
API clients use HTTP status codes to handle errors. Getting 503 makes 
them think the server is completely down and may retry unnecessarily 
or show wrong error messages to users.

prajwal2005singh:
**## Description**

This PR replaces semantically incorrect HTTP 503 responses in `event.py` and `submission.py`.

**### Changes**

* Changed `503 Service Unavailable` → `404 Not Found` in `EventThemeView` when the requested theme configuration is missing.
* Changed `503 Service Unavailable` → `500 Internal Server Error` in deprecated submission favourite handlers for unexpected internal exceptions.

**### Why**

HTTP 503 indicates temporary service unavailability or infrastructure failure. These branches instead represent:

* missing resources/configuration (`404`)
* unexpected internal application errors (`500`)

I intentionally did not modify the `order.py` `SendMailException` branch because it may represent a temporary downstream mail-service failure, for which `503 Service Unavailable` can still be semantically appropriate.

Fixes #3290

## Summary by Sourcery

Correct HTTP status codes returned by event theme and submission favourite views to better reflect missing resources and internal errors.

Bug Fixes:
- Return 500 Internal Server Error instead of 503 Service Unavailable for unexpected exceptions in deprecated submission favourite handlers.
- Return 404 Not Found instead of 503 Service Unavailable when an event's theme configuration is missing.

### DIFF
--- a/app/eventyay/api/views/event.py
+++ b/app/eventyay/api/views/event.py
@@ -595,7 +595,7 @@ class EventThemeView(APIView):
             )
             return Response(
                 'error happened when trying to get theme data of event: ' + kwargs['event_id'],
-                status=404,
+                status=HTTP_404_NOT_FOUND,
             )
 
 
--- a/app/eventyay/api/views/event.py
+++ b/app/eventyay/api/views/event.py
@@ -595,7 +595,7 @@ class EventThemeView(APIView):
             )
             return Response(
                 'error happened when trying to get theme data of event: ' + kwargs['event_id'],
-                status=503,
+                status=404,
             )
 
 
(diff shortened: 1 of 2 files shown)


Saksham-Sirohi:
Thank you for your contribution, the changes does not go in the right direction

PR #3546, opened 2026-05-13
fix(cfp): add public visibility for content locale

Sak1012:
This PR adds public visibility toggle for content locale.

image
image

When turned off:
image


Saksham-Sirohi (changes requested):
@Sak1012 The language field is still displayed even though I have turned public off; also, this should not have been toggleable in the first place, it should have appeared disabled. also in info we should inform that content languages need to be enabled in event settings in order to have appropriate languages as options

Saksham-Sirohi (approved):

mariobehling (approved):

PR #3654, opened 2026-05-26
Feature: include avatars import in speaker import 

ArnavBallinCode:
part of #3435 

### Problem
During speaker CSV import, image URLs provided in the `Picture` column were ignored. As a result, speaker avatars were only available if uploaded manually through the profile form.

#### Import field mapping
**`importers.py`**
- Added `avatar_url` to `SPEAKER_IMPORT_FIELDS`
- Added automatic mapping suggestions for:
  - `picture`
  - `photo`
  - `image`
  - `avatar`
  - URL variants

This enables common CSV image columns to map automatically.

#### Avatar download and storage
**`talkimport.py`**
Updated `_set_external_avatar_url()` to download and save external images directly into `user.avatar` instead of only storing the URL.

Additional improvements:
- Stream-based download handling
- Separate timeout configuration:
  - Connect: **5s**
  - Read: **10s**
- Redirects disabled by default with URL re-validation before following
- Fallback behavior: if image download fails, store the external URL in `user.profile` so the avatar still remains usable

#### Processing flow improvement
- `process_image()` is triggered **after `user.save()`**
- Prevents potential race conditions with async image processing

### Result
Users can now import speaker avatars directly from CSV image URLs, with locally stored images, safer download handling, and improved import reliability.

image

image

image


app/eventyay/base/services/talkimport.py
```diff
@@ -691,16 +693,66 @@ def _set_external_avatar_url(user: User, avatar_url: str) -> list[str]:
...
+        response.raise_for_status()
+
+        # Guard against huge files (10 MB cap)
+        max_bytes = 10 * 1024 * 1024
+        chunks = []
+        total = 0
+        for chunk in response.iter_content(chunk_size=65536):
+            total += len(chunk)
+            if total > max_bytes:
+                raise ValueError('Avatar image exceeds 10 MB limit')
```
Saksham-Sirohi, line 732: Good thing that we have a URL fallback, but I think we can add compression to this so that in rare cases when the URL is invalid, we still have an image

ArnavBallinCode: I think compression solves a different case here. If the URL itself is invalid/unreachable, there isn't any image content available to compress. The current implementation already falls back to storing the external URL when download fails. Optimizing large valid images via compression/resizing can be considered separately as a follow-up enhancement.

Sak1012 (approved):

PR #3663, opened 2026-05-26
fix: handle missing user emails in schedule notifications and gravatar generation

ArnavBallinCode:
part of #3435 

## Description

This PR fixes crashes and invalid behavior caused by users/speakers with missing or invalid email values.

Previously, several code paths assumed that every user always had a valid email address. Imported or incomplete user records could contain `None`, empty strings, or whitespace-only values, causing failures in schedule notifications and Gravatar generation.

### `mail.py`

Collect valid recipient email addresses before creating the recipient list.

Before:

```python
address = ','.join(user.email for user in users)
```

After:

```python
addresses = [user.email for user in users if user.email]

if not addresses:
    raise SendMailException(
        'Cannot create mail without at least one valid recipient email address.'
    )

address = ','.join(addresses)
```

This prevents mail generation from failing due to invalid recipient values and avoids attempting to send emails with empty recipient lists.

### `auth.py`

Handle missing and whitespace-only email values safely when generating Gravatar parameters.

Before:

```python
return md5(self.email.strip().encode()).hexdigest()
```

After:

```python
if not (self.email or '').strip():
    return None
```

Also updates the return type:

```python
def gravatar_parameter(self) -> str | None
```

### `tasks.py`

Add a guard in `gravatar_cache()` to skip Gravatar requests when no valid email exists:

```python
if not user.gravatar_parameter:
    logger.warning(...)
    return
```

This prevents invalid Gravatar URL generation and unnecessary requests.

### Tests

Extended `test_gravatar_parameter` to cover:

- `email=None`
- whitespace-only emails (`"   "`)

## Current behavior

- Notification generation may crash with users that have missing emails
- Schedule release can fail due to notification creation
- Gravatar generation can fail for users with invalid email values
- Background tasks may attempt invalid Gravatar requests

## Expected behavior

- Invalid email values are handled safely
- Notifications only use valid recipient addresses
- Explicit mail exceptions are raised when no recipients exist
- Avatar generation gracefully skips users without valid emails
- Existing behavior for valid users remains unchanged

## Testing

- Verified notification creation with missing email values
- Verified Gravatar generation with `None` and whitespace-only emails
- Verified Gravatar cache exits safely for invalid users
- Added unit tests for edge cases

### DIFF
--- a/app/eventyay/base/models/auth.py
+++ b/app/eventyay/base/models/auth.py
@@ -854,6 +854,8 @@ the eventyay team"""
 
     @cached_property
     def gravatar_parameter(self) -> str:
+        if not self.email:
+            return ''
         return md5(self.email.strip().encode()).hexdigest()
 
     @cached_property
--- a/app/eventyay/base/models/auth.py
+++ b/app/eventyay/base/models/auth.py
@@ -853,9 +853,9 @@ the eventyay team"""
         return str(uuid.uuid5(uuid.NAMESPACE_URL, f'acct:{self.email.strip()}'))
 
     @cached_property
-    def gravatar_parameter(self) -> str:
-        if not self.email:
-            return ''
+    def gravatar_parameter(self) -> str | None:
+        if not (self.email or '').strip():
+            return None
         return md5(self.email.strip().encode()).hexdigest()
 
     @cached_property
--- a/app/eventyay/base/models/auth.py
+++ b/app/eventyay/base/models/auth.py
@@ -850,7 +850,8 @@ the eventyay team"""
 
     @cached_property
     def guid(self) -> str:
-        return str(uuid.uuid5(uuid.NAMESPACE_URL, f'acct:{self.email.strip()}'))
+        identifier = (self.email or '').strip() or (self.code or str(self.pk))
+        return str(uuid.uuid5(uuid.NAMESPACE_URL, f'acct:{identifier}'))
 
     @cached_property
     def gravatar_parameter(self) -> str | None:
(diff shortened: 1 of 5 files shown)


Rachit7168:
Please resolve comments

Rachit7168 (approved):
LGTM : 


### What I tested(Used a script to test)
1. **`User.gravatar_parameter`** (`auth.py`)
   - `email=None` → `None`
   - `email='   '` → `None`
   - Valid email → stable MD5 hash (matches existing expectations)
2. **`gravatar_cache`** (`person/tasks.py`)
   - User with `get_gravatar=True` and no valid email
   - Confirmed no HTTP request to Gravatar
   - Confirmed `get_gravatar` is set to `False` and a warning is logged
3. **`MailTemplate.to_mail()`** (`mail.py`) with `commit=False`
   - User with `None` / whitespace-only email → draft mail with `to=None` (no crash)
   - Same user with `skip_queue=True` → `SendMailException` with a clear message
   - User with valid email → `to` set correctly

Saksham-Sirohi (changes requested):
Schedule notifications on actual schedule release/regeneration use generate_notifications(save=True) → to_mail(commit=True), which stores recipients in mail.to_users. 
When those queued mails are sent, QueuedMail.send() builds recipients with:
to += [user.email for user in self.to_users.all()] (no filtering) 
Meaning imported speakers with missing emails can still break at send time (or attempt invalid sends).

@ArnavBallinCode, please confirm that no such breakdown will happen

mariobehling:
@Saksham-Sirohi This is a valid point. So, let's think how this should work. There is actually another related question. Sometimes we have VIP speakers and they will not enter their info into the system. The organisers will enter the info. In the previous system we had an option "Email is not required" that could only be set by organisers. In that case speaker emails simply don't get sent. So, as far as I can see the result would be same here. Speaker emails simply don't get sent and their is no issue and nothing breaks really. 

@ArnavBallinCode Please share your insights.

Saksham-Sirohi (approved):

PR #3667, opened 2026-05-26
cfp: expose session type field in forms settings

### LINKED ISSUE #3664: [Talk] Always display 'Session Type' in Organizer Form Settings
The **Session Type** field is completely hidden from the Organizer Forms Settings dashboard. It only dynamically triggers and forces itself as a required field on the public-facing proposal form if an event happens to have more than one session type configured. Organizers cannot see, configure, or toggle this field ahead of time from the backend.

**Organizer Form**

Image

**Public Form** 

Image

### Steps to Reproduce
- Go to Call for Proposals > Session Types
- Test Case A: Set one Session Type only, go **Call for Proposals > Forms** observe that the entry is invisible in the backend. Observe what is shown on the public CFP
- Test Case B: Set two or more Session Types, go **Call for Proposals > Forms** observe that the entry is invisible in the backend. Observe what is shown on the public CFP

### Expected behavior
The Session Type configuration entry should always be visible in the Organizer Forms Settings, regardless of whether the event currently has one session type or multiple session type options configured.

1. Permanent Visibility: Always expose the "Session Type" row in the form settings backend.
2. Placement: Position it directly below the Description field configuration.
3. Organizer Control: Give organizers the option to enable, disable, or mark it as required/optional themselves, overriding any hardcoded conditional logic.
4. In the organizer backend this feature is called session type, but in the speaker submission it is called "session format". Rename it it "Session type" everywhere.

Rachit7168:
Closes : #3664

https://github.com/user-attachments/assets/3e671dba-bc39-4b6a-bd40-98336c54f575

### DIFF
--- a/app/eventyay/submission/forms/submission.py
+++ b/app/eventyay/submission/forms/submission.py
@@ -176,7 +176,18 @@ class InfoForm(
             elif self.event.cfp.default_type and self.event.cfp.default_type.pk in pks:
                 self.default_values['submission_type'] = self.event.cfp.default_type
             elif pks:
-                self.default_values['submission_type'] = submission_types.filter(pk__in=pks).first()
+                self.default_values['submission_type'] = submission_types.filter(pk__in=pks).order_by('pk').first()
+            else:
+                # Event misconfiguration: organizer hid the field, but no valid types exist.
+                # Keep the field visible with a clear message so the user gets feedback.
+                self.fields['submission_type'].queryset = submission_types.none()
+                self.fields['submission_type'].required = False
+                self.fields['submission_type'].empty_label = _('No session type is currently available')
+                self.fields['submission_type'].help_text = _(
+                    'No session type is currently available for submissions. '
+                    'Please contact the event organizers.'
+                )
+                return
             self.fields.pop('submission_type')
             return
 
@@ -202,8 +213,17 @@ class InfoForm(
             default_type = self.event.cfp.default_type
             if default_type and default_type.pk in allowed_pks:
                 return default_type
-            qs = self.fields['submission_type'].queryset
+            qs = self.fields['submission_type'].queryset.order_by('pk')
             return qs.first()
+        if visibility == 'do_not_ask' and value is None:
+            # If the field was kept visible due to misconfiguration, fail fast with
+            # a user-facing validation error instead of a save-time integrity error.
+            raise forms.ValidationError(
+                _(
+                    'No session type is currently available for submissions. '
+                    'Please contact the event organizers.'
+                )
+            )
         return value
 
     def _set_locales(self):
--- a/app/eventyay/submission/forms/submission.py
+++ b/app/eventyay/submission/forms/submission.py
@@ -14,6 +14,7 @@ from eventyay.base.models import (
     TalkQuestionTarget,
     Track,
 )
+from eventyay.base.models.cfp import default_fields as cfp_default_fields
 from eventyay.base.models.resource import get_slide_resources
 from eventyay.cfp.forms.cfp import CfPFormMixin
 from eventyay.common.forms.fields import ImageField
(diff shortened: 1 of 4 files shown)


sarafarajnasardi:
@codex review

MukundC25 (changes requested):
Hi @Rachit7168, take a look at the video. When we disable the 'Session type' in the organizer form, it appears to be hidden in the public CfP form when filling, but still shows in submissions

https://github.com/user-attachments/assets/59ed16c9-8d3b-48cc-bf82-8429db3b781d

app/eventyay/submission/forms/submission.py
```diff
@@ -158,20 +159,72 @@ def _set_submission_types(self, instance=None):
...
+        visibility = self.event.cfp.fields.get(
+            'submission_type', defaults['submission_type']
+        )['visibility']
+        if visibility == 'optional' and value is None:
+            allowed_pks = getattr(self, 'allowed_submission_type_pks', set())
+            default_type = self.event.cfp.default_type
+            if default_type and default_type.pk in allowed_pks:
+                return default_type
+            qs = self.fields['submission_type'].queryset.order_by('pk')
+            return qs.first()
```
MukundC25, line 217: This effectively turns “optional” into “optional in UI, but always saved with some value if any types exist”.

MukundC25:
@hpdang, if auto-filling of session type is reasonable and intentional behavior, even if disabled just to ensure a submission always has a type, then the changes are fine and working.

hpdang:
@MukundC25 good point did you test the change of @Rachit7168 locally? What is the behavior when session type is disabled?

SxxAq (changes requested):
When there are no session types configured and Session Type visibility is set to `optional` or `required`, the form shows an empty dropdown but submission fails with an error. This is confusing because the field appears usable even though no valid options exist.

Image

### Root cause

* `submission_type` is `NOT NULL` in the model.
* Organizers can hide all session types while still keeping the field visible.

### Suggested fix

Add validation in `CfPSettingsForm.clean()` to ensure Session Type cannot be set to `optional` or `required` unless at least one session type exists:

```python
def clean(self):
    cleaned_data = super().clean()

    if cleaned_data.get('cfp_ask_submission_type') != 'do_not_ask':
        if not SubmissionType.objects.filter(event=self.instance).exists():
            raise ValidationError(
                _('Session Type visibility requires at least one session type to be configured.')
            )

    return cleaned_data
```

mariobehling:
> @hpdang, if auto-filling of session type is reasonable and intentional behavior, even if disabled just to ensure a submission always has a type, then the changes are fine and working.

This is valid and before we consider a more complex implementation here we can consider the current functionality as good enough and working close to what is expected. Hence closing this PR for now. Thanks for working on this!

hpdang:
This is related to an issue users recently reported "We are currently finalizing the questions for the call for submissions and noticed that the 'session type' question cannot be rearranged. Could you let me know how to move the question so it appears directly below the proposal title?" @mariobehling @Rachit7168 

To make this work, perhaps we can lock the session type in the form (just like the Title) to ensure it is always asked even if there is only one session type. What do you think?

Rachit7168:
Seems valid I agree with @hpdang ! But let's hear from @mariobehling , if the solution and flow is good to go , I can start working on it

hpdang:
@Rachit7168 even if we enforce the session type on the form, there should be an option to drag and drop its position on the form though.

Rachit7168:
Yes , So for the darg and drop , I will raise a PR for it soon but not sure about the issue this PR was solving @hpdang

hpdang:
You can't drag and drop it if the session type is not visible in the form. It is currently not visible. Let's wait for Mario's feedback

hpdang:
@Rachit7168 you will reopen this PR? or will make a new one?

Rachit7168:
@hpdang , will open a new one but yet to discuss in the meeting and I have removed the branch  so But I have a direction in mind

Rachit7168:
As the field is always enforced then I will just show it there with the option of drag and drop but the field could not be made optional as schedule etc also uses this fields !

Rachit7168:
Working on the PR will be opened soon

PR #3693, opened 2026-05-30
Auto-create tracks on session CSV import

ArnavBallinCode:
needed for #3435 

Previously, if a CSV row referenced a track that did not exist, the track was silently skipped and the session was left unlinked. Rooms are already auto-created; this brings the same behaviour to tracks.

Changes:
- Add TRACK_NAME_MAX_LENGTH constant and _TRACK_AUTO_COLORS palette.
- Add _resolve_or_create_track(): finds an existing Track by exact/partial name or pk, and creates a new one (with a rotating default colour) when no match is found, mirroring _resolve_or_create_room().
- Update _import_submission_row() to call _resolve_or_create_track instead of the read-only _resolve_track.
- Update the Track help-text in SessionImportProcessForm to inform users that new tracks are created automatically.

image
image


ArnavBallinCode:
@codex review

Rachit7168 (approved):
LGTM — mirrors _resolve_or_create_room() cleanly and fixes silent unlinked tracks on import. Smoke-tested auto-create, dedup, and color rotation locally. 
Optional follow-up: add a unit test and align track help text with the room field wording.

Saksham-Sirohi (approved):

Issue #3398, opened 2026-04-26
Webhook HTTP Request Missing Timeout and Inconsistent Error Logging


Manishnemade12:
**Describe the bug**

The `send_webhook` Celery task in `app/eventyay/api/webhooks.py` (lines 295–351) has three issues:

1. **No timeout on `requests.post()` (Line 316):** The outgoing HTTP request has no timeout. If the remote webhook endpoint hangs (accepts TCP connection but never responds), the Celery worker blocks **indefinitely**. Over time, enough hanging targets can exhaust the entire worker pool — a denial-of-service on the app's own task queue.

```python
# Line 316 — no timeout parameter
resp = requests.post(webhook.target_url, json=payload, allow_redirects=False)
```

2. **Full response body loaded into memory before truncation (Lines 325, 344):** While `resp.text` is sliced to `MAX_SIZE_CONFIG`, the **entire response** is first downloaded into memory. A malicious endpoint returning a multi-gigabyte body can OOM-crash the worker before truncation occurs.

3. **Missing `success=False` on error path (Lines 336–345):** When a `RequestException` occurs, the `WebHookCall` record is created without explicitly setting `success=False`, relying on the model default. The success path (line 326) explicitly sets `success=True/False` — this inconsistency is a maintenance hazard.

**Expected behavior**

- `requests.post()` should have a reasonable timeout (e.g., 30 seconds).
- Response body should be streamed/limited before loading into memory.
- The `success` field should be explicitly set to `False` on error paths for consistency.

**Screenshots**

N/A — code-level bug.

**Additional context**

Proposed fix (~55 lines of changes):

```diff
 @app.task(base=ProfiledTask, bind=True, max_retries=9, acks_late=True)
 def send_webhook(self, logentry_id: int, action_type: str, webhook_id: int):
+    WEBHOOK_TIMEOUT = 30  # seconds
+
     with scopes_disabled():
         webhook = WebHook.objects.get(id=webhook_id)
     with scope(organizer=webhook.organizer):
         ...
         t = time.time()
         try:
             try:
-                resp = requests.post(webhook.target_url, json=payload, allow_redirects=False)
+                resp = requests.post(
+                    webhook.target_url,
+                    json=payload,
+                    allow_redirects=False,
+                    timeout=WEBHOOK_TIMEOUT,
+                )
                 WebHookCall.objects.create(
                     ...
                 )
             except RequestException as e:
                 WebHookCall.objects.create(
                     ...
[...]

PR #3694, opened 2026-05-30
Webhook HTTP Request Missing Timeout and Inconsistent Error Logging fixed

Saptak-cyber:
Fixed this issue #3398
Description
This PR resolves three production reliability and safety issues within the send_webhook Celery task (app/eventyay/api/webhooks.py):

Timeout on requests.post(): Added a WEBHOOK_TIMEOUT = 30 (seconds) parameter to requests.post(). Without this, if an outgoing remote webhook endpoint hangs or processes requests indefinitely, the Celery worker would block indefinitely. This prevents cascade worker-pool exhaustion and eventual denial-of-service on the application's task queues.

Streamed & Limited Response Body: Instead of using resp.text directly (which downloads the entire payload into memory before performing Python slicing), we now invoke requests.post with stream=True and incrementally read from the stream in chunks using resp.iter_content(). The loop breaks immediately once RESPONSE_SIZE_WEBHOOK bytes have been received. This prevents malicious/misconfigured endpoints from sending multi-gigabyte responses that OOM-crash the Celery workers.

Explicit success=False on Error Paths: Explicitly assigned success=False to the WebHookCall object inside the RequestException block. This ensures data schema consistency and eliminates the maintenance hazard of relying on the implicit model column default, matching the style of the happy path.

Proposed Changes

app/eventyay/api/webhooks.py
Introduced a module-level constant WEBHOOK_TIMEOUT = 30
Modified the HTTP post call to utilize stream=True and timeout=WEBHOOK_TIMEOUT
Read response content incrementally in 4096-byte chunks and decoded with 'replace' error handling to prevent decoding crashes
Added explicit success=False inside the catch block for RequestException


MukundC25 (changes requested):
@Saptak-cyber, thanks for your work. Please address the Copilot reviews and update the branch with the latest dev

MukundC25:
Thanks for the update, closing this one to avoid duplicate review/discussion. Please continue the conversation on the latest PR https://github.com/fossasia/eventyay/pull/3949

PR #3704, opened 2026-05-31
Fix log entries moving out of the panel

### LINKED ISSUE #2872: Email text overflows in Organizer "Change history" panel UI

Aqil-Ahmad:
fixes #2872 

we can also append long words/emails with "...", for now wrapping to next row
image

## Summary by Sourcery

Bug Fixes:
- Wrap long log entry text in the orders panel to keep it contained within the visible area.


SxxAq (commented):
@Aqil-Ahmad Could you please go through the related issue thread once? I think this might be a stale issue since the bug was not reproducible.

Were you able to reproduce it on your end? A few of us tried earlier but couldn’t reproduce the issue.

MukundC25:
> @Aqil-Ahmad Could you please go through the related issue thread once? I think this might be a stale issue since the bug was not reproducible.
> 
> Were you able to reproduce it on your end? A few of us tried earlier but couldn’t reproduce the issue.

@SxxAq, it seems to be there. I checked just now with the latest dev (varies across screen sizes)


https://github.com/user-attachments/assets/55e408b7-55ad-4947-a686-f6cef44c8868

Saksham-Sirohi (approved):

PR #3748, opened 2026-06-02
Fix venue map geocoding for international addresses

### LINKED ISSUE #3743: Fix embedded event map showing incorrect event location
On public event pages, the embedded map does not always show the correct location of the event.

In the example shown at https://eventyay.com/ev/88882f3e/ the event address is:

True Digital Park West, Soi Punnawithi 4, Bang Chak Subdistrict, Phra Khanong District, Bangkok, 10260, Thailand

However, the embedded map marker appears in the wrong location and does not correctly represent the venue address.

Image

## Current behavior

The event page shows a map under the Getting here section.

The address text is displayed, but the embedded map marker does not match the actual event location.

This can confuse attendees who rely on the event page for navigation.

## Expected behavior

The embedded map should show the correct location for the event venue.

The marker position should match the event address and, where available, the stored event latitude and longitude.

## Implementation details

* Review how the map location is currently generated on public event pages.
* Check whether the map uses:

  * stored latitude and longitude
  * geocoded address data
  * venue address text
  * default fallback coordinates
* Ensure the map uses the most accurate available location data.
* If latitude and longitude are available for the venue, use them directly instead of relying only on address geocoding.
* If coordinates are missing, geocode the full venue address reliably.
* Ensure the full address is passed to the map provider and not only a partial address.
* Check whether the map falls back to an incorrect default location when geocoding fails.
* Add handling for geocoding failures so the map does not silently show an incorrect marker.
* Ensure the displayed address and the map marker are consistent.
* Test the fix with international addresses, including venues in Thailand and other non-European locations.

## Rationale

The map is used by attendees to find the event venue. If the marker is wrong, attendees may go to the wrong place or lose trust in the event information.

The address text and embedded map should always point to the same venue location.

## Acceptance criteria

* The embedded map marker matches the event venue address.
* If venue latitude and longitude are available, the map uses them directly.
* If coordinates are not available, the full event address is geocoded correctly.
* The map does not silently show an incorrect fallback marker when geocoding fails.
* The address text and map marker are consistent.
* The fix works for international addresses.
* The issue is tested with the example address in Bangkok.
* Relevant tests or documented manual testing steps are included.

Rachit7168:
Fix : #3743
Fixes incorrect venue map markers on public event pages and unreliable address geocoding in event settings.
- Resolves public map coordinates from stored lat/lon when valid, otherwise geocodes the full localized address (with simpler fallback queries when providers reject long addresses).
- Consolidates geocoding in control/views/geo.py with provider fallthrough (OpenCage/MapQuest → Nominatim) and avoids caching failed lookups.
- Restores the admin settings map widget and improves location field detection for i18n address fields.
- Hides the public map when geocoding fails instead of showing a wrong marker; still shows the address text.

https://github.com/user-attachments/assets/970a72c1-b3ed-42b6-8fd3-4191de5939a5


Rachit7168:
@codex review

Rachit7168:
Test is not failed because of the changes

Saksham-Sirohi:
> The test is not failing because of the changes

CI was down yesterday; I have re-run the tests for you

prayag78 (changes requested):
@Rachit7168 On mobile view the `Geocoding data © OpenStreetMap` link is misaligned, it shows between the latitude and longitude inputs

Image

ArnavBallinCode (approved):

Saksham-Sirohi (approved):

PR #3753, opened 2026-06-02
fix: hide tickets tab when presale is over and products are hidden

### LINKED ISSUE #3750: Hide Tickets Navigation Tab When No Ticket Listings Are Visible
## Describe the bug

When ticket visibility is disabled after the presale period (or when tickets are otherwise hidden from the public event page), the **Tickets** navigation tab is still displayed in the event header.

Clicking the Tickets tab normally scrolls the page to the Tickets section. However, when ticket listings are hidden:

- The Tickets section is not rendered.
- No ticket products are visible.
- Clicking the Tickets tab does not navigate anywhere meaningful.
- The user remains on the same page.
- The tab appears functional even though its target content does not exist.

This creates a confusing user experience because the interface suggests that ticket information is available when it is not.

## Expected behavior

The Tickets navigation tab should only be displayed when a visible Tickets section exists on the page.

If ticket listings are hidden due to:

- Presale not yet started
- Presale ended
- Product visibility settings
- Event ticket visibility configuration

then:

- The Tickets section should not be rendered.
- The Tickets navigation tab should also be hidden.
- Navigation should only expose sections that are currently available to the user.

## Steps to reproduce

1. Create an event with ticket products.
2. Configure product visibility so tickets are hidden after the presale period.
3. Visit the public event page after the presale has ended.
4. Observe that:
   - The Tickets navigation tab is still displayed.
   - No ticket products are visible in the page content.
5. Click the Tickets tab.

## Actual behaviour

- Tickets tab remains visible.
- No ticket content exists.
- Clicking the tab does not lead to a visible Tickets section.

## Expected behaviour

- Tickets tab should automatically disappear when there are no visible ticket products or no Tickets section available.
- Navigation items should reflect actual visible content on the page.

## Screenshots

Image

## Additional context

The Tickets tab appears to function as an anchor link to the Tickets section further down the page. When the Tickets section is removed or hidden, the corresponding navigation item should also be removed to prevent broken or misleading navigation.

This issue is particularly noticeable when using the "Show products outside presale period" setting and other product visibility configurations that result in no publicly visible ticket listings.

MukundC25:
Fixes #3750 

Previously, clicking the "Tickets" tab in this state would simply scroll the user down to an empty or non-existent tickets section, leading to a confusing user experience.

#### Root Cause
The [tickets_tab_visible] template tag was checking for product availability using `event.products.filter_available()`, but it did not account for the event-level presale timeframe or the specific `show_products_outside_presale_period` setting. Because of this, it would incorrectly return `True` even when the template logic suppressed the actual ticket list.

#### Changes Made
* Updated the [tickets_tab_visible] template tag in [app/eventyay/common/templatetags/event_tags.py] to evaluate the active presale state.
* Added a condition to return `False` if `ev.presale_is_running` is false AND `event.settings.show_products_outside_presale_period` is disabled.

https://github.com/user-attachments/assets/6868027d-5a08-46e5-ad1e-8dbdab5b07b2


Aqil-Ahmad (commented):
tested. works but not sure if its good to hide it when presale ends/not-started as its more for accessibility.

Rachit7168:
@codex review

Rachit7168 (approved):
LGTM , Tested

https://github.com/user-attachments/assets/781a2ad3-be52-4c72-b86a-1479efb1a8db

sarafarajnasardi (approved):
LGTM

Sak1012 (approved):

Saksham-Sirohi (approved):

Issue #3765, opened 2026-06-02
Extend organisers into community profiles with follow functionality


mariobehling:
eventyay should support group-like community functionality by extending the existing organiser concept instead of introducing a separate Groups model at this stage.

Platforms such as Meetup use groups, while platforms such as Eventbrite allow users to follow organisers. eventyay already has organisers as persistent entities that can host multiple events. Therefore, the first implementation should start by improving organiser pages as public community profiles and adding follow/unfollow functionality for users.

The goal is to let users follow organisers, discover their upcoming events, and receive updates when organisers publish new public events. This gives organisers a stronger public presence and allows them to build recurring communities around their events without duplicating the existing organiser model.

This issue focuses on an implementation in two phases:

* Phase 1: Extend organisers with follow functionality.
* Phase 2: Add organiser community features as an extension of organisers.

## Current behavior

* Organisers exist as platform entities.
* Organisers can create and manage events.
* Public organiser pages exist but are still limited and need improvement.
* Users cannot follow organisers.
* Users do not have a dedicated way to see organisers they follow.
* Organisers cannot notify followers about newly published events.
* Organiser pages do not yet function like community pages.

## Expected behavior

Organisers should become more community-oriented public profiles.

Users should be able to:

* Visit a public organiser page.
* Follow an organiser.
* Unfollow an organiser.
* See organisers they follow in their user dashboard.
* Discover upcoming public events from followed organisers.
* Receive notifications when followed organisers publish new public events, depending on notification preferences.

Organisers should be able to:

* Have a useful public organiser profile.
* Show upcoming and past public events.
* Build a follower base.
* Configure whether their organiser profile can be followed.
* Eventually communicate with followers through future organiser updates or announcements.

### 1. Improve public organiser page as a community profile

* [ ] Show organiser name.
* [ ] Show organiser logo.
* [ ] Show organiser header image if configured.
* [ ] Show organiser description.
* [ ] Show organiser links where available.
* [ ] Show upcoming public events from the organiser.
* [ ] Show past public events where appropriate.
* [ ] Add clear empty states when the organiser has no public events.
* [ ] Ensure the public organiser page works on desktop and mobile.
* [ ] Ensure organiser design settings apply correctly.

### 2. Add follow and unfollow functionality

* [ ] Add a Follow button to public organiser pages.
* [ ] Add an Unfollow state or button when the current user already follows the organiser.
* [ ] Require login before a user can follow an organiser.
[...]

PR #3772, opened 2026-06-04
feat: extend organizers with follow functionality

prayag78:
part of #3765 

This PR adds:
- Organizers follow/unfollow functionality
- Notify followers about newly published events
- Add organizer community settings
- Fix some organizer public page UI issues

The follow-up PRs will cover:
- Followed organizers and events to the user dashboard
- Other UI enhancements in the organiser pages
- More organiser community features

Screenshot 2026-06-04 165301

Screenshot 2026-06-06 092842

Screenshot 2026-06-04 171020

Screenshot 2026-06-04 165624


sarafarajnasardi:
@prayag78 please address the copilot reviews...

Aqil-Ahmad:
@prayag78 only the number of followers are shown for now, right?

app/eventyay/base/services/notifications.py
```diff
@@ -133,3 +135,93 @@ def send_notification_mail(notification: Notification, user: User):
...
+@scopes_disabled()
+def notify_organizer_followers(event_id: int):
+    """
+    Send an email notification to all followers of an organizer when a new event is published.
+
+    This task is dispatched after an event transitions to ``live=True``.
+    It respects the organizer's ``community_follow_enabled`` setting and each
+    follower's locale / timezone.
+    """
+    with transaction.atomic():
```
Rachit7168, line 150: Because the transaction.atomic() block ends immediately after this check, the lock on the Event row is released immediately. If this Celery task is triggered twice concurrently (e.g., double-clicks on publish, or task retries), both executions will sequentially acquire the lock, find no LogEntry, and then both will proceed to send duplicate emails to all followers. The actual log entry is only written at the very end of the function, leaving a massive race condition window.

app/eventyay/api/views/organizer.py
```diff
@@ -60,19 +62,63 @@ class OrganizerViewSet(viewsets.ReadOnlyModelViewSet):
...
-                return Organizer.objects.filter(pk__in=self.request.user.teams.values_list('organizer', flat=True))
+                qs = Organizer.objects.filter(pk__in=self.request.user.teams.values_list('organizer', flat=True))
         elif hasattr(self.request.auth, 'organizer_id'):
-            return Organizer.objects.filter(pk=self.request.auth.organizer_id)
+            qs = Organizer.objects.filter(pk=self.request.auth.organizer_id)
         else:
-            return Organizer.objects.filter(pk=self.request.auth.team.organizer_id)
+            qs = Organizer.objects.filter(pk=self.request.auth.team.organizer_id)
+
+        qs = qs.annotate(_follower_count=Count('followers'))
```
Rachit7168: Using Count() without distinct=True can lead to cartesian products and wildly inaccurate numbers if the base queryset (qs) has other JOINs added to it elsewhere in the code or in future updates. Recommendation: Always use distinct=True when counting reverse relations.

app/eventyay/base/services/notifications.py
```diff
@@ -133,3 +135,93 @@ def send_notification_mail(notification: Notification, user: User):
...
+    followers = (
+        OrganizerFollower.objects.filter(organizer=organizer)
+        .select_related('user')
+        .iterator()
+    )
+
+    tpl_html = get_template('pretixbase/email/organizer_follower_new_event.html')
+    tpl_plain = get_template('pretixbase/email/organizer_follower_new_event.txt')
+
+    for follower in followers:
```
Rachit7168, line 185: If an organizer accumulates a massive following (e.g., 20,000+ followers), enqueuing that many tasks synchronously inside a single loop could take a long time, potentially causing the Celery worker to exceed time limits or consume too much memory. Recommendation: This is acceptable for a V1 implementation. For future scaling, consider dispatching "batch" jobs or using Celery's chunk() capabilities.

app/eventyay/api/views/organizer.py
```diff
@@ -60,19 +62,63 @@ class OrganizerViewSet(viewsets.ReadOnlyModelViewSet):
...
+        if not request.user.is_authenticated:
+            return Response({'detail': _('Authentication required.')}, status=status.HTTP_401_UNAUTHORIZED)
+        organizer = self.get_object()
+        deleted, _ = OrganizerFollower.objects.filter(user=request.user, organizer=organizer).delete()
+        return Response({'following': False, 'deleted': deleted > 0}, status=status.HTTP_200_OK)
+
+    @action(detail=True, methods=['get'], url_path='followers')
+    def followers(self, request, *args, **kwargs):
+        organizer = self.get_object()
+        if not organizer.settings.get('community_follow_enabled', as_type=bool, default=True):
```
Rachit7168: The string 'Following is not enabled for this organizer.' is missing the _() translation wrapper, whereas it is correctly localized in the follow POST action slightly above it.

Rachit7168 (changes requested):
Thanks for working on this.

Please address

Aqil-Ahmad (changes requested):
@prayag78 when disabling the "Allow users to follow this organizer", the number of followers are not shown, even its enabled. Also, when "Allow users to follow this organizer" is disabled the current users who followed can't unfollow. 

I think it would be nice if such edge cases are covered with tests.

I couldn't confirm the email works as i wasn't receiving any, maybe i missed something but not sure.

Rachit7168 (approved):
LGTM , Approved 

https://github.com/user-attachments/assets/f6e067e1-47c1-4abb-acd2-2965d1afd9e1

https://github.com/user-attachments/assets/53968286-7103-4578-bace-0f071edf7282

Saksham-Sirohi (approved):
default values are true so these apply unless organizer chooses to disable them

PR #3860, opened 2026-06-09
Fix UI for search form in Talks

### LINKED ISSUE #3780: Talks' Sessions Search UI is broken
Talks' Sessions Search/Filter Form UI is broken when there are many sessions and the sidebar is open.

Image

Aqil-Ahmad:
fixes #3780

https://github.com/user-attachments/assets/b550e2f2-1650-43e5-b9b3-20ca7987d1a8

## Summary by Sourcery

Unify the talks submission search form with the shared filter actions component and adjust styling for the search form actions layout.

Bug Fixes:
- Fix layout and alignment of search/filter action buttons in the talks submission search form.

Enhancements:
- Reuse the common filter actions partial in the talks submission filter form instead of a dedicated search button.
- Add specific styling for search form filter action buttons to match the overall UI.


Rachit7168 (changes requested):
https://github.com/user-attachments/assets/d1c46022-bb0c-4ae6-852d-194ba41edb81

This approach is not user friendly. 
I cant know or see what filters I have applied , nor i can deselect any selected option. It only shows number but not the filter applied such as pending etc. And so on ....

Aqil-Ahmad:
> This approach is not user friendly. I cant know or see what filters I have applied , nor i can deselect any selected option. It only shows number but not the filter applied such as pending etc. And so on ....

yes. its only easy to select and deselect, since we can't show the selected options like it was before-- options are long and it will break the UI. another option can be to use multi-select checkbox based dropdown. I wasn't sure about this myself to either keep the current approach or use checkbox based.

Aqil-Ahmad:
image

Rachit7168 (changes requested):
Please check , I am getting this error on your branch , I tried other and also rebuild the web , error still persists !
Please update your branch once !
Image

Aqil-Ahmad:
@Rachit7168 thanks for reivew!! updated the branch please have a look now. works on my end

Saksham-Sirohi (changes requested):
fix search bar height

app/eventyay/orga/templates/orga/widgets/filter_checkbox_dropdown.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}
+<details class="filter-select" data-filter-select>
+  <summary class="filter-select__summary" data-summary>
+    <span class="filter-select__label">{{ widget.attrs.title }}</span>
+  </summary>
+  <div class="filter-select__menu">
+    {% for group, options, index in widget.optgroups %}
+      {% for option in options %}
+        <label class="filter-select__option">
+          <input type="checkbox" name="{{ widget.name }}" value="{{ option.value|stringformat:'s' }}"{% if option.selected %} checked{% endif %}>
```
sarafarajnasardi: This drops the behavior from `SelectMultipleWithCount`: options are no longer sorted by count, zero-count options are no longer hidden, and the count is no longer displayed. The forms still build `CountableOption` values and annotated counts, so this keeps the database/counting work but removes the user-visible result. Can we preserve the previous count behavior in this new checkbox dropdown?

sarafarajnasardi (changes requested):

app/eventyay/orga/templates/orga/widgets/filter_checkbox_dropdown.html
Aqil-Ahmad: ~I dropped it intentionally, the count is shown in the top left when a filter is applied. but i think you are right about the counting if it doesn't affect the filtered items count, i will have a look~

restored the count

Rachit7168 (approved):
LGTM, Tested : 

https://github.com/user-attachments/assets/877e4e6a-22d7-4fdc-90ca-1bfc388e9a34

sarafarajnasardi (approved):
LGTM

app/eventyay/orga/templates/orga/widgets/filter_checkbox_dropdown.html
Rachit7168: The hardcoded <input type="checkbox" ...> element drops standard Django option attributes (like disabled, id, custom data-* fields, etc.) because it doesn't unpack option.attrs.

Consider replacing the manual input with the standard Django attrs include to prevent side effects with other Django form features

app/eventyay/common/forms/widgets.py
```diff
@@ -144,48 +145,17 @@ class EnhancedSelectMultiple(EnhancedSelectMixin, SelectMultiple):
...
-                    selected=selected,
-                    index=index,
-                    count=count,
-                )
-            )
-        return [(None, result, 0)]
-
-    def create_option(self, name, value, label, *args, count=0, **kwargs):
-        label = f'{label} ({count})'
+class FilterCheckboxDropdown(CheckboxSelectMultiple):
```
Rachit7168, line 148: By removing SelectMultipleWithCount (which inherited from EnhancedSelectMultiple), we lost the color_field support for Tracks, Tags, and States. The PR removes color_field=... entirely from the widget initialization in submission.py and mails.py. Since we are moving away from Select2, we'll need to explicitly render the colors next to the checkbox labels in our new template if we want to retain them.

If dropping the colors was intentional to simplify the UI, please mention it. Otherwise, we should add color_field to FilterCheckboxDropdown.__init__ and pass it down to create_option, then add a small colored dot next to the label in filter_checkbox_dropdown.html.

app/eventyay/common/forms/widgets.py
```diff
@@ -144,48 +145,17 @@ class EnhancedSelectMultiple(EnhancedSelectMixin, SelectMultiple):
...
+    """A multi-select rendered as a native <details> dropdown of checkboxes."""
+
+    template_name = 'orga/widgets/filter_checkbox_dropdown.html'
+    option_template_name = 'django/forms/widgets/checkbox_option.html'
```
Rachit7168, line 152: While the count is now successfully appended to the label, the options are no longer sorted by count (from highest to lowest), and options with a count of 0 are no longer hidden. Additionally, the original get_count logic supported callables, which was dropped here.

To restore the full original behavior, we should override optgroups in FilterCheckboxDropdown and bring back the get_count helper.

Rachit7168 (changes requested):
Functionality wise working but Please address the comments !

app/eventyay/common/forms/widgets.py
Aqil-Ahmad: yes its intentional. the current dropdown uses the already being used checkboxes throughout talks area. adding colors for these will add complexity but not that much value.

app/eventyay/common/forms/widgets.py
Aqil-Ahmad: again the same reason. since the dropdown changed all these will add complexity but not much value also sorting a small number of options wouldn't improve ux that much.

the callable count wasn't used by any filter before.

Aqil-Ahmad:
@Rachit7168 thankss for review

resolved one of your comment and left reasons for others. please have a look and if you think we need those, let me know i'll add those back.

hpdang:
I can't figure out what is the original bug of this PR. But the new implementation is not better than what we currently have. Hiding all the Proposal States and Track Names are not useful from the organizers' point of view.

## current behavior
Screenshot from 2026-06-23 22-22-56

## this fix 
Screenshot from 2026-06-23 22-24-45

PR #3877, opened 2026-06-11
Fix LSP violation in Sig1TicketSecretGenerator and align protobuf field names

### LINKED ISSUE #3871: Liskov Substitution Principle (LSP) Violation in Ticket Secret Generator Subclasses
**Describe the bug**
The base class `BaseTicketSecretGenerator` in `app/eventyay/base/secrets.py` defines the interface method `generate_secret` with the parameter `attendee_name: str = None`. However, the subclass `Sig1TicketSecretGenerator` overrides this method but omits the `attendee_name` parameter from its signature. 

This violates the Liskov Substitution Principle (LSP). If a caller uses the base class interface and passes `attendee_name` (either positionally or via keyword arguments), invoking it on a `Sig1TicketSecretGenerator` instance raises a `TypeError` at runtime:
- Positionally: `TypeError: generate_secret() takes 5 positional arguments but 6 were given`
- Keyword: `TypeError: generate_secret() got an unexpected keyword argument 'attendee_name'`

**Expected behavior**
Subclasses overriding `generate_secret` must maintain signature compatibility with the base class so that they can be substituted transparently without raising `TypeError`.

**Screenshots**
N/A

**Additional context**
The base class definition:
```python
class BaseTicketSecretGenerator:
    def generate_secret(
        self,
        product: Product,
        variation: ProductVariation = None,
        subevent: SubEvent = None,
        attendee_name: str = None,
        current_secret: str = None,
        force_invalidate=False,
    ) -> str:
```

## The violating subclass definition:

```
class Sig1TicketSecretGenerator(BaseTicketSecretGenerator):
    def generate_secret(
        self,
        product: Product,
        variation: ProductVariation = None,
        subevent: SubEvent = None,
        current_secret: str = None,
        force_invalidate=False,
    ):
```

DashratRajpurohit:
Closes #3871

### Description of Changes
This PR resolves two critical backend issues within the ticket secret generation logic:
1. **Liskov Substitution Principle (LSP) Violation**: Fixed `Sig1TicketSecretGenerator.generate_secret` which was missing the `attendee_name` argument required by the base interface `BaseTicketSecretGenerator`. This caused inheritance mismatch errors.
2. **Protobuf Schema Alignment**: Fixed a crash where `Sig1TicketSecretGenerator` attempted to assign and read the `.product` field on a compiled Protobuf `Ticket` payload. The compiled schema (`pretix_sig1.proto`) explicitly defines this field as `.item` to retain wire-format compatibility with ticket scanner apps. We aligned the Python code to access `.item` instead, resolving the `AttributeError`.

### Verification
- Updated `tests/tickets/base/test_secrets.py` to use the modern `event.products` relation rather than the deprecated `event.items`.
- Added explicit unit tests to verify LSP compatibility across all generator subclasses.


Rachit7168 (changes requested):
Which Issue does this PR solves ? Create an issue if dose not exist also add screenshot of the issue or how was it produced !

Also please go through readme once and follow community guidelines

MukundC25 (changes requested):
@DashratRajpurohit, thanks for the effort! Changes look good. One minor nit: `test_lsp_signature_compatibility` doesn't actually need a DB, a `SimpleNamespace` mock would be lighter.

app/tests/tickets/base/test_secrets.py
```diff
@@ -116,3 +116,34 @@ def test_sig1_parse_invalid_secret_returns_none(event):
         invalid_proto_secret = base64.b64encode(generator._sign_payload(b'\xff')).decode()[::-1]
         assert generator._parse(invalid_proto_secret) is None
 
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('scheme', schemes)
+def test_lsp_signature_compatibility(event, scheme):
```
MukundC25, line 122: ptal

ArnavBallinCode (approved):

app/tests/tickets/base/test_secrets.py
ArnavBallinCode: nit?

MukundC25 (approved):

Saksham-Sirohi (approved):

PR #3878, opened 2026-06-11
Fix btn-default border disappearing when clicked

### LINKED ISSUE #3783: Buttons with class btn btn-default looses borders on-focus
Buttons with class btn btn-default are used in many places in the Tickets area, clicking these buttons make their borders invisible

Image

Mayur404:
Fixes #3783

The `btn-default` style uses an inset `box-shadow` to show its border because shared button styles set the real border width to `0`.

A global focus rule was removing `box-shadow` on `:focus:not(:focus-visible)`. As a result, clicking `btn btn-default` buttons made their border disappear in places where these buttons are used.

This PR removes that `box-shadow: none` reset so the existing button border remains visible after focus.

### Testing

* Reproduced the issue locally.
* Verified the fix locally in **Event Settings -> Payment**.
* Confirmed that `btn btn-default` buttons retain their border after being clicked/focused.

https://github.com/user-attachments/assets/4a26df47-885a-41b2-acea-1061633cd776

## Summary by Sourcery

Bug Fixes:
- Prevent btn-default borders from disappearing when buttons are clicked or focused by keeping their box-shadow intact.

### DIFF
--- a/app/eventyay/static/pretixbase/scss/_theme.scss
+++ b/app/eventyay/static/pretixbase/scss/_theme.scss
@@ -194,7 +194,6 @@ button:focus:not(:focus-visible),
 summary:focus:not(:focus-visible),
 [role="button"]:focus:not(:focus-visible) {
     outline: none;
-    box-shadow: none;
 }
 
 @include table-row-variant('success', lighten($brand-success, 40%));


Rachit7168 (changes requested):
The changes looks good , but can you check other parts of the system as well. As far as I know this thing is happening at most of the parts. Check and also upload a updated Screen-cast for the UI changes !

Mayur404:
Thanks for the review.

I tested the fix across multiple screens where `btn btn-default` is used:

Tested:
- Event -> Tickets -> Settings -> Payment
- Main Dashboard - > My Orders
- Main Dashboard - > My Events
- Event -> Tickets -> Settings -> Product
In all cases, the button border remains visible after click/focus. I’ve also uploaded an updated screen recording showing the behavior.

https://github.com/user-attachments/assets/eb12d971-46ac-4502-9bbd-5ef80228e624



https://github.com/user-attachments/assets/b3c8501d-b401-4254-b2bd-161076e515e3

Rachit7168 (approved):
LGTM ! 

https://github.com/user-attachments/assets/af974423-a572-43d6-bcc8-361f98cbb72a

Aqil-Ahmad (approved):

prayag78 (approved):

Saksham-Sirohi (approved):

PR #3882, opened 2026-06-11
Restore BBB server cost update command

### LINKED ISSUE #3881: Restore unavailable ``bbb_update_cost`` management command
## Problem

  The documented bbb_update_cost Django management command is unavailable:

  docker compose exec web python manage.py bbb_update_cost
  Unknown command: 'bbb_update_cost'

  ## Root Cause

  - The command is under eventyay/core/management/commands/, but eventyay.core is not registered in INSTALLED_APPS, so Django cannot discover it.
  - It imports get_url from the removed eventyay.core.services.bbb module instead of eventyay.base.services.bbb.

  ## Expected Behavior

  - python manage.py bbb_update_cost should be available.
  - Active BBB server costs should be updated based on running meetings.
  - Inactive servers should be ignored.
  - One unavailable server should not prevent other servers from being updated.

  ## Suggested Fix

  - Move/register the command under an installed Django application such as eventyay.base.
  - Import get_url from eventyay.base.services.bbb.
  - Add automated tests covering command discovery, cost calculation, inactive servers, and unavailable servers.

Image

sarafarajnasardi:
Fixes #3881 

## Description

  Fixes the unavailable bbb_update_cost management command.

  The command was located under eventyay.core, which is not an installed Django application, so Django could not discover it. It also imported the BBB URL helper from a removed
  module.

  ## Changes

  - Move bbb_update_cost into the installed eventyay.base application.
  - Import get_url from eventyay.base.services.bbb.
  - Correct command help and error messages.
  - Add tests covering:
      - Idle-server cost reset.
      - Running-meeting cost calculation.
      - Inactive-server filtering.
      - Continuing updates when one server is unavailable.

### DIFF
--- a/app/eventyay/base/management/commands/bbb_update_cost.py
+++ b/app/eventyay/base/management/commands/bbb_update_cost.py
@@ -6,13 +6,14 @@ from django.core.management.base import BaseCommand
 from lxml import etree
 
 from eventyay.base.models import BBBServer
-from eventyay.core.services.bbb import get_url
+from eventyay.base.services.bbb import get_url
+
 
 logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
-    help = "Updated load balancing cost of BBB servers"
+    help = "Update load balancing costs of BBB servers"
 
     def _update_cost(self, server: BBBServer):
         try:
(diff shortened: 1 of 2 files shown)


Rachit7168 (approved):
LGTM, The command seems to be working 

image

Saksham-Sirohi (approved):

Issue #3883, opened 2026-06-12
Quota availability waiting_list returns pending order count


Rachit7168:
## Description
`GET .../quotas/{id}/availability/` sets `waiting_list` from `qa.count_pending_orders`
instead of `qa.count_waitinglist`.

## Steps to reproduce
1. Create event with quota (size 10) and product.
2. Create WaitingListEntry for that product (no pending orders).
3. `GET /api/v1/organizers/{org}/events/{event}/quotas/{id}/availability/`

## Expected
`waiting_list: 1`, `pending_orders: 0`

## Actual
`waiting_list: 0`

-------------------------------------------

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0f8588c9-2c32-4729-846d-5454fbfc27f0" />

-------------------------------------------

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/06720557-faa6-48eb-8722-af9ea49428ff" />

-------------------------------------------

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/9bfb6af9-8772-4fc5-bfec-e91643473b9d" />

## Test Using code : 

`from eventyay.base.models import Event, WaitingListEntry`
`from eventyay.base.services.quotas import QuotaAvailability`
`event = Event.objects.get(slug='YOUR_EVENT')`
`quota = event.quotas.first()`
`product = quota.products.first()`
`WaitingListEntry.objects.get_or_create(`
`    event=event, product=product, email='test@example.com', defaults={'locale': 'en'}`
`)`
`qa = QuotaAvailability()`
`qa.queue(quota)`
`qa.compute()`
`print('count_waitinglist:', qa.count_waitinglist[quota])   # correct: 1`
`print('count_pending_orders:', qa.count_pending_orders[quota])  # 0`
`# API wrongly returns pending_orders for waiting_list`

Rachit7168:
Closing as done via #3947

PR #3947, opened 2026-06-14
Fix quota handling in payment and availability APIs

### LINKED ISSUE #3884: Payment create with state=confirmed silently fails on quota exceeded
## Description 

When creating a payment via `POST .../orders/{id}/payments/` with `"state": "confirmed"`, quota exhaustion during `payment.confirm()` is **silently swallowed**. The API returns **HTTP 201** with a **confirmed** payment object, even though the order was **not** marked as paid.

The sibling endpoint `POST .../payments/{local_id}/confirm/` handles the same failure correctly and returns **HTTP 400**.

There is a **related inconsistency** on `POST .../orders/{code}/mark_paid/`: it returns **HTTP 400** (correct status), but still leaves a **confirmed** payment in the database because `OrderPayment.confirm()` marks the payment confirmed *before* running quota checks in `_mark_paid()`.

---

Image

Image

Image

Rachit7168:
## Summary

Close [#3884](https://github.com/fossasia/eventyay/issues/3884) and 
Close[#3883](https://github.com/fossasia/eventyay/issues/3883).

This PR corrects three related quota/payment API bugs: silent failure on confirmed payment create, incorrect waiting-list counts in quota availability, and orphaned confirmed payments when mark-paid is blocked by quota.

### `app/eventyay/api/views/order.py`

**Re-raise `QuotaExceededException` as `ValidationError` in `PaymentViewSet.create`**

When creating a payment with `"state": "confirmed"`, quota exhaustion during `payment.confirm()` was previously swallowed (`except Quota.QuotaExceededException: pass`). The API returned HTTP 201 with a confirmed payment even though the order was not marked paid. The sibling `POST .../payments/{local_id}/confirm/` endpoint already returned HTTP 400 for the same failure. This change surfaces the quota error as HTTP 400 and rolls back the enclosing transaction so no misleading success response is returned.

**Add `get_order_for_nested_route()` and use it in `PaymentViewSet` / `RefundViewSet`**

Nested payment and refund routes are registered under `orders/<int:order>/` (order primary key), but the viewsets looked up orders by `code`. A request to `.../orders/27035/payments/` therefore searched for an order with code `"27035"` instead of pk `27035`, returning 404 even for valid orders. The helper resolves the order by `pk`, matching the URL pattern. This also fixes the same lookup in `RefundViewSet`.

### `app/eventyay/base/models/orders.py`

**Roll back payment state when `_mark_paid()` raises `QuotaExceededException`**

`OrderPayment.confirm()` marks the payment confirmed in an early transaction before calling `_mark_paid()`. If quota checks fail, the payment remained confirmed in the database while the order stayed unpaid — notably affecting `POST .../orders/{code}/mark_paid/`, which returned HTTP 400 but left a confirmed payment behind. On quota failure, the payment's `state`, `payment_date`, and `info` are now restored to their pre-confirm values before re-raising the exception.

### `app/eventyay/api/views/product.py`

**Return `count_waitinglist` for `waiting_list` in quota availability**

`GET .../quotas/{id}/availability/` set `waiting_list` from `qa.count_pending_orders` instead of `qa.count_waitinglist`. With waiting-list entries and no pending orders, the API reported `waiting_list: 0` instead of the correct count. The endpoint now uses `QuotaAvailability(full_results=True)` (required for waiting-list statistics) and returns `qa.count_waitinglist[quota]`, consistent with the control panel and exporters.

## Test plan

- [x] **Quota availability** — `GET /api/v1/organizers/{org}/events/{event}/quotas/{id}/availability/` on a quota with waiting-list entries returns correct `waiting_list` and unchanged `pending_orders`.
- [x] **Payment create** — `POST .../orders/{order_pk}/payments/` with `"state": "confirmed"` on a quota-exceeded order returns HTTP 400 with quota error (not HTTP 201).
- [x] **Mark paid** — `POST .../orders/{code}/mark_paid/` on a quota-exceeded order returns HTTP 400; no new confirmed payment remains in the database.
- [x] **Payment/refund routes** — `POST .../orders/{order_pk}/payments/` resolves the order correctly (no spurious 404 for valid order IDs).

https://github.com/user-attachments/assets/9fcc6bd3-7456-4191-bbd6-17ce8415d485


app/eventyay/api/views/order.py
```diff
@@ -1155,6 +1155,10 @@ def perform_update(self, serializer):
         order_modified.send(sender=serializer.instance.order.event, order=serializer.instance.order)
 
 
+def get_order_for_nested_route(request, order):
+    return get_object_or_404(Order, pk=order, event=request.event)
```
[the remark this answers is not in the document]
SxxAq: Copilot is wrong. The URL routing converter is configured as <int:order>. Django will automatically return a 404 for any non-numeric string before it ever reaches the code. It is impossible to trigger a 500 error or break anything.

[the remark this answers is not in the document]
Rachit7168: Agree with @SxxAq — no change needed. Nested payment/refund routes are registered under orders/<int:order>/ in api/urls.py, so Django rejects non-numeric values with a 404 before the view runs. The previous code= lookup was the bug: passing a valid numeric pk (e.g. 27035) searched for code="27035", which does not exist. OrderViewSet actions like mark_paid still use order code via lookup_field = 'code' on a separate route — that behavior is unchanged.

Saksham-Sirohi (commented):
## Review: Quota handling in payment and availability APIs

Solid bug-fix PR — the three changes address real correctness issues rather than cosmetic cleanup.

### `get_order_for_nested_route` (order lookup)
Confirmed the nested routes use `<int:order>` in `api/urls.py`, so the previous `code=self.kwargs['order']` lookup was wrong for numeric PKs (e.g. searching for code `"27035"`). The helper is a good DRY improvement. Agree with @SxxAq and @Rachit7168 that Copilot's API-breaking concern doesn't apply here.

Minor nit: rename the parameter from `order_id` to something like `order_pk` in a follow-up — it matches the URL converter and avoids implying order *code*.

### Swallowing `QuotaExceededException` in `PaymentViewSet.create`
Raising `ValidationError({'detail': str(e)})` is the right direction — silently passing let quota failures look like success. Two follow-ups to consider:

1. **Stable error shape:** Other API endpoints often use a dedicated error code or non-localized key. If `str(e)` is translated or includes internal quota names, clients may parse inconsistently. A fixed message like `"Quota exceeded."` with an optional `code` field would be more stable.

2. **HTTP status:** `ValidationError` typically becomes 400. Some clients may expect 409 for quota conflicts — worth checking how pretix/eventyay APIs handle this elsewhere for consistency.

### Payment rollback on quota exceeded (`OrderPayment.confirm`)
The snapshot/restore of `state`, `payment_date`, and `info` before re-raising is the correct pattern when `_mark_paid` can fail after partial mutation. One thing to verify in manual QA:

- Confirm the outer `select_for_update` lock on the payment row and the inner `transaction.atomic()` rollback don't leave the *order* in a partially confirmed state (e.g. positions reserved but payment reverted). A test that asserts order state after a forced quota failure during `confirm()` would lock this in.

### Availability API: `count_waitinglist` vs `count_pending_orders`
This is clearly a bug fix — the response field was labeled `waiting_list` but returned pending orders. Using `full_results=True` is appropriate so all breakdown fields are populated. Please double-check the OpenAPI/schema docs if this endpoint is documented, since the returned `waiting_list` value will change for existing consumers.

**Verdict:** Approve directionally once quota-failure rollback is covered by a test (or confirmed by existing test suite). Nice targeted fix.

app/eventyay/api/views/order.py
```diff
@@ -1113,8 +1113,8 @@ def perform_destroy(self, instance):
             ocm.commit()
         except OrderError as e:
             raise ValidationError(str(e))
-        except Quota.QuotaExceededException as e:
-            raise ValidationError(str(e))
+        except Quota.QuotaExceededException:
+            raise ValidationError({'error_code': 'QUOTA_EXCEEDED', 'detail': 'Quota exceeded.'})
```
sarafarajnasardi: Since you've introduced QuotaExceededAPIException which nicely sets the HTTP status to 409 (Conflict), let's use it here as well for consistency across the API, rather than manually passing the dict into a ValidationError (which will return a 400 Bad Request)
suggestion
```
        except Quota.QuotaExceededException:
            raise QuotaExceededAPIException()
```

sarafarajnasardi (changes requested):

app/eventyay/api/views/order.py
```diff
@@ -1536,6 +1540,12 @@ class RetryException(APIException):
     default_code = 'retry_later'
 
 
+class QuotaExceededAPIException(APIException):
+    status_code = 409
+    default_detail = 'Quota exceeded.'
+    default_code = 'QUOTA_EXCEEDED'
+
```
ArnavBallinCode, line 1547: This changes the response from the original exception message (str(e)) to a fixed "Quota exceeded." I this intentional?

Rachit7168: No , a mistake

app/eventyay/api/views/order.py
Rachit7168: No need , will change the whole error message

SxxAq (approved):
LGTM

prayag78 (approved):

ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):

PR #3953, opened 2026-06-14
Fix HLS playback termination when schedule ends

### LINKED ISSUE #3920: HLS playback continues after the active stream schedule ends
Description

When an active HLS stream schedule ends and the room has no default HLS URL, Eventyay displays the offline screen but does not stop the existing HLS playback. Audio continues playing, and the browser may continue requesting HLS segments after the stream is considered offline.

Steps to Reproduce

  1. Create a stage room using the livestream.native module.
  2. Leave the room’s default HLS URL empty.
  3. Create an active HLS schedule using a working live HLS URL.
  4. Enter the room and confirm the stream is playing.
  5. Wait for the schedule to end or delete the active schedule.
  6. Observe the offline screen and audio playback.

  Actual Behavior

  - The offline screen appears.
  - The current-stream API returns 404, correctly indicating no active schedule.
  - Audio from the previous HLS stream continues playing.
  - HLS segment requests may continue in the browser Network panel.

  Expected Behavior

  - Audio and video stop immediately when the schedule ends.
  - The hls.js player is destroyed.
  - The <video> source is cleared.
  - HLS manifest and segment requests stop.
  - The offline screen is displayed.

sarafarajnasardi:
Fixes #3920 

## Summary

Fix HLS playback continuing after the active stream URL becomes empty or changes.

  ## Changes

  - Destroy the existing hls.js player before evaluating the next URL.
  - Pause and unload the video element when playback ends.
  - Cancel retry timers and native HLS metadata handlers from the previous source.
  - Reset stale playback and quality state.
  - Preserve the user's play/pause intent during player teardown.
  
  

https://github.com/user-attachments/assets/440f6117-f6b3-4c8d-8776-a64d7aff2e5a

### DIFF
--- a/app/eventyay/webapp/video/src/components/Livestream.vue
+++ b/app/eventyay/webapp/video/src/components/Livestream.vue
@@ -371,14 +371,14 @@ export default {
 			if (this.$refs.video.paused) {
 				this.playbackRequested = true
 				this.$refs.video.play()
-				this.player.startLoad()
+				this.player?.startLoad()
 				// force live edge after unpausing
 				// TODO make this less yarring
 				this.$refs.video.currentTime = this.$refs.video.buffered.end(this.$refs.video.buffered.length - 1)
 			} else {
 				this.playbackRequested = false
 				this.$refs.video.pause()
-				this.player.stopLoad()
+				this.player?.stopLoad()
 			}
 		},
 		chooseLevel(level) {
--- a/app/eventyay/webapp/video/src/components/Livestream.vue
+++ b/app/eventyay/webapp/video/src/components/Livestream.vue
@@ -65,6 +65,22 @@ const HLS_DEFAULT_CONFIG = {
 	manifestLoadingMaxRetryTimeout: 5000
 }
 
+const createPlayerState = (buffering = true) => ({
+	isLive: null,
+	buffering,
+	seeking: false,
+	currentTime: 0,
+	duration: 0,
+	bufferedRanges: null,
+	hoveredProgress: null,
+	levels: null,
+	autoLevel: null,
+	manualLevel: null,
+	showLevelChooser: false,
+	textTracks: [],
+	showCaptionsChooser: false
+})
+
 export default {
 	components: {},
 	props: {
@@ -88,28 +104,14 @@ export default {
 	data() {
 		return {
 			theme,
-			isLive: null,
 			playing: true,
 			playbackRequested: true,
-			buffering: true,
-			seeking: false,
 			offline: false,
 			fullscreen: false,
 			volume: 1,
 			muted: false,
 			automuted: false,
-			currentTime: 0,
-			duration: 0,
-			bufferedRanges: null,
-			hoveredProgress: null,
-			// Quality levels
-			levels: null,
-			autoLevel: null,
-			manualLevel: null,
-			showLevelChooser: false,
-			// Captions
-			textTracks: [],
-			showCaptionsChooser: false,
+			...createPlayerState(),
 			// Alternative sources
 			showSourceChooser: false,
 			chosenAlternative: null,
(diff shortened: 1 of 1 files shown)


Rachit7168 (changes requested):
Please provide a Screen cast of the changes

ViRUS-0-0 (changes requested):
@sarafarajnasardi Please Resolve the conflicts and also update the PR description with screen recording/screenshots

Saksham-Sirohi (commented):
## Review: HLS playback termination when schedule ends

This is a well-structured cleanup of the Livestream player lifecycle. The root cause — `onEnded` destroying the player while `playing` stayed `true`, so a re-init would immediately call `video.play()` — is correctly addressed by tracking `playbackRequested` separately from UI `playing` state.

### What works well
- **`destroyPlayer()`** centralizes teardown: clears retry timers, removes the native `loadedmetadata` listener, pauses/resets the `<video>` element, and nulls player state. This fixes listener leaks that would accumulate across schedule transitions.
- **`scheduleRetry(player, callback, delay)`** guards callbacks with `this.player === player`, preventing stale retries after re-init — a common HLS.js footgun.
- Capturing `shouldPlay` before `destroyPlayer()` in `initializePlayer()` avoids the race where `this.playing` is stale during async setup.
- Native Safari path now uses `{once: true}` and stores the handler for removal — good.

### Issues / suggestions

1. **`togglePlay` null-safety:** After `destroyPlayer()`, `this.player` is `null`. If the user toggles play while offline (or between schedule items), `this.player.startLoad()` / `this.player.stopLoad()` will throw. Consider:
   ```js
   if (this.$refs.video.paused) {
     this.playbackRequested = true
     this.$refs.video.play()
     this.player?.startLoad()
     ...
   } else {
     this.playbackRequested = false
     this.$refs.video.pause()
     this.player?.stopLoad()
   }
   ```

2. **Missing screencast / test plan:** @Rachit7168 and @ViRUS-0-0 asked for a screen recording — still needed. Specifically: (a) schedule item ends → stream stops and stays stopped, (b) user manually pauses → next item does not auto-play, (c) user resumes → playback continues, (d) network error retry still works.

3. **Branch conflicts:** Please rebase on `dev` before merge since this touches hot-path video code.

### Verdict
Logic looks correct for the reported bug. Requesting the `togglePlay` guard and a short screen recording before approval.

Saksham-Sirohi (commented):

Withdrawing the `togglePlay` / `this.player?.startLoad()` suggestion. The control path is `toggleVideo()`, which already bails out with `if (this.offline) return`, and play/pause controls are not shown when `offline` is true. So a null `this.player` during user interaction is not a realistic path.

The `playbackRequested` / `destroyPlayer()` / `scheduleRetry()` changes remain the right fix for schedule-end auto-replay. Screencast for schedule transitions is still the main thing I'd want before merge.

app/eventyay/webapp/video/src/components/Livestream.vue
Aqil-Ahmad, line 1: the file still has ` this.player?.destroy()`. check if those need to be replaced with `this.destroyPlayer()`

app/eventyay/webapp/video/src/components/Livestream.vue
```diff
@@ -201,27 +205,71 @@ export default {
...
+			const video = this.$refs.video
+			if (video) {
+				if (this.nativeMetadataHandler) {
+					video.removeEventListener('loadedmetadata', this.nativeMetadataHandler)
+				}
+				video.pause()
+				video.removeAttribute('src')
+				video.load()
+			}
+			this.nativeMetadataHandler = null
```
Aqil-Ahmad, line 235: can we extract all these into a factory function and reuse it? i think that would be nice

Aqil-Ahmad (changes requested):
Also please add screencast

Aqil-Ahmad:
tested the audio termination, works now.

please fix the following issues maybe in a follow-up:

1. when the session time and stream time are non-overlapping the following error is shown. its correct but clicking save shows success. Also the modal height increases but there is no scroll and the top part of modal goes out of window.

image

2. the time/date selection shows local time when opened but the times are in UTC which makes it confusing.

Aqil-Ahmad (approved):

sarafarajnasardi:
> tested the audio termination, works now.
> 
> please fix the following issues maybe in a follow-up:
> 
> 1. when the session time and stream time are non-overlapping the following error is shown. its correct but clicking save shows success. Also the modal height increases but there is no scroll and the top part of modal goes out of window.
> 
> image
> 2. the time/date selection shows local time when opened but the times are in UTC which makes it confusing.

These might be out of scope for this issue and can be addressed in a follow-up.

ArnavBallinCode (approved):

SxxAq (approved):
LGTM

Rachit7168 (approved):
LGTM , TESTED

PR #3911, opened 2026-06-13
chore: schedule movement bug

Saksham-Sirohi:
Screenshot 2026-06-02 at 6 22 25 PM
Schedule when moved horizontally breaks the session placement ability fixed here

PR #3954, opened 2026-06-14
feat(ci): add PR process reminder checklist workflow

Saksham-Sirohi:
Post and refresh a contributor checklist on open PRs to dev/main, including auto-detected media, AI review status, unresolved feedback, and updates when reviews or review threads change.


Saksham-Sirohi:
@mariobehling A sample comment can be found here: https://github.com/fossasia/eventyay/pull/3911#issuecomment-4702569880. This shall eliminate the need for repeated Copilot comment reminders and Copilot resolution screenshot addition requests, and so on

Saksham-Sirohi:
Thoughts to collect:
1. Should Sorcery-AI be considered for AI reviews
2. Should the workflow fail for pr in case there are unresolved comments or screenshots, etc not attached
3. Also, in case we do consider it in accepted AI reviews, do we fail it if it ran into limits and the user has not requested any other reviews



Once I am clear on this, we can maybe proceed and integrate this, and later we can add it to other repositories' plugins, etc and so on

mariobehling:
Very good idea. Do we have some examples how other projects outside of FOSSASIA are handling these? Need to discuss this in a meeting a bit more as well.

mariobehling (approved):

PR #3961, opened 2026-06-15
Separate stage playback mode from stream source

### LINKED ISSUE #3959: Separate stage playback mode from stream source
## Problem

  Stage configuration currently mixes two different concepts:

  - how a stage gets its playback source
  - what type of stream source is configured

  This makes the UI misleading and allows invalid combinations, such as configuring stream schedules for an always-on stage.

  ## Goal

  Separate stage playback mode from stream source and prevent mixed configuration at the UI level.

  Stage playback mode:

  - `schedule_driven`: playback comes from StreamSchedule entries only.
  - `always_on`: playback comes from the room’s default stream source only.

  Stream source type:

  - HLS
  - YouTube
  - Vimeo for StreamSchedule entries
  - Iframe

  ## Expected Behavior

  ### Schedule-driven stage

  - Can be created without any base stream URL.
  - Shows StreamSchedule management in admin.
  - Does not show or require default room stream source controls.
  - If no active schedule exists, the room shows an offline/no scheduled stream state.

  ### Always-on stage

  - Requires/configures a default room stream source.
  - Does not show StreamSchedule management in the normal admin flow.
  - Playback uses the room default source.
  - Existing stream schedules should not be part of the always-on stage workflow.

  ## UI Requirements

  - Show playback mode as its own choice.
  - Show default stream source controls only for always-on stages.
  - Show StreamSchedule controls only for schedule-driven stages.
  - Do not expose legacy `native` as a user-facing stream type.

  ## YouTube Regression Constraints

  Do not break existing YouTube playback behavior:

  - iframe URL must include `enablejsapi=1`
  - iframe URL must include `origin`
  - iframe permissions must preserve fullscreen and autoplay behavior
  - privacy-enhanced mode must keep using `youtube-nocookie.com`
  - loop must keep setting `playlist=<video id>`
  - translation mute/unmute behavior must remain intact

sarafarajnasardi:
Fix #3959 
### Overview

Previously, creating a stage required a base HLS URL — even for schedule-driven stages that should never have a default stream URL. This PR decouples **playback mode** from **stream source configuration**, letting each stage type behave correctly.

---

https://github.com/user-attachments/assets/38ea433d-0d8a-4871-a113-98f7158410b9

## Summary by Sourcery

Update livestream source selection to respect scheduled stream type and remove support for the broken native stream option.

Bug Fixes:
- Ensure HLS livestreams are used when the active room schedule specifies HLS, even if the room module is configured as another stream type.

Enhancements:
- Fall back to the room module type only when no explicit stream type is set on the current stream schedule.

Chores:
- Remove the unused native stream type from backend schedule model choices and the admin stream schedule UI options.


Rachit7168 (changes requested):
Address codex comments

Aqil-Ahmad:
@sarafarajnasardi fix the failing tests so its ready for reviews

app/eventyay/base/models/stream_schedule.py
```diff
@@ -68,6 +67,10 @@ def __str__(self):
             return f'{self.room.name} - {self.title}'
         return f'{self.room.name} - {self.start_time}'
 
+    def __init__(self, *args, **kwargs):
+        self._skip_validation = kwargs.pop('skip_validation', False)
+        super().__init__(*args, **kwargs)
+
```
Aqil-Ahmad: its only added for enabling a test. the test could instead bypass validation with `QuerySet.update()` or a raw insert

app/eventyay/webapp/video/src/components/RoomEditPrompt.vue
```diff
@@ -46,7 +46,8 @@ prompt.c-room-edit-prompt(:scrollable="false", @close="$emit('close')")
 					v-if="inferredType && typeComponents[inferredType.id]",
 					:is="typeComponents[inferredType.id]",
 					:config="config",
-					:modules="modules"
+					:modules="modules",
+					:creating="!wasConfigured"
 				)
```
Aqil-Ahmad, line 52: The sidebar-addons used to live inside `stage.vue`, so both room editors rendered them automatically. This PR moved `sidebar-addons` out of `stage.vue` and only re-added it to the inline editor (`EditForm`). This modal embeds the stage editor here but never renders `sidebar-addons`, so opening a stage via the "Edit" button (`item.vue`) no longer lets you enable Chat, Q&A, Polls, or configure the chat webhook. The inline form still has them, so it's an inline/modal inconsistency rather than a total loss. Was this intended? If not, please render `sidebar-addons` here for stages too.

Image

Aqil-Ahmad (changes requested):
I used this url to test the stream https://test-streams.mux.dev/x36xhzz/x36xhzz.m3u8 , but as soon as the stream is about to start it goes offline.

sarafarajnasardi:
> I used this url to test the stream https://test-streams.mux.dev/x36xhzz/x36xhzz.m3u8 , but as soon as the stream is about to start it goes offline.

this is recorded hls stream...we need live hls stream...

app/eventyay/webapp/video/src/components/RoomEditPrompt.vue
sarafarajnasardi: thanks for review....i have fixed this...

Aqil-Ahmad:
> this is recorded hls stream...we need live hls stream...

thanks, it works. one thing I noticed: when the stream goes offline after the end time passes, the audio keeps playing and only stops after a page reload. could this be related to this PR?

sarafarajnasardi:
> > this is recorded hls stream...we need live hls stream...
> 
> thanks, it works. one thing I noticed: when the stream goes offline after the end time passes, the audio keeps playing and only stops after a page reload. could this be related to this PR?

i have fixed this in other pr...

Rachit7168 (changes requested):
Image

On dev there is already migration file 31 , Please change that !

Aqil-Ahmad (changes requested):
maybe fix this if its not planned otherwise. also rename and relink the migration.

 i dont see any other issues.

Image

Saksham-Sirohi (commented):
## Review: Separate stage playback mode from stream source

Important architectural improvement — decoupling *when* a stage plays (schedule-driven vs always-on) from *what* plays (HLS/YouTube/iframe) fixes a real product limitation and sets up cleaner room creation.

### What looks solid
- `playback_mode` in livestream module config is the right abstraction; schedule-driven stages no longer require baking an HLS URL into `livestream.native`.
- `create_room` validation ensures exactly one livestream module + optional chat, with permission checks per module type.
- Data migration `native` → `hls` for `StreamSchedule.stream_type` is necessary cleanup.
- Auto-deleting stream schedules when switching away from `schedule_driven` prevents orphaned schedules from driving playback unexpectedly.
- `CreateStagePrompt` UX (radio for mode + conditional source fields) matches the backend model.

### Issues to address before merge

1. **Migration number collision** — this PR adds `0032_migrate_native_stream_schedules.py` while #3909 also adds `0032_team_force_hide_speaker_emails_and_more.py`. Only one can land on `dev` first; the other must renumber. Please coordinate with @Rachit7168.

2. **HLS test failure from @Aqil-Ahmad** — using `https://test-streams.mux.dev/x36xhzz/x36xhzz.m3u8`, playback stops immediately. This may be related to `shouldUseLivestream` / schedule-driven gating rather than the migration itself. Needs reproduction steps and fix before merge since this is core video functionality.

3. **`buildLivestreamModule` for schedule-driven** — when `playbackMode !== ALWAYS_ON`, the code always returns `type: 'livestream.native'` regardless of selected source. Confirm schedule-driven stages intentionally only support HLS native playback (YouTube/iframe schedule items still go through `StreamSchedule` entries).

4. **`shouldUseLivestream` / live edge** — per Sourcery/Codex feedback, extract the repeated `'hls'` string checks and verify schedule-driven + always-on transitions don't leave the player in a half-initialized state (related to #3953).

5. **Screen recording** — given the number of playback mode combinations, a screencast covering: create always-on HLS stage, create schedule-driven stage, add/remove stream schedule entries, switch modes in room settings, would help reviewers enormously.

### Tests
Please confirm automated coverage for:
- `uses_schedule_driven_stage()` true/false paths
- schedule deletion on module_config patch
- migration reversibility (noop reverse is fine if documented)

**Verdict:** Good design, but migration conflict + reported HLS regression need resolution first.

Saksham-Sirohi (commented):

Please ignore the migration-number collision note in my previous comment. `dev` is currently at `0031`; `0032` is unoccupied, so this PR's migration filename is fine.

The actionable items that still matter: the reported HLS regression from @Aqil-Ahmad, and a screencast covering schedule-driven vs always-on transitions.

Rachit7168 (changes requested):
Thanks for working on this 
The tests are failing. Please address

app/eventyay/webapp/video/src/views/admin/rooms/StreamSchedule.vue
```diff
@@ -126,6 +137,10 @@ export default {
 				url: {
 					required: required('Stream URL is required'),
 					url: url('Must be a valid URL'),
```
Rachit7168: In CreateStagePrompt.vue, the youtubeid validator allows users to enter either a full YouTube URL or just the Video ID (e.g., dQw4w9WgXcQ), which is very convenient since the normalizeYoutubeVideoId helper supports both. However, here in StreamSchedule.vue, we unconditionally enforce Vuelidate's url('Must be a valid URL') rule for all stream types.

This means if a user inputs just a raw YouTube Video ID here, it will fail the url check (even though our custom youtube validation rule would have accepted it). Consider conditionally applying the url rule so that it matches the flexibility of CreateStagePrompt.vue.

app/eventyay/webapp/video/src/views/admin/rooms/types-edit/stage.vue
```diff
@@ -218,6 +313,32 @@ export default defineComponent({
 })
 </script>
 <style lang="stylus">
```
Rachit7168, line 315: The .radio-options, .radio-option, .radio-title, and .radio-description stylus rules added here are practically identical to the ones in CreateStagePrompt.vue. This is completely fine for now, but we might want to abstract this out into a shared component or a global UI class in the future if we reuse this "radio card" pattern for other settings!

Rachit7168 (changes requested):
Please address

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/320dedf9-b63a-4e5e-b963-6c6b08208e3c

Aqil-Ahmad (approved):

Saksham-Sirohi (approved):

PR #3963, opened 2026-06-15
fix(talk): restore schedule public/non-public toggle and presale nav

Saksham-Sirohi:
…sync

Re-enable the schedule-public API, persist show_schedule reliably, and gate presale navigation on schedule visibility so hiding the schedule works again.


Rachit7168 (approved):
LGTM , Tested works 

https://github.com/user-attachments/assets/621b880c-91e0-4c15-817d-1ad35371f594

PR #3972, opened 2026-06-16
Show followed organizer upcoming events on startpage

prayag78:
Show followed organizer upcoming events on startpage

Screenshot 2026-06-16 205345


Rachit7168 (approved):
LGTM , Tested : 

https://github.com/user-attachments/assets/fabc4490-1833-46c0-89fc-23a3f35d9802

sarafarajnasardi (approved):

Rachit7168 (changes requested):
Thanks for working on this : Please change 
The same event is featured and also followed : It should be shown under Ideally, both

prayag78:
> Thanks for working on this : Please change The same event is featured and also followed : It should be shown under Ideally, both


https://github.com/user-attachments/assets/98b6258b-1354-47aa-9c8a-54ec2b36f48f

With the current changes it will show in both sections

Rachit7168 (changes requested):
https://github.com/user-attachments/assets/382e9e32-a859-40dc-98c1-bbac5f5e52f8

Please recheck once , the thing you mentioned is not working !

app/eventyay/presale/views/startpage.py
```diff
@@ -124,4 +124,23 @@ def get_context_data(self, **kwargs):
...
+                        organizer_id__in=followed_org_ids,
+                        live=True,
+                    )
+                    .filter(Q(startpage_visible=True) | Q(startpage_featured=True))
+                    .filter(Q(date_to__gte=today) | Q(date_to__isnull=True, date_from__gte=today))
+                    .select_related('organizer')
+                    .order_by('date_from')[:20]
+                )
+                followed_upcoming_events = [e for e in qs if not e.has_component_testmode]
+
```
ViRUS-0-0, line 145: Hey! The current code fetches the events but doesn't fetch their settings at the same time. Because the very next line checks `e.has_component_testmode` (which requires those settings), Django is forced to make a separate database query for *every single event* in the list just to get its settings. This creates an "N+1 query" performance issue that will slow down the page load. Adding `.prefetch_related('_settings_objects')` to your query fixes this by grabbing all the settings at once in a single database trip!

app/eventyay/presale/templates/pretixpresale/startpage.html
ViRUS-0-0, line 1: If I have followed an organizer, its upcoming event is visible to me two times, one in upcoming events and another in Upcoming events from followed organizers. Do we want this behaviour??

Image

ViRUS-0-0 (changes requested):
Changes look good. PTAL at the comments left.

app/eventyay/presale/templates/pretixpresale/startpage.html
prayag78: still in discussion.

app/eventyay/presale/views/startpage.py
prayag78: fixed

Saksham-Sirohi (approved):

PR #3979, opened 2026-06-17
fix(cfp): enhance PDF upload size validation and UI

Rachit7168:
### What is changed
- Added file size validation to `SlidesField` in `resource.py` so a user-friendly error is returned on the form level.
- Updated `cfp/flow.py` to preserve uploaded files in session storage when a form fails validation due to missing fields, so users do not have to repeatedly upload large files.
- Added real-time JavaScript validation in `slides_input.html` to instantly display an error when a user selects a file larger than `10MB`, and clear the file selection automatically.
- Restyled the `SlidesWidget` template to correctly hide the duplicate native `Choose file` input element while retaining full accessibility.
- Greatly simplified the previously verbose help text displayed in the widget (e.g. removed repetitive strings).

### Reason
When users attempted to upload a file slightly above `10MB` on the live server, they encountered an abrupt Nginx "Entity Too Large" error. Additionally, if they uploaded a valid PDF but left a required field blank, the form would re-render and lose their uploaded file, which was very frustrating. This PR ensures size limitations are validated instantly on the client side, safely handled gracefully on the server side, and keeps successfully uploaded files preserved across invalid form submissions.

NOTE : File is being stored in Server's Local files inside the MEDIA_ROOT/cfp_uploads

### Before 

https://github.com/user-attachments/assets/064809a0-ef10-4dc3-ab42-36ffca5f2200

### After

https://github.com/user-attachments/assets/e192b784-0512-44da-b504-aee655aac4ec


ArnavBallinCode (approved):
LGTM

app/eventyay/common/templates/common/widgets/slides_input.html
```diff
@@ -43,28 +43,141 @@
...
+                    byteNumbers[i] = slice.charCodeAt(i);
+                }
+                var byteArray = new Uint8Array(byteNumbers);
+                byteArrays.push(byteArray);
+            }
+            var blob = new Blob(byteArrays, {type: contentType});
+            return new File([blob], filename, {type: contentType});
+        }
+
+        if (input && label) {
```
SxxAq: The uploaded file stored in `sessionStorage` is never cleared on a successful form submission. If a user creates proposal A (uploading `slides.pdf`), submits it successfully, and then goes to create proposal B in the same browser tab, the old `slides.pdf` will automatically be restored from `sessionStorage` and populated into the new form, leading to unintended file leakage/uploads.

To fix this, we should only restore files if the page was re-rendered due to form validation errors. If there are no errors, we should clear the `sessionStorage` key.

app/eventyay/common/templates/common/widgets/slides_input.html
```diff
@@ -43,28 +43,141 @@
...
+                      var base64 = e.target.result.split(',')[1];
+                      filesData.push({
+                          name: file.name,
+                          type: file.type,
+                          base64: base64
+                      });
+                      processed++;
+                      if (processed === filesToProcess) {
+                          try {
+                              sessionStorage.setItem(storageKey, JSON.stringify(filesData));
```
SxxAq: Note that browser `sessionStorage` has a strict storage quota (usually ~5MB). Since Base64 encoding adds around 33% size overhead, any PDF file close to or larger than 3.7MB will trigger a `QuotaExceededError` here. 

Although catching it and logging to the console prevents a page crash, it means file preservation will silently fail for larger PDF slides. The user won't get any visual indication of this failure until they submit and see the file input cleared. 

Since this is a browser-enforced storage limit, we might want to document this limitation or eventually transition to a temporary database-based draft system or async file-upload bucket if we want it to work consistently for large files.

app/eventyay/common/templates/common/widgets/slides_input.html
```diff
@@ -43,28 +43,164 @@
...
+        {% endif %}
+      </div>
+      <div id='{{ widget.files_id }}_storage_warning' class='text-warning mt-1' style='display: none; font-size: 0.9em; font-weight: bold;'>
+        {% translate 'Note: This file is too large to be saved in the browser’s temporary storage. If the form has other errors, you will need to re-select it before submitting.' %}
+      </div>
+    </div>
+    <script>
+      (function() {
+        var input = document.getElementById('{{ widget.files_id }}');
+        var label = document.getElementById('{{ widget.files_id }}_label');
```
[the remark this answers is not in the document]
Rachit7168: Moved the inline script logic to a new external ES module file (app/eventyay/static/common/js/slides_input.js) and loaded it via <script src="{% static 'common/js/slides_input.js' %}" defer></script> in slides_input.html. Passed necessary template variables using data- attributes on the wrapper div.

app/eventyay/common/templates/common/widgets/slides_input.html
```diff
@@ -43,28 +43,164 @@
...
+
+          input.addEventListener('change', function() {
+            if (input.files && input.files.length > 0) {
+              var maxSize = parseInt(input.getAttribute('data-maxsize'), 10);
+              var totalSize = 0;
+              for (var i = 0; i < input.files.length; i++) {
+                totalSize += input.files[i].size;
+              }
+              
+              if (maxSize && totalSize > maxSize) {
```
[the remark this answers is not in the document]
Rachit7168: Updated the client-side JavaScript validation logic to evaluate the maxSize condition against each individual file instead of accumulating the sizes together.

app/eventyay/common/templates/common/widgets/slides_input.html
```diff
@@ -43,28 +43,164 @@
...
+        {% translate 'Choose file' %}
+      </label>
+      <input
+        type='file'
+        name='{{ widget.files_name }}'
+        id='{{ widget.files_id }}'
+        style='display: none !important;'
+        accept='.pdf,application/pdf'
+        {% if widget.max_size %}data-maxsize='{{ widget.max_size }}'{% endif %}
+        {% if widget.max_items != 1 %}multiple{% endif %}>
```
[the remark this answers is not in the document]
Rachit7168: Replaced style='display: none !important;' with class='sr-only sr-only-focusable' on the file <input> to ensure it remains accessible to screen readers and keyboard navigation.

hpdang:
@ArnavBallinCode @SxxAq please review this again

SxxAq (approved):
LGTM

Saksham-Sirohi (commented):
## Review: PDF upload size validation and UI

Practical CfP UX improvement. Server-side validation is the source of truth; client-side checks improve feedback before submit.

### What works
- `SlidesWidget` now passes `max_size` to the template and JS — users see limits before upload.
- Distinguishing `existing_resources` dict (DB initial) vs raw POST dict (re-render after validation error) fixes a real bug where re-rendered forms lost file context.
- Custom file picker label + `sr-only` input is more polished than the native file input.
- SessionStorage restore for re-renders is thoughtful, with an honest warning when files are too large for browser storage.

### Concerns / suggestions

1. **Removed "Previously uploaded files" banner** — the alert in `submission_base.html` was removed entirely. Users editing an existing submission may no longer see which slides are already on file unless they scroll to the widget's existing-resources section. Confirm this is intentional; if so, ensure the widget always shows existing files prominently.

2. **SessionStorage base64 storage** — storing file bytes in `sessionStorage` works for small PDFs but can hit quota (~5MB) quickly. The warning message is good; consider also skipping storage above e.g. 2MB proactively rather than failing silently on `setItem`.

3. **`slides_input.js` style** — uses `var` and IIFE on `DOMContentLoaded`. Project JS guidelines prefer ES modules. Not blocking if this matches other widget scripts, but worth aligning in a follow-up.

4. **Server-side validation test** — please confirm there's a test asserting oversize PDFs are rejected at the form/model layer, not only client-side. Client checks are bypassable.

5. **Help text regression** — the per-field "You can add N more slide entries" message was removed. The new copy is simpler but less informative when users already have partial uploads.

**Verdict:** LGTM with confirmation that server-side size validation is tested and the removed upload banner was deliberate.

Saksham-Sirohi (commented):

Server-side PDF size enforcement is already wired: `InfoForm` sets `max_size=pdf_max_size` on the slides field via `SizeFileInput` / `ExtFileField`, not only client JS. Withdraw the implication that validation might be client-only.

Still worth checking: intentional removal of the "Previously uploaded files" banner in `submission_base.html` — users editing existing submissions now rely entirely on the widget's existing-resources list.

Comment only.

app/eventyay/common/templates/common/widgets/slides_input.html
```diff
@@ -42,29 +42,46 @@
...
+      </label>
+      <input
+        type='file'
+        name='{{ widget.files_name }}'
+        id='{{ widget.files_id }}'
+        class='sr-only sr-only-focusable'
+        accept='.pdf,application/pdf'
+        {% if widget.max_size %}data-maxsize='{{ widget.max_size }}'{% endif %}
+        {% if widget.max_items != 1 %}multiple{% endif %}>
+      <span id='{{ widget.files_id }}_label' class='ml-2 text-muted' style='font-size:0.95em; overflow:hidden; text-overflow:ellipsis; white-space:nowrap;'>
```
prayag78: Avoid inline styles, rest LGTM

prayag78 (changes requested):

app/eventyay/common/templates/common/widgets/slides_input.html
Rachit7168: Fixed

prayag78 (approved):

SxxAq (approved):
LGTM

Saksham-Sirohi (approved):

PR #3993, opened 2026-06-19
fix(checkin): avoid crash on variation ticket require_attention

Saksham-Sirohi:
## Summary

- Fix `OrderPosition.require_checkin_attention` crashing on variation tickets because `ProductVariation` has no `checkin_attention` field.
- Use safe `getattr` for variation attention and return explicit `False` when there is no variation (fixes plain tickets returning `None` from the old `(variation_id and ...)` term).
- Align check-in list redeem responses and device sync serializer with the centralized property so `require_attention` is consistent everywhere.

## Summary by Sourcery

Ensure check-in attention flags are computed centrally on order positions and used consistently across check-in APIs and serializers to prevent crashes on variation tickets.

Bug Fixes:
- Prevent AttributeError crashes when computing require_checkin_attention for positions with product variations.
- Ensure plain tickets return a boolean require_attention flag instead of None when no attention is required.

Enhancements:
- Route all check-in attention checks through the OrderPosition.require_checkin_attention property for consistent behavior across redeem responses and device sync serializers.

### DIFF
--- a/app/eventyay/api/serializers/order.py
+++ b/app/eventyay/api/serializers/order.py
@@ -537,11 +537,6 @@ class OrderPositionSerializer(I18nAwareModelSerializer):
         return instance
 
 
-class RequireAttentionField(serializers.Field):
-    def to_representation(self, instance: OrderPosition):
-        return instance.require_checkin_attention
-
-
 class AttendeeNameField(serializers.Field):
     def to_representation(self, instance: OrderPosition):
         an = instance.attendee_name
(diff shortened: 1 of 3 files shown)


Rachit7168 (approved):
LGTM , Tested

ArnavBallinCode:
> Tested

 were you able to work this on local?

ArnavBallinCode (approved):
lets test on dev

Rachit7168:
> > Tested
> 
> were you able to work this on local?

Made a test case file for the testing

PR #3998, opened 2026-06-20
feat(deps): replace vat-moss-forked with vat-moss-lite

### LINKED ISSUE #3997: Replace `vat-moss-forked` dependency with `vat-moss-lite`
Our Eventyay is depending on [`vat-moss-forked`](https://pypi.org/project/vat-moss-forked/) but this library's development has been stale (there is only [one release](https://pypi.org/project/vat-moss-forked/#history) on PyPI). It is also not compatible with Python 3.13+, due to its usage of now-removed `cgi` module, blocking us from being upgradable to Python 3.13+.

I forked this library to [`vat-moss-lite`](https://pypi.org/project/vat-moss-lite/) and modernized it:
- Replaced the use of `cgi`.
- Added type annotation
- Applied modern project layout.

And added [CI](https://github.com/hongquan/vat-moss-lite/actions) to test it against Python 3.12, 3.13, 3.14.

Eventyay can use `vat-moss-lite` instead of `vat-moss-forked` now. The API is the same, just need to change the `import vat_moss` to `import vat_moss_lite`.

Rachit7168:
Closes :  #3997

- Switched from vat-moss-forked to vat-moss-lite, unblocking Python 3.13+ upgrades
- Removed Python <3.13 version constraint in pyproject.toml
- Updated all references of `vat_moss` to `vat_moss_lite` across app and tests
- Updated uv.lock file


app/pyproject.toml
```diff
@@ -202,8 +202,8 @@ dependencies = [
     "tzdata>=2026.2",
     "urllib3>=2.7.0",
     "urlman>=2.0.3",
-    # This library prevents us from being compatible with Python v3.13+.
-    "vat-moss-forked>=2020.3.20.0.11.0",
+    # vat-moss-forked was replaced by vat-moss-lite which supports Python 3.13+.
```
hongquan: Who care about the old library, why have to leave a comment mentioning it?

hongquan (changes requested):

app/pyproject.toml
Rachit7168: Seems valid , removed !

MukundC25 (commented):
@Rachit7168, please resolve branch conflicts and address reviews

Saksham-Sirohi (commented):
## Review: Replace vat-moss-forked with vat-moss-lite

Dependency swap with straightforward import renames across forms, invoices, and order views.

### What looks good
- Mechanical `vat_moss` → `vat_moss_lite` migration in call sites with exception types preserved.
- Logger calls updated to `%s` formatting — minor cleanup.
- Addresses unmaintained `vat-moss-forked` dependency.

### Things to verify before merge
1. **API compatibility** — confirm `vat_moss_lite.id.validate()`, `exchange_rates.fetch()`, and `exchange_rates.format()` return the same shapes the invoice/order flows expect. A quick regression test against a known valid/invalid VAT ID would help.
2. **ECB rate fetch** — `fetch_ecb_rates` signal handler still catches `urllib.error.URLError`; verify vat-moss-lite raises the same on network failure.
3. **Branch conflicts** — @MukundC25 noted conflicts with `dev`; rebase before final review.
4. **Diff size** — if the PR includes more than the import swap (lock file, deleted vendored code, etc.), call that out in the PR description so reviewers know scope.

PR #4003, opened 2026-06-20
feat: Add reusable beta badge for plugins

Rachit7168:
## What Changed

*   Created a reusable `_beta_badge.html` partial template.
*   Utilized the built-in Bootstrap 3 `label label-success` classes for a standard, clean green pill UI without any custom CSS or inline styles.
*   Added a `BETA_PLUGINS` frozenset in `eventyay/base/plugins.py` to seamlessly add the beta flag to external plugins (e.g., `socialmedia`, `teamshifts`) that cannot be modified directly.
*   Added `beta = True` to the `webcheckin` plugin's metadata.
*   Added spacing between the plugin name and the badge in the plugin list template.

## Why

*   Allows users to easily identify which plugins are currently in beta testing.
*   Relying on existing Bootstrap classes keeps the codebase clean, avoids adding unnecessary or conflicting CSS, and ensures visual consistency with the rest of the application.

image

### DIFF
--- a/app/eventyay/base/plugins.py
+++ b/app/eventyay/base/plugins.py
@@ -28,6 +28,17 @@ class PluginType(Enum):
     EXPORT = 4
 
 
+# Module names of plugins that are currently in beta.
+# For plugins whose source cannot be edited (e.g. external git packages), add
+# their Python module name here to have the beta badge shown automatically.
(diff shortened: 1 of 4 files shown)


Saksham-Sirohi (approved):
LGTM. Nice and reusable for any future plugins we may create

PR #4017, opened 2026-06-21
Fix account enumeration via signup flow

### LINKED ISSUE #4016: Signup with existing email leaks account existence
I tried signing up with an email that's already registered. Here's what happens:
  
  1. The signup page shows a generic "Verify Your Email Address" message - good, no leak here
  2. But then I get redirected to the login page which says "Confirmation email sent to admin@example.com"
  
 Image
  
  That second message confirms the email exists in the system. An attacker could use this to build a list of valid accounts before attempting password attacks ( 6 character length enforcement in our case) or phishing.
  
  ### Steps to reproduce
  1. Create an account with test@example.com
  2. Log out
  3. Try to sign up again with test@example.com
  4. Notice the login page reveals the email was recognized
  
  ### Expected behavior

  Both pages should show the same generic message without echoing back the email address.

  ### Possible fixes
  - Remove or genericize the flash message on the login redirect
  - Consider setting `ACCOUNT_PREVENT_ENUMERATION = "strict"` in allauth config

dkoul:
Override allauth's email_confirmation_sent.txt template to remove the email address from the flash message. The default template reveals whether an email is registered by showing "Confirmation email sent to user@example.com" - this allows attackers to enumerate valid accounts.

The fix uses a generic message that doesn't disclose the email.

Fixes #4016

## Summary by Sourcery

Prevent account enumeration via signup email confirmation messages by updating template configuration and overriding the confirmation message template.

Bug Fixes:
- Hide specific email addresses in signup confirmation flashes to prevent revealing whether an email is registered.

Enhancements:
- Add base templates directory to Django template search paths to support custom account messaging templates.

### DIFF
--- a/app/eventyay/config/settings.py
+++ b/app/eventyay/config/settings.py
@@ -1257,6 +1257,8 @@ ACCOUNT_USER_MODEL_USERNAME_FIELD = None
 # account_email_verification_sent (not to the login page), so ACCOUNT_SIGNUP_REDIRECT_URL
 # below is only reached when the user is already verified (e.g. social auth signup).
 ACCOUNT_EMAIL_VERIFICATION = 'mandatory'
+# Prevent account enumeration by showing identical behavior for existing and new emails
+ACCOUNT_PREVENT_ENUMERATION = 'strict'
 # Prefer Jinja2 templates for django-allauth
 ACCOUNT_TEMPLATE_EXTENSION = 'jinja'
 ACCOUNT_ADAPTER = 'eventyay.eventyay_common.adapter.CustomAccountAdapter'


Rachit7168 (changes requested):
Seems Valid , 

No need to show email we can say a verification for this email has been sent please complete verification and signin using an appropriate account

ArnavBallinCode (changes requested):
Thank you for looking into this issue! 

While changing the message text is a good thought, this fix does not completely prevent enumeration. An attacker can still observe the difference in behavior: signing up with a new email shows the verification page, while signing up with an existing email redirects to the login page. 

Additionally, adding `BASE_DIR / 'base' / 'templates'` to `TEMPLATES['DIRS']` elevates all templates in the `base` app to have absolute global priority, which prevents other apps from overriding them in the future.

To fix the enumeration leak completely and correctly, we should use `django-allauth`'s built-in enumeration prevention. Please revert the current changes and add the following to `app/eventyay/config/settings.py`:

```python
# app/eventyay/config/settings.py
ACCOUNT_PREVENT_ENUMERATION = 'strict'

ArnavBallinCode:
https://docs.allauth.org/en/dev/release-notes/2022.html

Rachit7168:
@ArnavBallinCode  , seems valid Please address once

dkoul:
Fair point. Made changes and tested locally. No extra flash message when we set ACCOUNT_PREVENT_ENUMERATION = True

Rachit7168:
Thanks for addressing the feedback and switching to django-allauth's built-in enumeration prevention. The updated approach looks much better.

One small suggestion for future contributions: please try to avoid force-pushing after reviews have been submitted unless necessary, as it can make review history harder to follow and may dismiss comments tied to previous commits. Also, if AI assistance was used (e.g., Claude Opus 4.5), consider adding an explicit co-author attribution in the commit message instead of relying on the default `Claude Opus 4.5 <noreply@anthropic.com>` signature, which makes authorship less clear.

hongquan (commented):
Check my comment in https://github.com/fossasia/eventyay/issues/4016#issuecomment-4767215578

It may be a better way so that we can prevent leak but not confuse users who already have account.

Saksham-Sirohi (commented):
## Review: Account enumeration fix

Thanks for tackling #4016 — this is the right class of fix (configuration over bespoke templates).

### What looks good
- Using django-allauth's built-in enumeration prevention is cleaner than overriding `email_confirmation_sent.txt`, and aligns with the feedback from @hongquan and the thread on the issue.
- The inline comment in `settings.py` documents *why* the setting exists, which helps future readers.

### `strict` vs default — worth confirming intent
The project already sets `ACCOUNT_EMAIL_VERIFICATION = 'mandatory'`. Per [allauth's docs](https://docs.allauth.org/en/dev/account/configuration.html), with mandatory verification the default `ACCOUNT_PREVENT_ENUMERATION = True` already makes "email already taken" indistinguishable from a fresh signup (both paths send a verification email). The `"strict"` value is mainly needed when verification is optional/disabled, because it allows duplicate signups with the same email.

**Suggestion:** If local testing with `True` already fixes the flash-message leak (as noted in your comment), consider whether `"strict"` is necessary here. `"strict"` trades off UX (possible duplicate unverified accounts) for defense-in-depth. If the issue reporter only needed identical signup responses, `True` may be sufficient and lower-risk.

### Housekeeping
- The PR description still describes a template override that is no longer in the diff — please update it so reviewers and future git archaeology match the actual change.
- A one-line note in the PR or a short test comment referencing #4016 (e.g. "signup with existing email shows same page as new email") would make regression coverage explicit, even if allauth owns the behavior.

Overall this is a small, sensible fix once the `strict` choice is confirmed intentional.

dkoul:
Well, it turns out neither 'True' nor ''Strict' fixes the actual email leak issue. I have done extensive manual testing over the last two days for this flow. I apologise for the earlier statement that 'true' worked. It was either a glitch or cache. The ACCOUNT_PREVENT_ENUMERATION setting controls only the behavior not the message template. So, I think we can either keep it as it is and close this PR or make changes to the email_confirmation_sent.txt template.

Saksham-Sirohi (commented):

Nothing beyond the earlier review on this PR.

Saksham-Sirohi:
Closing this, as it needs to be added upstream in Django allauth (https://codeberg.org/allauth/django-allauth/issues/4748). Thank you for your contributions.

Issue #3763, opened 2026-06-02
Add standard font options for organiser and event page design without external font loading


mariobehling:
Organiser and event settings allow organisers to customize the visual design of organiser and event pages. Organisers can choose colors for their organiser pages, and event settings can override these design settings for individual events.

Currently, the font selection is limited. Only Open Sans is available as a font option. Other standard font options are not available.

The design settings should offer additional standard font choices, while avoiding loading fonts from external services.

<img width="1966" height="928" alt="Image" src="https://github.com/user-attachments/assets/2fc728f0-3e8b-4ff3-8d2f-d2b80f4690d5" />
<img width="1832" height="1020" alt="Image" src="https://github.com/user-attachments/assets/4df5d914-f344-47c6-91d5-48d9ceeb8035" />

## Current behavior

* Organisers can configure design settings for organiser and event pages.
* Event settings can override organiser-level design settings.
* The available font option is currently limited to Open Sans.
* Organisers cannot choose from other standard fonts.
* Font loading should not depend on external font services.

## Expected behavior

Organisers should be able to choose from several standard font options for organiser and event pages.

The available font options should use safe, commonly available system fonts or locally bundled fonts.

External font loading services should not be used.

## Font selection behavior

The font setting should be available in:

* Organiser design settings
* Event design settings, where event settings can override organiser settings

The event-level setting should continue to follow the existing override logic:

* If an event-specific font is selected, use it for that event page.
* If no event-specific font is selected, use the organiser-level font.
* If no organiser-level font is selected, use the platform default.

## Suggested font options

The final list should be confirmed with the team, but possible options include common system font stacks such as:

* Open Sans or current platform default
* Arial
* Helvetica
* Verdana
* Georgia
* Times New Roman
* Trebuchet MS
* Courier New
* System UI or default sans-serif stack

The implementation should prefer font stacks that work across operating systems and do not require external network requests.

## Implementation details

* Review the current organiser and event design settings.
* Add additional standard font options to the font selector.
* Ensure fonts are loaded locally or use system font stacks.
* Do not load fonts from external services such as Google Fonts or other third-party font CDNs.
* Ensure organiser-level font settings apply to organiser pages and related event pages.
* Ensure event-level font settings can override organiser-level font settings.
* Ensure the selected font is applied consistently across public organiser and event pages.
* Ensure the selected font does not break layout, spacing, buttons, forms, or navigation.
* Update frontend validation and backend validation for the allowed font values.
[...]

PR #4052, opened 2026-06-22
feat: Add system font options for events and organizers

SxxAq:
This PR implements the system font options feature for organizers and events, allowing administrators to customize the primary fonts (Arial, Helvetica, Georgia, Verdana, etc.) alongside dynamic/web fonts. 

To prevent any UI layout regressions, the font overrides are applied using scoped CSS variables (`--font-family` and `--font-family-title`) served via a lightweight settings-based stylesheet rather than loading compiled heavy Bootstrap styles globally.

https://github.com/user-attachments/assets/ba5e5452-d66c-43f2-bda0-6f214a561062

### Key Changes
1. **Settings & Configurations**:
   - Registered system fonts stack and choices in configurations and wired them to the primary font kwargs.
   - Implemented settings validation (`validate_primary_font`) to ensure only valid system or dynamic web fonts are saved.
2. **Forms & Inheritance**:
   - Updated event settings form choices to reflect inherited defaults from the parent organizer settings.
   - Cleans empty selections (`''`) and maps them to `None` upon saving to respect hierarchical settings fallback.
3. **Dynamic CSS & Previews**:
   - Enhanced the dynamic `event_css` view to serve scoped `--font-family` CSS custom properties on `:root`.
   - Updated the font selection option template to use inline fallback declarations (`style="font-family: ...; --font-preview: ...;"`) to ensure dynamic font previews render correctly regardless of stylesheet compilation cache states.
4. **Testing**:
   - Added comprehensive coverage in `tests/tickets/presale/test_style.py` verifying settings form defaults, inheritance, validation, SCSS compilation, and CSS variable output.

### Verification Results
All targeted style tests completed successfully:
```bash
pytest tests/tickets/presale/test_style.py
# 8 passed in 14.70s
```
- ref : #3763


Saksham-Sirohi (commented):
## Review: System font options for events and organizers

Well-executed feature — font inheritance from organizer → event mirrors how colors already work, and the test additions in `test_style.py` are thorough.

### Highlights
- `escape_font_name()` sanitizes CSS injection via font names (`;`, `{}`, newlines) — important for a user-controlled setting.
- `validate_primary_font()` wired into both event and organizer settings validation closes a gap where `validate_organizer_settings` was a no-op stub.
- Event form "Default (Inherit: …)" empty-string choice with `save()` converting `''` → `None` correctly preserves inheritance semantics.
- `fully_overridden_events` query fix (only skip CSS regen when *all* affected keys are explicitly set) is more correct than the previous `key__in` any-match logic.
- Including `primary_font` in `color_etag` / cache-bust query params prevents stale font CSS.

### Suggestions

1. **`event_css` view duplication** — font resolution logic now exists in both `presale/style.py` (SCSS compile) and `agenda/views/widget.py` (`event_css`). Consider extracting a shared `resolve_font_css(event)` helper to avoid the two paths diverging (e.g. widget gets Georgia but presale doesn't).

2. **Fallback behavior** — when `primary_font` is set to an invalid/removed custom font, `event_css` silently falls back to Open Sans. The presale CSS path may behave differently. A single fallback strategy would be safer.

3. **Widget etag cache** — `event_css` sets `Cache-Control: no-cache` but the `<link>` tag uses `?c=…&f=…` params. Confirm CDN/proxy layers respect this for embedded widgets on third-party sites.

4. **Font preview in admin** — `.font-preview-label { font-family: var(--font-preview) }` is a nice touch. Verify it works for system fonts (stack values) not just Google/custom fonts.

5. **Missing newline** at EOF of `_forms.scss` — trivial but pre-commit may flag it.

**Verdict:** Strong feature PR with good tests. The shared font-resolution helper would be my main follow-up before merge.

app/eventyay/agenda/views/widget.py
```diff
@@ -297,7 +308,27 @@ def event_css(request, organizer=None, event=None, **kwargs):
...
+    if primary_font:
+        fonts_dict = get_fonts()
+        font_family_value = None
+
+        if primary_font in SYSTEM_FONTS:
+            font_family_value = SYSTEM_FONTS[primary_font]
+        elif primary_font in fonts_dict:
+            escaped_font = escape_font_name(primary_font)
+            font_family_value = f'"{escaped_font}", {BASE_SANS_STACK}'
+            font_css = get_font_stylesheet(primary_font, fonts=fonts_dict)
```
Rachit7168: get_font_stylesheet() returns CSS that is designed to be compiled by SASS (it includes custom SASS static('...') function calls like url(static('fonts/...'))).

Because this event_css view returns pure CSS directly to the browser without compiling it with django_libsass, these static(...) pseudo-functions will be sent raw to the client browser, which will break the font loading completely.

Suggestion: Add a flag to get_font_stylesheet (e.g., for_sass=True). When called from this view with for_sass=False, use Django's built-in static template tag (from django.templatetags.static import static) to resolve the paths into standard URLs in Python before joining the CSS string.

Rachit7168 (changes requested):
Please address

app/eventyay/agenda/views/widget.py
SxxAq: fixed

ViRUS-0-0:
Screenshot 2026-06-24 at 7 56 43 PM

Why not use a dropdown to select fonts instead of radio buttons? Would be useful for future additions of fonts. @SxxAq 

cc: @Saksham-Sirohi

app/eventyay/presale/style.py
```diff
@@ -35,6 +37,25 @@
...
+    'Verdana': 'Verdana, Geneva, sans-serif',
+    'Times New Roman': '"Times New Roman", Times, Baskerville, Georgia, serif',
+    'Trebuchet MS': '"Trebuchet MS", "Lucida Grande", "Lucida Sans Unicode", "Lucida Sans", Tahoma, sans-serif',
+    'Courier New': '"Courier New", Courier, monospace',
+}
+
+SYSTEM_FONT_CHOICES = [
+    (font_key, 'System UI (Default Stack)' if font_key == 'System-UI' else font_key)
+    for font_key in SYSTEM_FONTS.keys()
+]
```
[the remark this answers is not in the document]
SxxAq: Not relevant. Converting the choices list to dictionary labels breaks Django's rendering of choice values, causing the duplicate fonts

Rachit7168:
@ViRUS-0-0 , 
I suggest the radio button is well and good, as we can show an example of the font style and also show an example of a sentence!

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/e8007794-48cb-47c9-8a0a-39f2f2b6c5cf

Aqil-Ahmad:
just a quick question: should the font also apply here:

image

or only to the public pages?

app/eventyay/presale/style.py
```diff
@@ -140,27 +187,32 @@ def regenerate_organizer_css(organizer_id: int):
...
         if organizer.settings.get('presale_widget_css_checksum', '') != checksum:
             newname = default_storage.save(fname, ContentFile(css.encode('utf-8')))
             organizer.settings.set('presale_widget_css_file', newname)
             organizer.settings.set('presale_widget_css_checksum', checksum)
 
-        non_inherited_events = set(
-            Event_SettingsStore.objects.filter(object__organizer=organizer, key__in=affected_keys).values_list(
-                'object_id', flat=True
-            )
+        fully_overridden_events = set(
```
Aqil-Ahmad, line 204: Count(key) counts DB rows not distinct keys, corrupt data with a duplicate row could inflate the count and wrongly skip CSS regeneration.

app/eventyay/common/templates/common/base_public.html
```diff
@@ -72,7 +72,11 @@
         {% endcompress %}
         {% block stylesheets %}{% endblock stylesheets %}
         {% block public_stylesheets %}{% endblock public_stylesheets %}
-        {% if current_event and current_event.visible_primary_color %}<link rel="stylesheet" type="text/css" href="{{ current_event.urls.settings_css }}?c={{ current_event.visible_primary_color|urlencode }}" />{% endif %}
+        {% if current_event %}
```
Aqil-Ahmad, line 76: current_event.settings.primary_font hits the DB if the settings proxy isn't already cached. The previous condition used only visible_primary_color (a model field, no query). pass the font value via view context instead

app/eventyay/control/templates/pretixcontrol/font_option.html
```diff
@@ -1,2 +1,19 @@
...
+    {% include "django/forms/widgets/input.html" %}
+
+    {% if widget.wrap_label %}
+    <strong>{{ widget.label.title }}</strong><br>
+    {% trans "The quick brown fox jumps over the lazy dog." context "typography" %}
+
+    {% if "sample" in widget.label.data %}
+    <br>{{ widget.label.data.sample }}
+    {% endif %}
+</label>
```
Aqil-Ahmad, line 18: The opening <label> is inside {% if widget.wrap_label %} but </label> is outside it if wrap_label is False the tag is never closed.

ViRUS-0-0 (approved):
The changes are working as tested on my local dev.

@Rachit7168 We can have TEST TEXT below the dropdown showing how it will look if the selected the font. I guess radio button also works for now

Approving the PR

Aqil-Ahmad (changes requested):
maybe resolve these comments, and also the font change in talks doesn't seem intentional.

SxxAq:
> and also the font change in talks doesn't seem intentional.

it wasnt intentional . fixed the leak and addressed comments .

Saksham-Sirohi (approved):

PR #4065, opened 2026-06-26
Fix voucher deletion for canceled orders

ArnavBallinCode:
https://github.com/user-attachments/assets/88bbd328-0eac-4a47-b0d0-9ba38db1af0c

Vouchers referenced exclusively by canceled orders were incorrectly
blocked from deletion. The old check (redeemed == 0 and no order
positions) did not account for the fact that order cancellation
decrements the redeemed counter but leaves OrderPosition rows in place.

New logic: a voucher is deletable unless at least one non-canceled
OrderPosition (active/pending/paid) references it.

### DIFF
--- a/app/eventyay/api/views/voucher.py
+++ b/app/eventyay/api/views/voucher.py
@@ -16,7 +16,7 @@ from rest_framework.filters import OrderingFilter
 from rest_framework.response import Response
 
 from eventyay.api.serializers.voucher import VoucherSerializer
-from eventyay.base.models import Order, OrderPosition, Voucher
+from eventyay.base.models import CartPosition, Voucher
 
 with scopes_disabled():
 
@@ -130,7 +130,7 @@ class VoucherViewSet(viewsets.ModelViewSet):
             auth=self.request.auth,
         )
         with transaction.atomic():
-            instance.cartposition_set.filter(addon_to__isnull=False).delete()
+            CartPosition.objects.filter(addon_to__voucher=instance).delete()
             instance.cartposition_set.all().delete()
             instance._clear_canceled_order_positions()
             super().perform_destroy(instance)
--- a/app/eventyay/api/views/voucher.py
+++ b/app/eventyay/api/views/voucher.py
@@ -16,7 +16,7 @@ from rest_framework.filters import OrderingFilter
 from rest_framework.response import Response
 
 from eventyay.api.serializers.voucher import VoucherSerializer
-from eventyay.base.models import Voucher
+from eventyay.base.models import Order, OrderPosition, Voucher
 
 with scopes_disabled():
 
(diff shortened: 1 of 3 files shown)


app/eventyay/api/views/voucher.py
```diff
@@ -16,7 +16,7 @@
 from rest_framework.response import Response
 
 from eventyay.api.serializers.voucher import VoucherSerializer
-from eventyay.base.models import Voucher
+from eventyay.base.models import Order, OrderPosition, Voucher
```
Rachit7168: Order and OrderPosition are imported but never used in this file. The view only calls instance._clear_canceled_order_positions(), which handles the logic internally on the model. Please remove these unused imports.

app/eventyay/api/views/voucher.py
```diff
@@ -132,6 +132,7 @@ def perform_destroy(self, instance):
         with transaction.atomic():
             instance.cartposition_set.filter(addon_to__isnull=False).delete()
```
Rachit7168: This existing logic has a flaw. instance.cartposition_set only returns positions that use this exact voucher. If an addon item doesn't use the voucher itself, it won't be deleted here, which will cause the next line (instance.cartposition_set.all().delete()) to crash with a ProtectedError (since CartPosition.addon_to has a PROTECT constraint).

You should fix this by changing it to CartPosition.objects.filter(addon_to__voucher=instance).delete() to match how it is correctly handled in VoucherDelete.form_valid(). (Make sure to add CartPosition to your imports).

app/eventyay/control/views/vouchers.py
```diff
@@ -520,6 +526,7 @@ def post(self, request, *args, **kwargs):
                     obj.log_action('eventyay.voucher.deleted', user=self.request.user)
                     OrderPosition.objects.filter(addon_to__voucher=obj).delete()
```
Rachit7168: This looks like a copy-paste error from the original codebase that you should definitely fix in this PR. It uses OrderPosition.objects instead of CartPosition.objects.

Since your PR correctly allows vouchers to be deleted if their orders are canceled, this line will now wrongly delete actual order line items (addons) from canceled orders, destroying historical records. Additionally, it will fail to clear CartPosition addons during bulk delete, causing a ProtectedError on the very next line.

Please change this to CartPosition.objects.filter(addon_to__voucher=obj).delete() so it matches VoucherDelete.form_valid().

Rachit7168 (changes requested):
Please address

Rachit7168 (approved):
https://github.com/user-attachments/assets/e52f7c56-9786-455d-9422-552e958153c6

LGTM , tested

Saksham-Sirohi (approved):

PR #4066, opened 2026-06-26
Fix ticket code button to show PDF-matching QR code

ArnavBallinCode:
The Ticket code button was showing a QR code of just the raw position secret, while the actual ticket PDF QR code encodes a JSON object {event, ticket, lead}. Add a ticket_qrcode_content property to OrderPosition and use it in the order detail template.

image

## Summary by Sourcery

Align ticket QR code content in the order detail view with the JSON payload used in generated ticket PDFs.

New Features:
- Expose a ticket_qrcode_content property on order positions providing the JSON-encoded data for ticket QR codes.

Bug Fixes:
- Update the order detail 'Ticket code' button to use the same JSON-based QR code content as the ticket PDF instead of the raw position secret.

### DIFF
--- a/app/eventyay/base/models/orders.py
+++ b/app/eventyay/base/models/orders.py
@@ -2223,6 +2223,15 @@ class OrderPosition(AbstractPosition):
             self.event.settings.ticket_download_nonadm or self.product.admission
         )
 
+    @property
+    def ticket_qrcode_content(self):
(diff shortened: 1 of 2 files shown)


Rachit7168 (approved):
https://github.com/user-attachments/assets/27d9617e-5fe8-4d0a-bd80-f10159089d8d

LGMT tested on both the current dev and this PR branch

Saksham-Sirohi (approved):

PR #4083, opened 2026-06-27
fix: use InvalidCacheBackendError instead of bare except in cache fallback

prajakta128:
## What

Replace a bare `except:` clause with the specific exception type `InvalidCacheBackendError` in the Redis cache fallback in `presale/views/cart.py`.

## Why

A bare `except:` catches all `BaseException` subclasses including `KeyboardInterrupt` and `SystemExit`, which can silently hide real errors and interfere with process shutdown.

The only expected failure when calling `caches['redis']` is that the `redis` alias is not configured in Django's CACHES setting, which raises `InvalidCacheBackendError` specifically.

## Changes

- Added import: `from django.core.cache.backends.base import InvalidCacheBackendError`
- Changed `except:` → `except InvalidCacheBackendError:`

## Summary by Sourcery

Bug Fixes:
- Handle only InvalidCacheBackendError when the Redis cache alias is unavailable, avoiding unintended swallowing of other exceptions in the cart view cache fallback.


ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):

PR #4127, opened 2026-06-28
feat: Jitsi video integration

### LINKED ISSUE #4074: Integrate Jitsi as a video channel provider in Eventyay Video
## Summary

Eventyay Video currently supports BigBlueButton, Zoom, Janus, native livestreams, and iframe-based media rooms, but it does not have a dedicated Jitsi room integration. This issue proposes adding Jitsi as a first-class video channel provider in Eventyay Video.

The goal is to let organizers create and manage Jitsi-powered video rooms from the Eventyay Video admin UI, while keeping Eventyay’s existing room visibility, role, trait, and permission checks as the source of truth.

  ## Motivation

Jitsi is a widely used open-source video conferencing platform and would be a useful provider option for Eventyay
events, especially for communities that prefer open infrastructure or already run their own Jitsi Meet instance.

Eventyay already has a comparable provider pattern through the BigBlueButton integration:

  - room module: `call.bigbluebutton`
  - WebSocket command: `bbb.room_url`
  - permission checks: `room:bbb.join`, `room:bbb.moderate`, `room:bbb.recordings`
  - admin room configuration UI
  - frontend iframe/media rendering through the video frontend

A Jitsi integration can follow the same architecture instead of being handled as a generic iframe page.

  ## Proposed Behavior

Organizers should be able to:

  - create a Jitsi video channel from the Eventyay Video admin UI
  - configure the Jitsi domain and room name
  - edit Jitsi room settings later
  - allow attendees to join through the Eventyay Video frontend
  - rely on Eventyay permissions before any Jitsi iframe or join URL is returned

Attendees should be able to:

  - enter a Jitsi room only if they have the required Eventyay room permission
  - join with their Eventyay display name where available
  - see a clear error if they do not have access or have no required profile name

  ## Permission Considerations
The video permission system is currently spread across event roles, room roles, trait grants, explicit grants, frontend aliases, and WebSocket command checks. The Jitsi integration should not bypass this system.

The backend WebSocket command should enforce permissions using the same pattern as BBB and Zoom, for example:

  - require the room to contain call.jitsi
  - require room:jitsi.join
  - deny unauthenticated or unauthorized users
  - avoid exposing provider secrets in frontend room config

> A follow-up decision is needed on whether Jitsi room creation should reuse the existing video-channel creation permission or introduce a provider-neutral permission such as

  event:rooms.create.video.

  ## Initial Scope

  The first implementation should focus on a basic Jitsi room integration:

  - Jitsi room type in the admin UI
  - backend WebSocket command to return a join URL or join config
  - frontend rendering through iframe or Jitsi IFrame API
  - Eventyay permission enforcement
  - tests for allowed and denied access

  Advanced secure Jitsi features can be handled later, including:

  - JWT-based Jitsi authentication
  - moderator claim mapping
  - self-hosted Jitsi provider settings
  - lobby/wait-for-host behavior
  - per-event Jitsi server configuration

  ## Acceptance Criteria

  - Admin users can create and edit a Jitsi video channel.
  - A room with call.jitsi can be opened in the Eventyay Video frontend.
  - Users without the required permission cannot get a Jitsi join URL/config.
  - Users with the required permission can join the Jitsi room.
  - Jitsi room configuration exposed to the frontend does not leak secrets.
  - Existing BBB, Zoom, Janus, chat, and livestream behavior remains unchanged.
  - Tests cover permission denial, missing profile name, wrong room type, and successful Jitsi join config generation.

  ## References

  Jitsi IFrame API documentation:
  https://jitsi.github.io/handbook/docs/dev-guide/dev-guide-iframe/

  Jitsi token authentication documentation:
  https://jitsi.github.io/handbook/docs/devops-guide/token-authentication/

sarafarajnasardi:
Fixes #4074

Adds basic Jitsi integration to Eventyay Video.

This PR introduces Jitsi as a first-class video provider foundation, including server configuration, room-level integration support, admin access, and safer room configuration handling. Admins can configure Jitsi servers from the video admin area, and video rooms can use Jitsi through the existing Eventyay Video management flow.

## Highlights

- Adds basic Jitsi video room integration.
- Adds Jitsi server configuration support.
- Exposes Jitsi configuration in the Video Admin interface.
- Keeps Jitsi server credentials out of room config.
- Supports event-scoped Jitsi server selection.
- Adds Jitsi documentation and deployment notes.
- Fixes video room config updates failing due to missing event scope.

https://github.com/user-attachments/assets/c7fd6989-7b17-4bf4-bb52-47a95a135d4a


app/eventyay/features/live/modules/jitsi.py
```diff
@@ -0,0 +1,176 @@
...
+            raise ConsumerException("jitsi.missing_jwt_config")
+
+        now = int(time.time())
+        payload = {
+            "aud": "jitsi",
+            "iss": app_id,
+            "sub": domain,
+            "room": room_name,
+            "nbf": now - 10,
+            "exp": now + 60 * 60,
```
Aqil-Ahmad: JWT expires in 1h.  i recommend either a longer exp, or handle the External API's tokenWillExpire/participantKicked/videoConferenceLeft events to re-request room_config and update the token.

app/eventyay/webapp/video/src/components/MediaSource.vue
```diff
@@ -481,11 +512,115 @@ async function initializeIframe(mute, skipConsentCheck = false) {
...
+
+function loadJitsiExternalApi(domain) {
+	if (window.JitsiMeetExternalAPI) return Promise.resolve();
+	if (jitsiScriptLoads.has(domain)) return jitsiScriptLoads.get(domain);
+	const promise = new Promise((resolve, reject) => {
+		const script = document.createElement('script');
+		script.async = true;
+		script.src = `https://${domain}/external_api.js`;
+		script.onload = resolve;
+		script.onerror = reject;
```
Aqil-Ahmad: If the script fails to load we cache the rejected promise, so we never retry loading external_api.js for this domain. Let's jitsiScriptLoads.delete(domain) in onerror before rejecting.

app/eventyay/base/models/event.py
```diff
@@ -175,6 +176,38 @@ def default_grants():
...
+    return {
+        role: [
+            normalize_permission_value(permission)
+            for permission in permissions
+        ]
+        for role, permissions in roles.items()
+    }
+
+
+def roles_with_jitsi_defaults(roles):
```
Aqil-Ahmad: This runs on every permission check and force-adds the perms even if an admin removed them from a role — so a role can't opt out of Jitsi. Was a one-time default (data migration) the intent instead?

app/eventyay/base/services/event.py
```diff
@@ -372,6 +380,25 @@ async def create_room(event, data, creator):
         m = [m for m in data.get("modules", []) if m["type"] == "call.bigbluebutton"][0]
         m["config"] = event.config.get("bbb_defaults", {})
         m["config"].pop("secret", None)  # legacy
+    elif types == {"call.jitsi"}:
+        if not await event.has_permission_async(
+            user=creator, permission=Permission.EVENT_ROOMS_CREATE_BBB
```
Aqil-Ahmad: Jitsi room creation is gated on the BBB create permission. Granting BBB-create silently grants Jitsi-create — should we add a dedicated EVENT_ROOMS_CREATE_JITSI, or at least note the reuse here?

app/eventyay/base/services/jitsi.py
```diff
@@ -0,0 +1,105 @@
...
+        event=locked_room.event,
+        prefer_server=prefer_server or selected_server_url,
+    )
+    if server is None:
+        return None
+
+    normalized = normalize_server_url(server.url)
+    if normalized and selected_server_url != normalized["url"]:
+        jitsi_config["selected_server_url"] = normalized["url"]
+        locked_room.save(update_fields=["module_config"])
```
Aqil-Ahmad, line 98: When the room has no call.jitsi module, _get_jitsi_config returns a throwaway dict, so this save persists nothing and the sticky-server pick is silently dropped. Either guard earlier or skip the write.

app/eventyay/features/live/modules/jitsi.py
```diff
@@ -0,0 +1,176 @@
...
+        app_secret = server.app_secret
+        if not app_id or not app_secret:
+            raise ConsumerException("jitsi.missing_jwt_config")
+
+        now = int(time.time())
+        payload = {
+            "aud": "jitsi",
+            "iss": app_id,
+            "sub": domain,
+            "room": room_name,
```
Aqil-Ahmad, line 237: room claim is the admin-set room_name. Two rooms with the same name share a Jitsi conference and a token for one works on the other. Probably fine (admin-controlled) but worth a uniqueness note.

Aqil-Ahmad (changes requested):

Saksham-Sirohi:
@sarafarajnasardi I am converting this to a draft for the time being, and the base implementation of Janus needs a bit more polish. I would like this Jitsi implementation to be resumed once you get Janus in a workable state, as part of this may use shared logic.

Saksham-Sirohi (changes requested):

1. Migration chain still depends on `0035` while `dev` is at `0047` — regenerate on current `dev`.
2. Jitsi room creation reuses `Permission.EVENT_ROOMS_CREATE_BBB` — that name no longer matches the behavior.

sarafarajnasardi:
## test guide for this PR

  ### What this PR needs for testing

  This integration expects a Jitsi Meet server with JWT authentication enabled. Eventyay generates JWTs for
  both moderators and participants, and the Jitsi server validates those tokens.

  For full moderation behavior, the Jitsi server should also load the custom Prosody plugins from this PR:

  - `deployment/jitsi/prosody-plugins-custom/mod_eventyay_token_affiliation_force.lua`
  - `deployment/jitsi/prosody-plugins-custom/mod_end_conference.lua`

  ### 1. Start Eventyay locally

  From this branch:

  ```bash
  cp deployment/env.dev.sample .env.dev
  docker compose up -d --build
  docker compose exec web python manage.py migrate
  docker compose exec web python manage.py create_admin_user

  Eventyay should be available at:

  http://localhost:8000

  Admin video area:

  http://localhost:8000/admin/video/

  ### 2. Start a local Jitsi server

  Use the official docker-jitsi-meet release package, not the git checkout:

  wget $(wget -q -O - https://api.github.com/repos/jitsi/docker-jitsi-meet/releases/latest | grep zip | cut-d\" -f4)
  unzip <downloaded-file>.zip
  cd docker-jitsi-meet-*
  cp env.example .env
  ./gen-passwords.sh
  mkdir -p ~/.jitsi-meet-cfg/{web,transcripts,prosody/config,prosody/prosody-plugins-custom,jicofo,jvb,jigasi,jibri}

  Edit .env and set:

  HTTP_PORT=8008
  HTTPS_PORT=8443
  PUBLIC_URL=https://localhost:8443

  ENABLE_AUTH=1
  AUTH_TYPE=jwt
  JWT_APP_ID=eventyay
  JWT_APP_SECRET=eventyay-local-secret
  JWT_ACCEPTED_ISSUERS=eventyay
  JWT_ACCEPTED_AUDIENCES=jitsi
  JWT_ALLOW_EMPTY=0

  XMPP_MUC_MODULES=eventyay_token_affiliation_force,end_conference

  Copy the custom plugins from the Eventyay checkout into the docker-jitsi-meet config directory:

  cp /path/to/eventyay/deployment/jitsi/prosody-plugins-custom/*.lua ~/.jitsi-meet-cfg/prosody/prosody-plugins-custom/

  Start Jitsi:

  docker compose up -d

  Open this once in the browser and accept the local/self-signed certificate if prompted:

  https://localhost:8443

  This is important because Eventyay loads https://localhost:8443/external_api.js in the browser.

  ### 3. Configure the Jitsi server in Eventyay admin

  Go to:

  http://localhost:8000/admin/video/jitsi/new/

  Create a Jitsi server with:

  URL: https://localhost:8443
  App ID: eventyay
  App secret: eventyay-local-secret
  Key ID: leave empty for HS256 local testing
  Active: checked
  Event exclusive: optional; leave empty for a shared test server

  Save it.

  ### 4. Enable Jitsi for the test event

  In the video admin event settings, enable the jitsi feature flag for the event.

  Then open the video frontend for that event, create or edit a channel, and select the Jitsi channel type.

  Channel settings to test:

  Room name: Main Stage
  Preferred Jitsi server URL: https://localhost:8443
  Start with audio muted: optional
  Start with video muted: optional

  ### 5. Expected behavior

  Moderator / room owner:

  - Can create and enter the Jitsi room.
  - Receives a JWT with owner/moderator affiliation.
  - Can start the conference.

  Participant:

  - Can join only after a token owner/moderator is present.
  - Receives a JWT with member affiliation.
  - Should not be promoted to owner by Jitsi/Prosody.

  Also test:

  - Two Eventyay rooms with the same configured Jitsi room name should not collide; the generated Jitsi
    room name is namespaced by event id and room id.
  - If the jitsi feature flag is disabled, the Jitsi room type should not crash the create-channel prompt.
  - If external_api.js fails to load once, re-opening should retry instead of caching the failed script
    load forever.

 
  Notes:

  - Local Jitsi must be served over HTTPS for camera/microphone access.
  - If using a real domain instead of localhost, set PUBLIC_URL=https://your-jitsi-domain in docker-jitsi-meet and use the same URL in Eventyay admin.
  - The Eventyay App ID and App secret must exactly match JWT_APP_ID and JWT_APP_SECRET from the
    Jitsi .env.


  Sources used for the Docker Jitsi bits: official Jitsi Docker self-hosting docs and JWT auth docs:
  - https://jitsi.github.io/handbook/docs/devops-guide/devops-guide-docker/
  - https://jitsi.github.io/handbook/docs/devops-guide/token-authentication/

Rachit7168 (changes requested):
Please address Copilot's comments

app/eventyay/base/migrations/0052_merge_jitsi_server_and_video_permissions.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('base', '0051_consolidate_video_team_permissions'),
+        ('base', '0051_jitsiserver'),
+    ]
+
+    operations = []
```
Aqil-Ahmad: this seems useless, and I'm confused why would two migrations with same number be there in base

Aqil-Ahmad (changes requested):
CI also fails please fix that as well

app/eventyay/base/migrations/0052_merge_jitsi_server_and_video_permissions.py
sarafarajnasardi: fixing

Rachit7168 (changes requested):
I’m unable to join the meeting even though I’ve followed all the steps you provided. It would be really helpful if you could also provide a setup video demonstrating the complete process. This would make it easier for others to set it up and review as well, while saving time and reducing the hassle of troubleshooting setup issues.

https://github.com/user-attachments/assets/a5e3f352-be9f-4d2a-8af7-c6350c018570

sarafarajnasardi:
> I’m unable to join the meeting even though I’ve followed all the steps you provided. It would be really helpful if you could also provide a setup video demonstrating the complete process. This would make it easier for others to set it up and review as well, while saving time and reducing the hassle of troubleshooting setup issues.
> 
>  simplescreenrecorder-2026-08-10_14.02.33.mp4

okay wait i"ll host that jitsi server on VM and share the credentials with you

Rachit7168:
Also renumber the migration

ViRUS-0-0:
@sarafarajnasardi Can you please update the video in the PR description with the latest commit and functionality. It is a month old.

Rachit7168 (approved):
https://github.com/user-attachments/assets/0b528dec-c09b-4adb-8354-c5e4a994a7ab

LGTM , Tested

Saksham-Sirohi (changes requested):
## Review: Jitsi video integration

Thanks for the thorough work here — permission-gated `jitsi.room_config`, secret stripping on room config, HS256 room-scoped JWTs (no `*` claim), and Prosody affiliation forcing are in good shape overall.

### Requested changes

1. **Sticky server selection can still split a room** (`choose_server_for_room`)
   When `prefer_server` is set but matches no active server, selection randomly falls through and rewrites `selected_server_url`. Two attendees can get different domains and never meet.
   **Fix:** If preference matching fails, fall back to `selected_server_url` (then global/event pool) instead of treating a bad preference as “no sticky state.”

2. **Conference name should not depend on `room.name`**
   `normalize_jitsi_room_name(..., self.room.name)` makes the JWT `room` claim follow a mutable/I18n field even though `event-{id}-room-{id}` is already unique. Renames (and localized names) can move or split the live meeting.
   **Fix:** Use a stable id-only room claim (or persist an immutable suffix at room creation and never derive it from live `room.name`).

3. **Moderator JWT lifetime / revocation**
   Tokens live **12 hours** and carry moderator affiliation, with no re-issue path in the iframe flow. Demotion in Eventyay does not demote an already-issued Jitsi owner token.
   **Fix:** Shorten `exp` and/or document + implement refresh; do not describe these as “short-lived” until that exists.

4. **Runtime permission force-add blocks opt-out**
   `permissions_with_jitsi_defaults()` re-injects Jitsi permissions on every check. After the migration, roles cannot drop `room:jitsi.*` even intentionally.
   **Fix:** Keep the data migration; drop (or gate) the permanent runtime augment so custom roles remain editable.

### Non-blocking

- Prefer **HTTPS-only** Jitsi server URLs while JWTs are delivered in the query string.
- JWT secret handling, client secret stripping, and `room_action` auth for join look correct; no websocket auth bypass found in this diff.

app/eventyay/base/services/jitsi.py
```diff
@@ -0,0 +1,158 @@
...
+    for preferred_url in (prefer_server, selected_server_url):
+        if not preferred_url:
+            continue
+        server = _choose_preferred_server(servers, locked_room.event, preferred_url)
+        if server:
+            break
+    if server is None:
+        server = _choose_any_available_server(servers, locked_room.event)
+    if server is None:
+        return None
```
ArnavBallinCode, line 93: can you clarify this 

falling back to another Jitsi server when a room has a `prefer_server` / `selected_server_url` that is no longer available, when it has these set, it still falls back ?

sarafarajnasardi: actually i am also not sure about this bcz i have not tested it with multiple server but as per the logic ``prefer_server`` and ``selected_server_url`` are soft preferences. ``choose_server_for_room()`` first tries ``prefer_server``, then the saved ``selected_server_url`` if neither resolves to an active event/global Jitsi server, it falls back to any available server and persists that new server as ``selected_server_url``.

SxxAq, line 93: **Bug: Server selection priority can split room attendees**

`prefer_server` is currently checked *before* `selected_server_url`. If `prefer_server` matches a domain or pool with multiple active server records in the DB, `_choose_preferred_server` calls `random.choice(preferred_servers)` on every join request—overriding any established `selected_server_url`.

As a result, attendees joining the same room can be randomly assigned to different Jitsi server instances from the pool.

**Suggested Fix:**
Prioritize `selected_server_url` first so established sticky room assignments are preserved:
```python
for preferred_url in (selected_server_url, prefer_server):
    if not preferred_url:
        continue
    server = _choose_preferred_server(servers, locked_room.event, preferred_url)
    if server:
        break

File: app/eventyay/webapp/video/src/components/MediaSource.vue [hunk 9]
```diff
@@ -641,13 +671,136 @@ async function initializeIframe(mute, skipConsentCheck = false) {
...
+				resolve(window.JitsiMeetExternalAPI);
+			} else {
+				reject(new Error('Jitsi external API did not load'));
+			}
+		};
+		script.onerror = () => reject(new Error('Jitsi external API could not be loaded'));
+		document.head.appendChild(script);
+	});
+	jitsiExternalApiLoaders.set(scriptUrl, loader);
+	return loader;
```
SxxAq: **Bug: Failed script loader promise is permanently cached**

If `script.onerror` triggers (e.g. due to a temporary network glitch, DNS failure, or self-signed cert issue on `https://<jitsi-domain>/external_api.js`), `jitsiExternalApiLoaders` retains the rejected promise.

If the user re-enters or switches back to the room, `jitsiExternalApiLoaders.get(scriptUrl)` returns the cached rejected promise without attempting to append a new script tag and retry.

**Suggested Fix:**
Clean up the loader entry on error:
```javascript
script.onerror = () => {
    jitsiExternalApiLoaders.delete(scriptUrl);
    reject(new Error('Jitsi external API could not be loaded'));
};

File: app/eventyay/features/live/modules/jitsi.py [hunk 10]
```diff
@@ -0,0 +1,239 @@
...
+JITSI_JWT_LIFETIME_SECONDS = 10 * 60
+
+
+def normalize_jitsi_room_name(
+    configured_room_name,
+    event_id,
+    room_id,
+    room_name_fallback="room",
+):
+    return f"event-{event_id}-room-{room_id}"
```
SxxAq: **Cleanup: Unused positional parameters in `normalize_jitsi_room_name`**

`configured_room_name` and `room_name_fallback` are unused in the function body:

app/eventyay/webapp/video/src/components/MediaSource.vue
```diff
@@ -641,13 +671,136 @@ async function initializeIframe(mute, skipConsentCheck = false) {
...
+	}
+	if (!Object.keys(commands).length) return;
+	try {
+		jitsiApi.executeCommands(commands);
+	} catch (error) {
+		console.warn('Failed to apply Jitsi display overrides', {
+			roomId: props.room?.id,
+			error,
+		});
+	}
```
SxxAq, line 749: **Safety Check: Guard against `null` `jitsiApi` instance**

The `videoConferenceJoined` event callback invokes `applyJitsiDisplayOverrides`. If the component or iframe is unmounted right as the event fires, `jitsiApi` will be `null`, causing `jitsiApi.executeCommands(...)` to throw an uncaught error.

**Suggested Fix:**
Add a null check at the start of the function:
```javascript
function applyJitsiDisplayOverrides(config) {
    if (!jitsiApi) return;
    ...
}

SxxAq (changes requested):
The core architecture is well designed.

Before merging, please address the following issues:

ArnavBallinCode (approved):

Saksham-Sirohi (approved):

PR #4135, opened 2026-06-29
fix: replace bare except clauses with specific exceptions in reldate, signals, and rebuild

prajakta128:
## Problem
Three files used bare `except:` clauses which silently catch everything including `SystemExit` and `KeyboardInterrupt`, making bugs hard to trace and the application difficult to stop during debugging.

## Changes

| File | Before | After | Reason |
|------|--------|-------|--------|
| `plugins/returnurl/signals.py` | `except:` | `except Resolver404:` | `resolve()` raises `Resolver404` for unmatched URLs — most specific fix |
| `base/reldate.py` | `except:` | `except (ValueError, Exception):` | `from_string()` raises `ValueError` for invalid date strings |
| `core/management/commands/rebuild.py` | `except:` | `except Exception:` | Catches DB config errors on startup without swallowing system signals |

## Why This Matters
- **`Resolver404`** is the most precise fix — directly matches what Django' `resolve()` raises for unmatched URLs
- **`ValueError`** in reldate matches what the parser actually raises
- **`Exception`** in rebuild.py ensures `KeyboardInterrupt` and `SystemExit` are never accidentally swallowed

## Testing
Behaviour is fully preserved — only the exception scope is narrowed to what can realistically be raised in each block.


ArnavBallinCode (approved):
Thanks for the contribution, please try to consolidate such similar small changes into 1 pr instead of opening multiple prs for same small fixes

PR #3778, opened 2026-06-05
Introduce shared Tiptap editor system for rich text and email editing

Rachit7168:
## Summary

This PR introduces a unified Tiptap-based editor layer for eventyay as the shared foundation for new rich text and email editing use cases. **The primary production wiring is the Tickets Message center** (compose, outbox edit, team compose). Order-level “Send email” also uses the same email editor as a secondary path. Toast UI (Talk Markdown) and Quill (video) are unchanged.

- Adds a new Vite-built frontend bundle at `app/eventyay/webapp/editor/`
- Adds reusable Django widgets/fields with progressive enhancement over plain textareas
- Adds server-side HTML sanitization via nh3
- **Wires the email editor into Message center sendmail forms** (`MailForm`, `EmailQueueEditForm`, `TeamMailForm`)
- Wires the email editor into order send-mail screens in control (secondary)
- Adds `I18nEmailBodyFormField` / `I18nEmailEditorWidget` for multi-locale Message center bodies
- Adds `compile_email_body()` so Tiptap HTML is not re-processed as Markdown on preview/send
- Fixes placeholder serialization so variables expand correctly via Python `str.format_map()`
- Shows per-locale flag icons in i18n Tiptap message fields (matching previous Markdown editor behavior)

Part one of #3684

## Motivation

eventyay currently uses multiple editor stacks:

| Area | Current editor |
|------|----------------|
| Talk submission forms | Toast UI / Markdown |
| Video app | Quill |
| Ticket/email forms | Plain textarea / Markdown (Message center) |

Ticket and email workflows need simple rich text editing, reliable HTML output, and placeholder support. Markdown is not a good canonical format for transactional email, and expanding Quill or Toast UI into more areas would increase fragmentation.

Review feedback on the initial PR correctly noted that **Message center** is the main user-facing email composition area—not the rarely used per-order “Send email” action. This update focuses Tiptap integration and documentation there.

### Frontend: shared Tiptap bundle

Added `app/eventyay/webapp/editor/` with two profiles:

- **richtext** — bold, italic, underline, bullet/numbered lists, links, clear formatting, undo/redo
- **email** — extends richtext with placeholder insertion and in-editor preview

Key behavior:

- Mounts on `textarea[data-tiptap-profile]`
- Syncs editor HTML back to the original textarea before form submit
- Falls back to plain textarea when JavaScript is unavailable
- Placeholder chips serialize as `{variable}` (single braces) for compatibility with `str.format_map()`
[...]

PR #4146, opened 2026-06-29
Revert "Introduce shared Tiptap editor system for rich text and email editing"

Rachit7168:
Reverts fossasia/eventyay#3778

As changes were not being shown in Dev


Saksham-Sirohi (approved):
Does not work properly on prod; make a fixed scoped pr

PR #4181, opened 2026-07-02
feat:Add mirrored Admin Video Settings UI synced with main Admin Settings

### LINKED ISSUE #3208: [Video] Add mirrored Admin Video Settings UI synced with main Admin Settings
## Summary
Create a duplicate/mirrored Admin Video Settings UI in the Video area that stays synchronized with the primary Admin Settings UI.

## Scope
- Provide Video-side settings interface for admin video settings.
- Keep values in sync bidirectionally with the primary Admin Settings source.
- Ensure consistent validation rules, defaults, and save behavior.
- Prevent drift/inconsistency between the two settings UIs.

## Acceptance Criteria
- Updating setting from Video-side UI is reflected in main Admin Settings.
- Updating in main Admin Settings is reflected in Video-side UI.
- No conflicting values after refresh/reload.
- Error handling is consistent across both UIs.

## Notes
This issue is for the **Video area only**.

sarafarajnasardi:
## Summary
  - Add an embedded Video Admin entry and page for users in video admin mode.
  - Restrict `/admin/video/` access to authenticated staff users only.
  - Allow same-origin iframe embedding for the video admin panel.
  - Add regression coverage for video admin access rules.

Fixes #3208

https://github.com/user-attachments/assets/50338b8d-912d-4022-9c07-53cb301ef59c


ArnavBallinCode (approved):
LGTM

app/eventyay/webapp/video/src/views/admin/config/video-admin.vue
```diff
@@ -0,0 +1,212 @@
...
+				this.embedAdminUrl(link, 'href')
+			}
+			for (const form of doc.querySelectorAll('form[action]')) {
+				this.embedAdminUrl(form, 'action')
+			}
+			requestAnimationFrame(() => {
+				this.frameReady = true
+			})
+		},
+		embedAdminUrl(element, attribute) {
```
Rachit7168, line 143: Using element.getAttribute(attribute) retrieves the raw HTML string (e.g., "events/" or "?page=2"). By resolving it with new URL(value, window.location.origin), you are resolving it against the parent window's origin instead of the iframe's base URL. This means relative links like href="events/" will incorrectly resolve to http://origin/events/ and fail the startsWith(ADMIN_BASE) check, meaning they won't get the embedded=1 flag attached.

You can fix this by using the DOM properties element.href and element.action, which the browser automatically resolves to absolute URLs based on the iframe's current location

app/eventyay/webapp/video/src/views/admin/config/video-admin.vue
```diff
@@ -0,0 +1,212 @@
...
+				return
+			}
+			this.activePath = nextPath
+			this.frameReady = false
+			this.$router.replace({
+				name: 'admin:video-admin',
+				query: this.activePath ? { admin_path: this.activePath } : {}
+			})
+		},
+		prepareFrame() {
```
Rachit7168, line 93: When a user clicks a link or submits a form inside the iframe, the new page will start loading, but this.frameReady will remain true. This causes the user to see an unstyled/flickering page (including the original navbar/sidebar) while it is loading until the load event fires again. You can fix this by adding a beforeunload event listener to reset the loading state

app/eventyay/webapp/video/src/views/admin/config/video-admin.vue
```diff
@@ -0,0 +1,212 @@
...
+			const frame = this.$refs.frame
+			const doc = frame?.contentDocument
+			if (!doc || !doc.location.pathname.startsWith(ADMIN_BASE)) {
+				this.frameReady = true
+				return
+			}
+
+			if (!doc.getElementById('video-admin-embedded-style')) {
+				const style = doc.createElement('style')
```
Rachit7168, line 104: Since this CSS is injected on the iframe's load event (which fires after all assets are downloaded and the page has been painted), users will likely experience a Flash of Unstyled Content (FOUC)—seeing the navbar and sidebar for a split second before they disappear on every click.

Long-term recommendation: Instead of hiding elements with injected CSS on the frontend, update the Django backend template (e.g. the base admin layout) to detect the request.GET.embedded query parameter and simply skip rendering the navbar, sidebar, and footer server-side.

Rachit7168 (changes requested):
Please address

SxxAq (approved):
https://github.com/user-attachments/assets/24356176-38ac-48a3-8145-5860ee1f747d

Saksham-Sirohi (approved):
Let's get this moving; however, there needs to be a ui optimization at a later point

PR #4185, opened 2026-07-02
feat: implement meetup event support

### LINKED ISSUE #4236: Feature: Streamlined Event View and RSVP for Meetups
## Goal
We want to make Eventyay easier to use for small events like meetups, workshops, and community gatherings. Currently, setting up any event requires a lot of complex steps (like configuring schedules, speaker profiles, and multiple ticket types). We want to introduce a "Simple Event View" (Meetup Style) that allows organizers to create a meetup quickly, configure a live video stream room easily, and let users register/RSVP with one simple step.

## Current Behavior
- Creating an event is complex and requires setting up multiple configurations.
- Organizers must fill out many details and configure schedules, speakers, and rooms.
- Organizer dashboards always display advanced conference tools like Call for Proposals and review boards.
- Attendees must navigate a multi-page checkout wizard (including cart, questionnaire, and confirmation steps) to register.
- There is no native one-step RSVP option to sign up quickly for free simple events.

## Expected Behavior
- Organizers can select a "Create Meetup" option if enabled by the site administrator.
- The system presents a simplified single-page creation form.
- The creation form includes options to configure a live video stream.
- A virtual room is automatically configured behind the scenes upon creation.
- Organizer dashboards automatically hide Call for Proposals, Schedule Editor, Speakers, Submissions, and Reviews.
- Attendees can register for the meetup using a simplified registration flow.
- Registered attendees can access the video stream directly through the video platform.

## Features
- Global setting toggle in the site admin panel to enable or disable meetup creation.
- Buttons on the dashboards for organizers to "Create Meetup".
- Streamlined creation form showing only basic event inputs.
- Integrated video fields for YouTube, HLS, or Iframe streams in the creation form.
- Automated default virtual room setup when creating a meetup.
- Simplified sidebar navigation hiding advanced conference focused tabs.
- One-step RSVP registration form for attendees to sign up quickly.

## Expected Outcome
- Organizers can set up meetup events easily without advanced configurations.
- Organizers manage simple events with a clean sidebar dashboard interface.
- Attendees register for meetups without entering complex billing information.
- Attendees gain direct access to watch streams on the video platform.
- Admins maintain control over whether simple meetup options are available on the platform.

prayag78:
Fixes #4236

## Summary
Introduces the Meetup Creation feature from the global admin toggle through event creation, RSVP registration, video streaming while leaving all existing behavior for standard events completely unchanged.

### Global admin toggle
- Added `MEETUP_CREATION_ENABLED` constant and `is_meetup_creation_enabled()` helper in `settings.py`, following the same pattern as `EVENT_SERIES_CREATION_ENABLED`.
- Registered the toggle field in `GlobalSettingsForm` under the "Event Creation" tab with a dedicated UI block in the admin settings template.
- Default is `False` admins must explicitly enable it.

### Meetup creation entry points
- "Create Meetup" buttons added on four surfaces: the control dashboard, the eventyay-common dashboard, the events index page, and the organizer detail page all gated behind `meetup_creation_enabled`.
- `EventCreateView.dispatch()` raises `PermissionDenied` if `?meetup=1` is requested when the toggle is off.

### Streamlined creation form
- Added `MeetupEventWizardBasicsForm` extending `EventWizardBasicsForm` with optional `video_type` / `video_url` fields and cross-field validation.
- `get_video_module_config()` shared helper converts the video type choice into the `module_config` JSON shape expected by the video platform.
- Currency field is made optional and defaulted to `DEFAULT_CURRENCY`, meetup events are always free so organizers don't need to set one.
- Creation template shows meetup-specific copy and hides the currency field when `is_meetup=True`.
- An `is_meetup` hidden field carries the meetup flag across POST round-trips.

### Auto-provisioning on creation
- Sets `event.settings.event_type = 'meetup'`.
- Auto-publishes the event (`live=True`, `tickets_published=True`) since meetups are simple free events.
- Provisions JWT secrets and all `venueless_*` settings so the video platform is ready from day one without organizer configuration.
- Creates a Room ("Main Room") with `module_config` set from the chosen video type.
- Creates a free Product ("RSVP Ticket", `admission=True`, `price=0`) and an unlimited Quota ("RSVP", `size=None`) linked to it.

### RSVP registration flow
- New `MeetupRsvpView` registered at `rsvp/` in `frame_wrapped_urls`.
- Authenticated users get an instant single-click RSVP; duplicate submissions are silently de-duplicated.
- After registration, users are redirected back to the event index page where a success panel is shown.
- `EventIndex` checks both the user's orders (authenticated) and the session (guest) to set `attendee_already_registered`.

### Video stream
- `JoinOnlineVideoView` performs a double-checked locked lazy init of JWT/venueless config for meetup events at join time, guarding against race conditions.
- `show_online_video_link` in the context processor is based on `room.module_config` presence for meetup events rather than the `venueless_url` setting so the nav link only appears when a stream is actually configured.

### Video settings editable post-creation
- `EventCommonSettingsForm` dynamically adds `video_type` and `video_url` fields for meetup events, pre-populated from the existing room's `module_config`.
- Saving updates the room's `module_config` in place (or creates the room if absent).
- The settings page renders a "Video stream" fieldset with the same show/hide behaviour as the creation form.

### Code quality
- Extracted shared meetup logic into `app/eventyay/base/meetup.py` to eliminate repetition across forms, views, and context processors.
- All helpers are gated on `event_type == meetup` no behaviour change for standard events.

> **Scope:** This PR covers the first implementation of meetup creation. Further changes will be added after review and additional testing. Organizer sidebar simplification (hiding CFP/Schedule/Speakers for meetup events), video component sidebar, email updates, public event page and more is scoped to follow-up PRs.

Screenshot 2026-07-11 084139

https://github.com/user-attachments/assets/4159af1b-0faa-450d-a306-42285d174b4b


app/eventyay/common/context_processors.py
```diff
@@ -9,7 +9,9 @@
 from django.urls import resolve
 from django.utils import translation
 from django.utils.translation import gettext_lazy as _
-from django_scopes import get_scope
+from django_scopes import get_scope, scope
```
Rachit7168: Context processors are executed on every single page render across the entire application (both the frontend public shop and the backend control panel). Querying the Room table here forces an un-cached database lookup on every single page load for meetup events, which is a major performance regression. Fix: Instead of querying the database, cache this boolean in the event settings:

In app/eventyay/eventyay_common/forms/event.py (EventSettingsForm.save), after getting the module_config, add: self.event.settings.set('meetup_video_active', bool(module_config))
In _provision_meetup_event(), add: event.settings.set('meetup_video_active', bool(module_config))
Then, in the context processor, you can simply do: meetup_video_active = event.settings.get('meetup_video_active', as_type=bool, default=False)

Rachit7168 (changes requested):
Please address : 

Cloning a Meetup Event In app/eventyay/eventyay_common/views/event.py, the _provision_meetup_event() function is responsible for setting up the Room and JWT secrets. If an organizer clones an existing meetup event, the cloning framework will copy the event_type='meetup' setting. However, because the clone URL usually doesn't have ?meetup=1 and is_meetup isn't in the POST data, self.is_meetup_request evaluates to False.

Therefore, _provision_meetup_event() won't run.
While the "RSVP Ticket" product and quota will be duplicated by the standard cloning process, the Room and Venueless JWT secrets will not be copied or generated. The cloned meetup will be completely broken for live streaming. Fix: You need to either register a signal receiver for event cloning that duplicates the Room and generates new JWT secrets, or explicitly detect during the EventCreateView clone process if the source event was a meetup and provision the video resources accordingly.

app/eventyay/common/context_processors.py
prayag78: fixed

sarafarajnasardi (approved):
LGTM

sarafarajnasardi (changes requested):
When we try to edit the room from the rooms it shows the blank page ptal

https://github.com/user-attachments/assets/22b8641f-ef49-4747-8084-24f93cf2b9c1

app/eventyay/presale/views/meetup.py
```diff
@@ -0,0 +1,177 @@
...
+            return self._handle_authenticated_rsvp(request, product, quota)
+        else:
+            return self._handle_guest_rsvp(request, product, quota)
+
+    def _handle_authenticated_rsvp(self, request, product, quota):
+        with scope(event=request.event):
+            existing = request.event.orders.filter(
+                email=request.user.email,
+                status__in=[Order.STATUS_PAID, Order.STATUS_PENDING],
+            ).first()
```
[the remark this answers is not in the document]
prayag78: fixed

app/eventyay/common/context_processors.py
```diff
@@ -9,7 +9,8 @@
...
```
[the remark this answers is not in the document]
prayag78: invalid

prayag78:
> When we try to edit the room from the rooms it shows the blank page ptal
> 
>  Screencast.From.2026-07-14.21-32-18.mp4

Fixed

https://github.com/user-attachments/assets/85f8d760-1103-4768-b66a-327334a363fa

app/eventyay/control/forms/event.py
```diff
@@ -1810,3 +1818,53 @@ def __init__(self, *args, **kwargs):
...
+        choices=VIDEO_TYPE_CHOICES,
+        required=False,
+        label=_('Video stream type'),
+        help_text=_('Optional: configure a live video stream for this meetup.'),
+    )
+    video_url = forms.URLField(
+        required=False,
+        label=_('Video URL / stream identifier'),
+        help_text=_('YouTube video URL, HLS stream URL, or embed URL.'),
+    )
```
sarafarajnasardi: This field is labeled as “Video URL / stream identifier”, and the video admin UI accepts “YouTube Video ID or URL”, but `forms.URLField` rejects a raw YouTube ID. Repro: create a meetup, choose YouTube as video type, enter a valid 11-char YouTube ID instead of a full URL, and submit. Expected: accepted and stored as `ytid`, matching the video admin behavior. Currently Showing error...

prayag78: fixed

app/eventyay/presale/views/meetup.py
```diff
@@ -0,0 +1,148 @@
...
+    def _render_index_with_form_errors(self, request, form):
+        # Imported here to break the circular import with the presale event views.
+        from eventyay.presale.views.event import EventIndex
+
+        request._rsvp_guest_form = form
+        view = EventIndex()
+        view.setup(request, *self.args, **self.kwargs)
+        return view.get(request, *self.args, **self.kwargs)
+
+    def _create_rsvp_order(self, request, product, email, name):
```
Rachit7168, line 116: This one-click RSVP implementation manually creates the order but completely bypasses quota availability checks. If an organizer sets a capacity limit (Quota size) for their meetup, this view will ignore it and allow unlimited RSVPs.

Please check quota availability inside the atomic block before creating the order. For example:
```python
        with scope(event=request.event), transaction.atomic():
            avail = quota.availability()
            if avail[0] != Quota.AVAILABILITY_OK:
                messages.error(request, _('Sorry, this meetup is already full.'))
                return self._redirect_to_index(request)
            # ... proceed with Order creation
```
*(Also note: by bypassing the `OrderManager`, you are also bypassing the `order_placed` signal. If any plugins rely on `order_placed` to sync data, they will be silently broken for Meetups).*

app/eventyay/base/meetup.py
```diff
@@ -0,0 +1,276 @@
...
+                admission=True,
+                active=True,
+            )
+            product.save()
+
+        quota = event.quotas.first()
+        if quota is None:
+            quota = Quota(event=event, name=DEFAULT_QUOTA_NAME, size=None)
+            quota.save()
+            quota.products.add(product)
```
Rachit7168: There is a logic flaw here. If a product was just created, but a quota already existed on the event (`quota is None` is False), the newly created product will never be added to the existing quota because `quota.products.add(product)` is trapped inside the `if` block. A product with no quota cannot be registered for.

Please move the `add()` call outside the `if` block to ensure the product is always assigned to the quota:
```python
        quota = event.quotas.first()
        if quota is None:
            quota = Quota(event=event, name=DEFAULT_QUOTA_NAME, size=None)
            quota.save()
            
        quota.products.add(product)
```

Rachit7168 (changes requested):
This is a great feature that brings a much-needed simplified flow for meetups! The UI integration and video settings logic are solid. However, I found two critical logic bugs regarding how RSVPs and quotas are handled which will cause either overselling or broken registrations. Please address these blockers so we can get this merged.

prayag78:
> This is a great feature that brings a much-needed simplified flow for meetups! The UI integration and video settings logic are solid. However, I found two critical logic bugs regarding how RSVPs and quotas are handled which will cause either overselling or broken registrations. Please address these blockers so we can get this merged.

Thanks for the review! Both quota-related points are valid. I've already created a follow-up PR #4557 for registration limits and quota handling. The availability check and product-quota linking will be addressed there alongside the full limit implementation.

sarafarajnasardi (approved):
LGTM

Rachit7168 (changes requested):
https://github.com/user-attachments/assets/ffef9655-65c1-46b1-8587-5bdbbda09be6

On 1:52 : Check when there is no schedule it keeps showing loading but for speaker it shows no speaker found which is correct !

Rachit7168 (approved):
@prayag , This is an overall issue in events. Approving this PR !

Saksham-Sirohi (changes requested):
Two bugs still in the RSVP path:

1. In `ensure_rsvp_product`, `quota.products.add(product)` is only called when creating a new quota. If a quota already exists and you create a new product, it never gets attached. Move the `add()` outside that `if`.

2. `_create_rsvp_order` never checks quota availability, so a capacity limit can be overbooked. Check `quota.availability()` before creating the order (ideally go through the normal order path so signals still fire).

Please fix both before merge.

Saksham-Sirohi (commented):

One `except Exception: pass` when copying meetup video settings should be narrowed to the expected lookup/settings errors.

Rachit7168:
Please resolve merge conflicts

Rachit7168:
Please resolve merge conflicts

ArnavBallinCode (changes requested):
please resolve merge conflicts

Aqil-Ahmad (approved):

Aqil-Ahmad (approved):

Rachit7168 (approved):
LGTM , TESTED AGAIN

Saksham-Sirohi (approved):

ArnavBallinCode (approved):

PR #4194, opened 2026-07-03
fix: allow multi-word search in attendee and invoice names

Rachit7168:
## Before 

https://github.com/user-attachments/assets/678a5314-669d-4e7d-8282-9009b31a66da

## After 

https://github.com/user-attachments/assets/ef6c729e-b9a1-47ad-96a7-54d5dfd768ce

Fixes an issue where searching for names with reversed word order (e.g., 'Uzumaki naruto' instead of 'Naruto Uzumaki') yielded no results. The search query is now split by space, requiring all terms to match independently.


Saksham-Sirohi (approved):

ArnavBallinCode (approved):
LGTM

PR #4205, opened 2026-07-03
feat: Add Return to Check-in Lists button on Connected devices page

### LINKED ISSUE #101: Improve navigation from Connected Devices to Check-in Lists
### Description
From the **Check-in Lists** page, click `Connected devices`, it is directed to the **Connected Devices** page. There is no option to return to the Check-In Lists page 

There is an event link within each device row, it only redirects the user to the top-level **Event Dashboard**. Also, because it is just text inside a table column, it isn't obvious as a primary navigation route back to the previous page. 

### Proposed Solution
- Add navigation link back to the event **Check-in Lists** page
- Add: **`← Return to Check-in Lists`** action button at the top of the page (next to `+ Connect a device`) that uses browser history or session memory to instantly take the organizer back to the exact list they were just configuring.

Rachit7168:
Close: https://github.com/fossasia/eventyay-checkin/issues/101

Resolves an issue where navigating to Connected devices from the Check-in lists page had no obvious way to return to the previous page. Added a 'Return to Check-in Lists' button that leverages browser history.

https://github.com/user-attachments/assets/dc66684c-9ba4-4cd5-9bd7-06b55efd6131

## Summary by Sourcery

New Features:
- Introduce a 'Return to Check-in Lists' button on the organizer Connected devices page that leverages browser history for navigation.


app/eventyay/control/templates/pretixcontrol/organizers/devices.html
```diff
@@ -19,11 +19,15 @@ <h1>
...
             </p>
 
+            <button type="button" onclick="window.history.back()"
+                    class="btn btn-default btn-lg"><i class="fa fa-arrow-left"></i> {% trans "Return to Check-in Lists" %}</button>
             <a href="{% url 'eventyay_common:organizer.devices.add' organizer=request.organizer.slug %}"
                     class="btn btn-primary btn-lg"><i class="fa fa-plus"></i> {% trans "Connect a device" %}</a>
         </div>
     {% else %}
         <p class="clearfix">
+            <button type="button" onclick="window.history.back()"
```
ArnavBallinCode, line 29: `window.history.back()` works well for the common flow, but it depends on browser history. If the page is opened directly or in a new tab, the button may not return to Check-in Lists. Would it make sense to add a fallback URL ?

Rachit7168: But which event's check-in list should the fallback URL direct to ?

Saksham-Sirohi (changes requested):
this will not work as this needs context awareness i am already working on this no need for duplicate work

Issue #4202, opened 2026-07-03
Badge preview downloads the file instead of displaying it in the browser


veliadang:
When an organizer wants to review a badge design, the preview action automatically downloads the badge instead of displaying it in the browser. This makes it difficult to quickly review and make further adjustments.

**Steps to reproduce:**

1. Log in as an organizer.
2. Go to Tickets → Badges.
3. Create a new badge layout or open an existing one.
4. Click Preview.
5. Observe that the system automatically downloads the badge instead of displaying a preview in the browser.

**Expected behavior:**
The badge should be displayed directly in the browser for preview without downloading the file.

Rachit7168:
Closing already done and tested

PR #4206, opened 2026-07-03
fix(Badge): allow changing badge/layout size in the PDF editor

### LINKED ISSUE #4201: Organizer cannot change badge size when creating or editing a badge
The organizer cannot change the badge size. The badge size is fixed or automatically follows the size of the background image uploaded by the organizer. This issue also occurs when creating a new badge, as there is no option to adjust the badge size.

**Steps to reproduce:**

1. Log in as an organizer.
2. Go to Tickets → Badges.
3. Create a new badge layout or open an existing badge layout.
4. Click Start Editing.
5. Change the badge size and save the layout.
6. Click on a text element, then click anywhere outside the editor.
7. Observe that the badge size value resets to the default value.

https://github.com/user-attachments/assets/d1be9622-683a-43ab-bc9d-fe813a0eb4d7

**Expected behavior:**
The organizer should be able to set or adjust the badge size independently of the uploaded background image.

Aqil-Ahmad:
fixes #4201 
#4202 

Page-size fields in the layout editor now apply on change (regenerating the background at the new dimensions) and are populated from the actual PDF viewport on load, instead of being reset to default whenever an element was deselected. Note: since page size is defined by the background PDF, changing the size regenerates a blank background, and this shared editor also affects PDF ticket layouts.

https://github.com/user-attachments/assets/fa984712-18c3-45d8-82cc-797269f4dd7c

download fix:

https://github.com/user-attachments/assets/a876ba74-6377-4fd5-b56d-a8347eae7996


ArnavBallinCode (changes requested):
when I upload a background and then change the width it isnt working 

https://github.com/user-attachments/assets/13be2611-f5ed-4f13-99d9-79b5293da386

Aqil-Ahmad:
> when I upload a background and then change the width it isnt working
> 
>  Screen.Recording.2026-07-03.at.10.33.40.PM.mov

As mentioned in PR description, size change needs background regeneration. Once a custom PDF background is uploaded, its dimensions define the badge size, and that size can't be changed afterward in the editor. you can't keep an uploaded design and give it a different size without rebuilding the background.

Supporting size changes for an uploaded background would need extra implementation, and even then it would distort the PDF whenever the new width:height ratio differs from the original. So it's not a good idea either.

For now, size changes are only meaningful for blank/generated backgrounds (via the "Create empty background" button). Uploaded backgrounds keep their own size and are preserved on save.

What we can do now:
1. Rely on the existing "Create empty background" button — type the width/height, then click it to generate a blank page at that size. In that case the background auto generated code needs to be removed. 

2. keep the auto generation but show a message before size is changed when already a pdf background is uploaded, so user is aware of data loss.

ArnavBallinCode:
@hpdang , please let us know how to proceed, Is [this](https://github.com/fossasia/eventyay/pull/4206#pullrequestreview-4627227219) is needed or we can measure it beforehand and edit and upload, etc?

Aqil-Ahmad:
@ArnavBallinCode I removed the auto regeneration of pdf, the PR fixes both bugs now with no unnecessary changes. If we need to change size for uploaded background then we will need to implement pdf cropper maybe but not sure.

Rachit7168 (changes requested):
Working well , But i think the background pdf and the removal can be done in another PR as we are not sure about the approach , and for the pdf background we add It can be enchaned further in another PR not this. I have already checked the badge there some things there needs to be added. For this PR dont add unnessary things

What do u say ? @ArnavBallinCode , @hpdang ?

hpdang:
We do not want to change the size of the uploaded background here. The current fix of the page-size is what is expected.

@Aqil-Ahmad for the preview, can we have it on the same page so users don't need to switch tabs. Thanks

Aqil-Ahmad:
> We do not want to change the size of the uploaded background here. The current fix of the page-size is what is expected.
> 
> @Aqil-Ahmad for the preview, can we have it on the same page so users don't need to switch tabs. Thanks

yes we can do i think. but i am not sure if we can have native pdf viewer (zoom, print, download) inline easily.

hpdang:
> > We do not want to change the size of the uploaded background here. The current fix of the page-size is what is expected.
> > @Aqil-Ahmad for the preview, can we have it on the same page so users don't need to switch tabs. Thanks
> 
> yes we can do i think. but i am not sure if we can have native pdf viewer (zoom, print, download) inline easily.

then keep the fix as it is first and open another issue for future enhancement.

hpdang:
@Rachit7168 @ArnavBallinCode please review

Rachit7168 (approved):
LGTM , TESTED

Further enhancement can be done later 

https://github.com/user-attachments/assets/0f255c57-2077-444c-bcc0-df2fa491a33d

app/eventyay/control/views/pdf.py
```diff
@@ -261,7 +261,7 @@ def post(self, request, *args, **kwargs):
 
             resp = HttpResponse(data, content_type=mimet)
             ftype = fname.split('.')[-1]
-            resp['Content-Disposition'] = 'attachment; filename="ticket-preview.{}"'.format(ftype)
+            resp['Content-Disposition'] = 'inline; filename="ticket-preview.{}"'.format(ftype)
```
ArnavBallinCode, line 264: you just need to add an  <iframe>  to the frontend page where the preview button/action currently resides,
and load the PDF into that frame instead of opening it in a new window or downloading it.

ArnavBallinCode (changes requested):
small change, can be done in this pr itself

Rachit7168 (changes requested):
Please address the double preview thing  we discussed

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/ad353c5a-b1e6-43f4-a971-35ca15211a8d

ArnavBallinCode (approved):
LGTM

PR #4209, opened 2026-07-03
fix: pause interpretation audio when main livestream is paused

### LINKED ISSUE #4172: Pause interpretation audio when main livestream is paused
**Describe the bug**
When viewing a livestream with an audio interpretation channel active, pausing the main video player does not pause the translation audio. The background translation audio continues to play while the visual video feed is paused, causing the audio and video to become out of sync.

**Expected behavior**
When a user pauses the main video feed, the background interpretation audio stream should also pause simultaneously to maintain sync and prevent unwanted audio from playing. When the main video is unpaused, the translation audio should resume.

**Additional context**
This issue occurs when using the "audio-only" translation mode, where the main video is muted, and a separate youtube video or WHEP client handles the translated audio. The pause event from the main player needs to be intercepted and passed down to the background audio player.

sarafarajnasardi:
Fixes #4172

## Summary
  - Pause and resume interpretation audio when the main livestream playback state changes.
  - Keep newly loaded translation audio paused if the main player is already paused.

  ## Screencast
  Unable to provide a screencast because I cannot record a screencast with audio on my laptop.


Rachit7168:
Don't force push unless necessary !

ArnavBallinCode (commented):
can you add a video, also are you dealing with whep based input in the same pr?

sarafarajnasardi:
> can you add a video, also are you dealing with whep based input in the same pr?

yes whep based input is also handled here....about the video I am not able to record the screen-recording with audio

ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):
LGTM

PR #4213, opened 2026-07-04
fix(api): resolve by-version URL namespace and confusion

### LINKED ISSUE #4143: bug(api): schedule by-version redirect crashes with NoReverseMatch (wrong URL namespace)
### Summary

`GET /api/v1/events/{event}/schedules/by-version/?version={name}` fails when a matching schedule version exists. The lookup succeeds, but the redirect step crashes with HTTP **500** because the view uses the wrong Django URL namespace.

There is also confusion around the `version` query parameter: it must be the **schedule version name** (e.g. `0.23`), not the schedule **primary key** (e.g. `50`). Using the id returns HTTP **404** with `No Schedule matches the given query.`

https://github.com/user-attachments/assets/368b4e0b-5440-4252-9478-91c294c8e850

Rachit7168:
Fixes #4143

https://github.com/user-attachments/assets/3de64c1b-9555-4897-8a3e-40284ac7fe9d

**Description:**
- Fixed the URL namespace from `api` to `api-v1` for the schedule detail redirect in the `by-version` endpoint, resolving the HTTP 500 `NoReverseMatch` crash when a version is found.
- Added explicit validation to require the `version` query parameter, returning a 400 Bad Request if missing.
- Addressed confusion around the `version` query parameter: if a user accidentally passes a schedule ID instead of a version name, it now returns a helpful 400 Bad Request with a clear message rather than a generic 404 Not Found.
- Tested the endpoint explicitly for the 400 status on missing params.

**Note:** I have checked the codebase and this API is not used anywhere internally (frontend or backend). It exists purely for external API consumers.


ArnavBallinCode (approved):

Saksham-Sirohi (approved):

app/eventyay/api/views/schedule.py
```diff
@@ -128,10 +128,25 @@ def get_object(self):
     @action(detail=False, url_path='by-version')
     def redirect_version(self, request, event, organizer=None):
         version = request.query_params.get('version')
-        schedule = get_object_or_404(self.event.schedules, version=version)
```
Saksham-Sirohi, line 132: This changes the error code to non-desired behaviour; please fix

Saksham-Sirohi (changes requested):

Rachit7168:
image

Working after the recent commit

Saksham-Sirohi (approved):

PR #4220, opened 2026-07-04
feat(checkin): list limits, popup fields, and gate-aware check-in

Saksham-Sirohi:
## Summary

- Per-day and per-gate check-in limits on lists
- Configurable success pop-up fields for the check-in app
- Gate-aware checkout-required errors and improved redeem API responses
- Control panel bulk check-in routed through `perform_checkin`

Closes #4219

## Migration

`0042_checkin_list_limits_and_popup_fields` (depends on `0041_make_event_email_optional`).

## Check-in stack

Part of a five-PR check-in stack. Each PR is **one commit off `dev`** with only its own changes — any branch can be checked out and tested independently (`python manage.py migrate` works on `dev`).

**Merge order:** #4216 → #4218 → **#4220** → #4222 → #4224

**Expected merge conflicts:**
- **#4220 + #4222** → `base/services/checkin.py` (keep both gate logic and admission enforcement)
- #4222 + #4224 → `api/serializers/order.py`

## Notes

No automated tests in this PR. Per-device list scoping is in #4218; not required to test limits/pop-up fields here.

## Testing guide

**Setup:** checkout this branch, `python manage.py migrate`. Event with at least one paid admission ticket and a check-in list.

1. **List settings (control panel)**
   - **Check-in → [list] → Settings**
   - Toggle **Limit to one check-in per day** and **Limit to one check-in per gate**.
   - Select **Check-in app display fields** (pop-up fields on successful check-in).
   - Save and reopen — values persist.

2. **Per-day limit**
   - Check in an attendee once on the list.
   - Attempt a second entry the same day → should be denied.

3. **Per-gate limit**
   - Assign different gates to two devices (or simulate gate on check-in).
   - Check in at gate A, then attempt entry at gate B without checkout → denied with checkout-required / cross-gate response.

4. **Checkout then re-entry**
   - Check out at gate A, then check in again at gate B → should succeed.

5. **Pop-up fields (check-in app / API)**
   - After check-in, confirm configured extra fields appear in the success response / pop-up payload.

6. **Bulk check-in (control panel)**
   - On the check-in list attendee view, select multiple unchecked attendees.
   - Bulk check-in → all should check in via the same rules as single check-in.

7. **API redeem**
   - `POST` check-in redeem for an already-checked-in ticket → confirm error code and `cross_gate` flag when applicable.

PR #4218, opened 2026-07-04
feat(checkin): device security profiles and control panel device setup

Saksham-Sirohi:
## Summary

- Restricted device security profiles (Check-In Staff, Badge Station)
- Improved device connect/edit UI and check-in list scoping per device
- CORS and middleware support for the check-in web app

Closes #4217

## Migration

`0036_device_checkin_lists_and_security` (depends on `0035`) — renumber to `0037` on rebase after #4216 merges.

## Check-in stack

Part of a five-PR check-in stack. Each PR is **one commit off `dev`** with only its own changes — any branch can be checked out and tested independently (`python manage.py migrate` works on `dev`).

PRs #4218–#4222 each add a `0036_*` migration also depending on `0035`. **Renumber to `0037`–`0039` during rebase** as earlier stack PRs merge.

**Merge order:** #4216 → **#4218** → #4220 → #4222 → #4224

**Expected merge conflicts:**
- #4220 + #4222 → `base/services/checkin.py`
- #4222 + #4224 → `api/serializers/order.py`

## Testing guide

**Setup:** checkout this branch, `python manage.py migrate`.

1. **Device profiles (control panel)**
   - **Organizer → Devices → Add device**
   - Confirm security profile choices: **Check-In Staff** and **Badge Station (kiosk)** with descriptions.
   - Create one device per profile.

2. **Check-in list scoping**
   - Edit a device: select events and specific **check-in lists**.
   - Save and confirm list checkboxes filter by selected events.

3. **Device connect flow**
   - Open the device connect URL / QR flow from the control panel.
   - Initialize the device in the check-in web app.
   - Confirm setup-token gate configure step works.

4. **Badge Station restrictions**
   - With a **Badge Station** device token, verify allowed: scan, check-in, search, badge download.
   - Verify **denied**: `POST /orders/`, `GET /products/` (no order creation or product listing).

5. **Check-In Staff**
   - With a **Check-In Staff** device token, verify check-in, search, attendee edit, and badge download work.
   - Verify **allowed** (for now): `GET /products/`, `POST /orders/`, `POST /orders/…/mark-paid/` for on-site registration. This may move to the embedded widget later.

6. **Regression**
   - `pytest tests/tickets/api/test_deviceauth.py -q`

PR #4216, opened 2026-07-04
perf(checkin): add trigram indexes for attendee search

Saksham-Sirohi:
## Summary

Adds PostgreSQL trigram indexes on order position attendee name, email, and order code fields so check-in list search is faster on large events (including partial order code lookup, e.g. `hmbvg`).

Closes #4215

## Migration

`0036_orderposition_search_trigram_indexes` (depends on `0035`) — lands as `0036` when merged first. Uses `AddIndexConcurrently` (no write locks during index build).

## Check-in stack

Part of a five-PR check-in stack. Each PR is **one commit off `dev`** with only its own changes — any branch can be checked out and tested independently (`python manage.py migrate` works on `dev`).

PRs #4218–#4222 each add a `0036_*` migration also depending on `0035`. **Renumber to `0037`–`0039` during rebase** as earlier stack PRs merge.

**Merge order:** #4216 → #4218 → #4220 → #4222 → #4224

**Expected merge conflicts:**
- #4220 + #4222 → `base/services/checkin.py`
- #4222 + #4224 → `api/serializers/order.py`

## Testing guide

**Setup:** checkout this branch, `python manage.py migrate` (PostgreSQL required).

1. **Migration**
   - Confirm migration applies without errors.
   - In PostgreSQL, verify indexes exist: `orderpos_name_trgm`, `orderpos_email_trgm`, `order_code_trgm` (all GIN / `gin_trgm_ops`).

2. **Control panel — check-in list search**
   - Open an event with existing orders/attendees.
   - Go to **Check-in → [list] → Attendees** and use the search box.
   - Search by partial **attendee name** (substring, not just prefix).
   - Search by partial **email**.
   - Search by partial **order code** (e.g. first few characters of a known code).

3. **Regression**
   - `pytest tests/tickets/control/test_search.py -q`

**What this PR does not change:** no API or UI behaviour — only query performance. Results should match `dev`; searches should feel the same, faster on large datasets.

PR #4222, opened 2026-07-04
feat(checkin): product admission validity and check-in enforcement

### LINKED ISSUE #4221: Product admission validity windows for check-in
## Summary

Organizers need to restrict when admission tickets can be checked in — for example workshop slots, day passes, or fixed time windows — and have that enforced consistently at the door.

## Expected change

- Configure admission validity per product (and variation): fixed times, event dates, or subevent dates with optional offsets
- Store the effective window on each sold ticket at purchase time
- Enforce validity during check-in (`invalid_time` when outside the window)
- Tickets sold before this feature existed should follow the product's current validity settings when no snapshot was stored
- Show validity in control panel, cart, order detail, and on PDF tickets
- Expose issued validity on check-in API position payloads

## Out of scope

- Live walk-in registration product filtering
- Per-gate / per-day list limits (separate change)
- Badge editing

Saksham-Sirohi:
## Summary

- Product and variation admission validity configuration in the control panel
- Purchase-time validity snapshot on order positions
- Check-in enforcement with catalog fallback when the issued snapshot is empty
- Validity display in cart, orders, PDFs, and the check-in API

Closes #4221

## Migration

`0051_admission_validity_fields` depends on `0050_checkin_list_limits_and_popup_fields` (#4220).

Verified after rebase onto latest `dev`:
- Chain: `0049` → `0050_checkin_list_limits_and_popup_fields` → `0051_admission_validity_fields`
- No duplicate migration numbers in `base`
- `makemigrations --check` reports no model/migration drift
- `manage.py check` reports no issues

## Issued-ticket semantics

- New positions snapshot catalog bounds at create time when configured.
- If either snapshot bound is set, check-in uses that frozen window (product edits do not change it).
- If both snapshot fields are empty (legacy positions or issued with no restriction), check-in falls back to the current product/variation catalog configuration.

## Variation semantics

- Default mode is `inherit` (same as product), with field-by-field overlay for set variation fields.
- Explicit empty mode = no check-in time restriction for that variation.

## Check-in stack

Part of the check-in stack. This PR is **one commit off `dev`**.

**Merge order:** #4216 → #4218 → #4220 → **#4222** → #4224

**Rebase notes (done):**
- Rebased onto `origin/dev` after #4220 merge
- Kept both gate/list-limit helpers from #4220 and admission validity enforcement in `base/services/checkin.py`
- Renumbered migration from conflicting `0050_*` to `0051_*`

**Expected merge conflicts with later stack PRs:**
- **#4222 + #4224** → `api/serializers/order.py`

## Testing guide

**Setup:** checkout this branch, `python manage.py migrate`. Event with admission product and check-in list.

1. **Product validity (control panel)**
   - **Products → [admission product] → Edit**
   - Set admission validity mode: **fixed dates**, **event dates**, or **subevent dates** (try at least fixed).
   - Save; configure a variation with inherit / overlay / explicit no-restriction.

2. **Purchase snapshot**
   - Place an order for the product (presale or backend).
   - On the order detail page, confirm **admission valid from/until** is shown on the position.
   - In DB or admin, confirm `admission_valid_from` / `admission_valid_until` are stored on the order position.

3. **Check-in enforcement (in window)**
   - With validity window that includes now, check in the attendee → success (exact boundaries allowed).

4. **Check-in enforcement (out of window)**
   - Use a product with validity entirely in the past (or wait until after `valid_until`).
   - Attempt check-in → denied with invalid-time error; `force=True` still works.

5. **Legacy / empty snapshot tickets**
   - Use an order position where snapshot fields are empty but the product has catalog validity configured.
   - Confirm check-in respects current catalog bounds.

6. **Display**
   - Presale cart / order confirmation shows validity window where applicable.
   - Ticket PDF includes validity text when configured.

7. **API**
   - Order position serializer includes `admission_valid_from` / `admission_valid_until` in check-in list responses.

## Local verification

- [x] Rebased onto latest `origin/dev`
- [x] Migration order/dependencies checked
- [x] Conflict in `checkin.py` resolved (gate limits + admission validity)
- [x] `python manage.py check` (test settings)
- [x] Admission validity tests in `tests/tickets/base/test_admission_validity.py`
- [ ] Manual UI screenshots (please add before merge if required by process reminder)


ArnavBallinCode (changes requested):
Overall, the feature direction makes sense, but I think there are a few correctness issues worth addressing before merging:

1. **`app/eventyay/base/admission_validity.py` - issued ticket snapshot can fall back to mutable product configuration**
   `get_issued_admission_bounds()` falls back to `resolve_catalog_admission_bounds()` whenever both stored bounds are `None`. This means a ticket that was originally issued with no admission restriction can become restricted later if the product configuration is changed. Since the stated goal of the `OrderPosition` fields is to snapshot the admission window at purchase time, the issued-ticket path should not fall back to the current catalog configuration once the position has been created. This also needs consideration for existing order positions when introducing the new fields.

2. **`app/eventyay/base/admission_validity.py` - variation overrides are currently all-or-nothing**
   `_variation_overrides_product()` treats any configured variation field as an override and then `_catalog_source()` uses the variation as the complete source. This means a variation that only sets, for example, an offset will stop inheriting the product's validity mode and other validity settings. Similarly, because the "no restriction" mode is represented by an empty string, a variation cannot explicitly override a product's restriction back to "no restriction". We should either merge variation fields with product values field-by-field or otherwise make the override semantics explicit and consistent.

3. **`app/eventyay/base/admission_validity.py` - validity offsets need stronger validation**
   The current validation only checks `offset_from > offset_until` when both values are set. Negative offsets and offsets extending beyond the underlying event/subevent window are still accepted. This can produce windows that start before the event or remain valid after the event has ended. Please validate the supported range/semantics explicitly and ensure the resulting window cannot become invalid or unintentionally extend beyond the event/date window.

4. **`app/eventyay/base/admission_validity.py` / `app/eventyay/base/models/orders.py` - snapshotting in `OrderPosition.save()` should be covered carefully**
   The snapshot is assigned whenever `not self.pk`. Since this is attached to the generic `OrderPosition.save()` path, please add tests around all creation paths, including products with variations, subevents, bundles/add-ons, and orders created/imported through non-standard flows. We should make sure `product`, `variation`, `subevent`, and `order.event` are always available before `assign_issued_admission_bounds()` runs.

5. **`app/eventyay/api/serializers/product.py` and `app/eventyay/control/forms/product.py` - validation logic is duplicated across layers**
   The new admission validity validation is now invoked from multiple serializers/forms, with some existing validation also being rewritten to first construct `full_data` from `to_representation()`. Please make sure this doesn't introduce inconsistent behavior between API updates and the control-panel forms, particularly for partial updates. Ideally the validation should have one clear source of truth and the tests should cover both paths.

6. **`app/eventyay/base/services/checkin.py` - add explicit check-in boundary tests**
   The enforcement uses strict `< valid_from` and `> valid_until` comparisons, meaning check-in exactly at either boundary is allowed. If that is intentional, please add tests documenting it. Also cover no restriction, fixed windows, event windows, subevent windows, offsets, expired tickets, not-yet-valid tickets, and `force=True` behavior.

7. **`app/eventyay/base/admission_validity.py` - existing tickets need a migration/backfill strategy**
   The migration adds the issued validity fields but does not populate them for existing `OrderPosition`s. Because the intended behavior is purchase-time snapshotting, relying on the current catalog configuration as a fallback makes historical tickets dependent on future product edits. Please define how existing positions should behave and preferably make that behavior explicit in the migration/data handling rather than silently resolving against today's product configuration.

8. **`app/eventyay/base/pdf.py` - font fallback behavior deserves tests**
   The new `_draw_textarea()` fallback can transliterate text with `text_unidecode` when the selected font doesn't support it, and then falls back to `AND`. This can alter attendee-visible text rather than simply selecting a compatible font. Please add coverage for multilingual/non-Latin text and verify that the fallback preserves the intended output. Also, catching a broad `Exception` around font registration/support detection makes failures harder to diagnose.

The most important issues for me are **the mutable-catalog fallback for issued tickets and the variation override semantics**. Those can change the effective validity of already-issued tickets or produce a different validity window from what the organizer configured.

Saksham-Sirohi:
@ArnavBallinCode added what made sense

Rachit7168 (approved):
LGTM , TESTED 

https://github.com/user-attachments/assets/194bad6c-dba4-45c0-b46d-6684f624a687

ArnavBallinCode (approved):
LGTM, tested

PR #4231, opened 2026-07-04
Fix : Check-in Log export to include all check-in records

### LINKED ISSUE #4228: Check-in Log should include all check-in records instead of only the latest one
The Check-in Log export currently only includes the latest check-in time for an attendee. If the same attendee is checked in multiple times, the previous check-in records are overwritten in the exported file.

However, when viewing the attendee's order in Orders → All Orders, the complete check-in history is still available and represented by multiple check marks next to the ticket. This indicates that all check-in records are stored, but the exported Check-in Log only includes the most recent one.

**Steps to Reproduce**

1. Sign in to access.eventyay.com.
2. Scan a ticket to check in an attendee.
3. Scan the same ticket multiple times.
4. Go to Orders → Export and export the Check-in Log.
5. Observe that the exported file only contains the latest check-in time for the attendee.
6. Go to Orders → All Orders and open the attendee's order.
7. Notice that multiple check marks are shown next to the ticket, indicating that multiple check-in records exist.

Image

**Expected Behavior**

The exported Check-in Log should include all check-in records for an attendee, preserving every check-in timestamp instead of only the latest one.

Rachit7168:
Close #4228

## Description
This PR fixes the Check-in Log (CSV export) so that it includes **all** check-in records for an attendee instead of only the most recent one.

### Why this change is needed
Previously, the Check-in List export fetched the latest check-in time for each ticket using a `Max('datetime')` subquery. If the same attendee was checked in multiple times, the previous check-in records were overwritten in the exported file. However, the system stores all check-ins (as seen in Orders → All Orders, where the complete check-in history is represented by multiple check marks). Users rely on the export to analyze total check-ins, so all historical scans must be included.

### What changed
- Refactored `CSVCheckinList` to iterate over all check-in entries (`Checkin` models) rather than using a subquery that groups by position.
- Added a `_csv_row_for_position` helper method to yield individual rows for each checkin record.
- Optimized `CheckinLogList` queryset to filter by `list__event` for performance.
- Added test coverage (`test_csv_includes_all_checkins` and `test_checkinlog_includes_all_checkins`) to verify that all check-in events are exported properly.

https://github.com/user-attachments/assets/a0677b06-eeeb-4fcc-8f05-13e2bb70cc08


ArnavBallinCode (approved):
LGTM

hpdang (approved):

PR #4240, opened 2026-07-05
fix: correctly map get_duration field in base SubmissionSerializer for team token auth (#4156)

### LINKED ISSUE #4156: POST /submissions/ fails with team token (wrong serializer, get_duration, missing event)
## Summary

`POST /api/v1/organizers/{org}/events/{event}/submissions/` returns **500** when authenticated with a **team API token**, even with a valid JSON body. The view uses `SubmissionSerializer` instead of `SubmissionOrgaSerializer`, so create logic that sets `event` and maps `duration` never runs.

User API tokens and session auth (organizer user) use `SubmissionOrgaSerializer` and work in tests. Team tokens authenticate as `AnonymousUser`, so `is_orga` is false and the broken code path is used.

https://github.com/user-attachments/assets/ece7956f-c2cc-43bb-9974-5bd9929ce025

Ujj-CodeX:
Fixes #4156

## Problem

`POST /api/v1/organizers/{org}/events/{event}/submissions/` returned a 500 
error when authenticated with a team API token, even with a valid request body.

## Root cause

The `duration` field on `SubmissionSerializer` is declared with 
`source='get_duration'`. Because of this, DRF stores the value under the 
key `get_duration` in `validated_data` instead of `duration`.

`SubmissionOrgaSerializer` (used for session/user-token auth) overrides 
`create()`/`update()` to remap `get_duration` back to `duration` before 
saving. However, team API tokens authenticate as `AnonymousUser` 
(`TeamTokenAuthentication.authenticate_credentials`), which makes 
`is_orga` evaluate to `False` in `SubmissionViewSet`. This routes the 
request through the base `SubmissionSerializer`, which had no such 
remapping and relied on DRF's default `ModelSerializer.create()`. 
Passing `get_duration` straight to `Submission.objects.create()` then 
raised:

## Fix

Moved the common `create()`/`update()` logic (tags, image, event, and the 
`get_duration` → `duration` remapping) into the base `SubmissionSerializer`, 
since `SubmissionOrgaSerializer` inherits from it. `SubmissionOrgaSerializer` 
now only adds its orga-specific side effects (`update_duration()`, 
`update_talk_slots()`, `update_review_scores()`) on top of the shared logic.

This means the fix applies uniformly regardless of auth method (team token, 
user token, or session auth), without duplicating logic between the two 
serializers.

## Verification

Manually reproduced and verified via Postman against a local dev instance:

- **Before fix:** `POST .../submissions/` with a team token → `500 Internal 
  Server Error` (`TypeError: Submission() got unexpected keyword arguments: 
  'get_duration'`)
- **After fix:** same request → `201 Created`, submission created with 
  correct `duration` value

## Note on automated tests

I attempted to add a test covering this scenario, but ran into a 
pre-existing, unrelated issue in the `organiser`/`team` fixtures in 
`tests/talk/conftest.py` (`TypeError: Team() got unexpected keyword 
arguments`), suggesting a mismatch with the current `Team` model. Happy 
to open a separate issue for that if useful.
support

## Summary by Sourcery

Bug Fixes:
- Fix submission creation with team API tokens by correctly mapping the serializer's get_duration field to the model's duration field before saving.


Rachit7168:
Closing this PR in favor of #4282.

Please avoid working on pre-assigned issues without prior discussion or assignment.

Thank you for your contribution! I look forward to seeing more contributions from you in the future.

Ujj-CodeX:
> Closing this PR in favor of #4282.
> 
> Please avoid working on pre-assigned issues without prior discussion or assignment.
> 
> Thank you for your contribution! I look forward to seeing more contributions from you in the future.

Thanks for letting me know. I overlooked that the issue was already self-assigned. I'll make sure to check assignments before working on such issues in the future. Looking forward to contributing to other issues!

PR #4165, opened 2026-06-30
fix: unified import process ui code for tickets & talks import

ViRUS-0-0:
## Closes #3741 

> [!NOTE]
Celery has been turned off to show case the ui, as import small csv would finish the ui pop up very fast.

### Speaker's import process ui

https://github.com/user-attachments/assets/b00365ec-b74f-4753-a59f-4dd58ee459e7

---
### Schedule import process ui

https://github.com/user-attachments/assets/5725acc2-3a0d-47cd-9ce2-0e514b41ba46

---
### Tickets import process ui

Screenshot 2026-07-04 at 12 03 53 PM

---

https://github.com/user-attachments/assets/14b57b3d-7a8b-4b80-a79f-9439f700bcb0

---

PR #4267, opened 2026-07-05
fix: unified import process ui code for tickets & talks import 

### LINKED ISSUE #3741: The export processing UI is inconsistent across different event import and export workflows.
## Describe the bug

The export processing UI is inconsistent across different event export workflows.

For ticket exports, the system displays a full-page processing screen with a centered loading state while the export is being generated.

For talk exports, the system displays a modal dialog with a progress bar overlaying the current page.

Since both exports perform the same asynchronous export operation, the user experience should be consistent across all event export modules. Having different loading and processing interfaces can be confusing and makes the platform feel inconsistent.

## Expected behavior

All event export workflows should use the same processing UI and follow a unified export experience.

Suggested behavior:

- Use a single standardized export processing component across the platform.
- Display the same loading/progress interface for:
  - Ticket related exports
  - Talk related exports
- Ensure consistent messaging, styling, and progress indication throughout the application.
- Users should receive the same visual feedback regardless of which export type they initiate.

### Talks related Export Processing UI
(Current implementation uses a full-page loading screen)

Image

### Ticket related Export Processing UI
(Current implementation uses a modal with a progress bar)

Image

## Additional context

This appears to be a UI consistency issue rather than a functional bug.

The export generation process is conceptually the same across modules, but different loading/progress components are being used. Standardizing the export processing experience would improve usability and provide a more cohesive user experience throughout the event management platform.

ViRUS-0-0:
Close #3741 

Re-implements https://github.com/fossasia/eventyay/pull/4165

Screenshot 2026-07-07 at 12 19 40 PM

---

Tickets Import

https://github.com/user-attachments/assets/842078d5-7793-450a-a1ee-19fc5b167ab3

----

Speaker imports

https://github.com/user-attachments/assets/e264a88e-d9f7-4149-8b23-afe9249287df

----

Schedule Imorts

https://github.com/user-attachments/assets/f697bd1e-4aa5-4d59-97bb-b99c76b41d0e

---

Exports works as usual

https://github.com/user-attachments/assets/4e80d8fc-f5a9-44a0-8764-7a7e4fe43b29


ViRUS-0-0:
This PR needs a little more time than expected.

app/eventyay/base/services/orderimport.py
```diff
@@ -84,11 +84,14 @@ def import_orders(self, event: Event, fileid: str, settings: dict, locale: str,
     user = User.objects.get(pk=user)
     with language(locale, event.settings.region):
         cols = get_all_columns(event, settings)
-        parsed = parse_csv(cf.file)
+        parsed = list(parse_csv(cf.file))
```
Rachit7168: The parse_csv function can return None if the CSV file is completely invalid or cannot be sniffed. If parse_csv(cf.file) returns None, list(None) will raise a TypeError: 'NoneType' object is not iterable and crash the Celery background task unexpectedly.

Consider explicitly checking for None before converting it to a list

app/eventyay/base/services/talkimport.py
```diff
@@ -627,16 +627,20 @@ def import_speakers(event: Event, fileid: str, settings: dict, locale: str, user
         acting_user = User.objects.get(pk=user_id)
         with language(locale, event.settings.region):
             with scope(event=event):
-                parsed = parse_csv(cf.file, django_settings.MAX_SIZE_CONFIG[SizeKey.UPLOAD_SIZE_CSV])
-                if parsed is None:
+                parsed = list(parse_csv(cf.file, django_settings.MAX_SIZE_CONFIG[SizeKey.UPLOAD_SIZE_CSV]) or [])
+                if not parsed:
```
Rachit7168: Changing if parsed is None: to if not parsed: introduces a subtle bug. If the user uploads a valid but empty CSV file (or one with only headers and no data), parsed will be evaluated as [], and not parsed will become True. This will incorrectly throw an ImportExecutionError saying 'Could not parse the CSV file' because it treats an empty file the same as a parsing failure.

To correctly catch parsing failures while still evaluating the generator into a list, you should revert to checking None explicitly

Rachit7168 (changes requested):
Please address

app/eventyay/base/services/talkimport.py
ViRUS-0-0: fixed

app/eventyay/base/services/orderimport.py
ViRUS-0-0: fixed

app/eventyay/static/eventyay-common/js/asynctask.js
```diff
@@ -0,0 +1,263 @@
...
+    body.append('ajax', '1')
+
+    fetch(form.action || location.href, {
+        method: 'POST',
+        body,
+        headers: { 'X-Requested-With': 'XMLHttpRequest' },
+    })
+        .then(async r => {
+            const contentType = r.headers.get('content-type')
+            if (contentType && contentType.includes('text/html')) {
```
Rachit7168, line 165: If the import form fails validation (e.g., missing file, bad CSV format), the server responds with the HTML of the page displaying the error. Replacing document.documentElement with a new DOMParser document destroys the active DOM tree. This wipes out all event listeners on the document and window objects (dropdowns, modals, other JS) and prevents any scripts in the new HTML from executing. The page will freeze and become completely unresponsive. Fix: You should extract just the <form> element from the new document and replace the old form (similar to how pretixbase/js/asynctask.js does it), or extract and replace the main content container (e.g., #page-wrapper).

app/eventyay/static/eventyay-common/js/asynctask.js
```diff
@@ -0,0 +1,263 @@
...
+    const progress = el.querySelector('.progress')
+    const bar = el.querySelector('.progress-bar')
+    if (progress) progress.style.display = ''
+    if (bar) {
+        bar.classList.remove('progress-bar-striped', 'active')
+        bar.style.width = pct + '%'
+    }
+}
+
+const poll = () => {
```
Rachit7168, line 82: When an import task completes successfully, the server returns a redirect URL in the JSON response to take the user to the import summary/results page. You are completely ignoring data.redirect in the poll() function. When the modal hides, the user is left stranded on the import start page instead of being redirected to see the results. Fix: After the 2-second timeout, you must check for data.redirect and navigate

Rachit7168 (changes requested):
Please address : 
 Singleton #loadingmodal conflict with normal async tasks You are reusing the global #loadingmodal DOM element for the background import task. This causes severe conflicts if the user minimizes an import task and then triggers a regular async task (like a ticket export) from another part of the UI while the import is still polling:

The two scripts will simultaneously poll and overwrite each other's progress bars and status text inside the same #loadingmodal.
Because body still has .loading-minimized, the new export task will be shown shrunk in the bottom right corner.
Because the new task is not an import task, .is-import-task is missing from body, so your CSS rule (body:not(.is-import-task) #loadingmodal .btn.loadingmodal-minimize { display: none; }) hides the maximize button. The user is stuck with a minimized, un-maximizable loading modal for a blocking task. Fix: Do not reuse the global #loadingmodal. Instead, clone it or dynamically generate a separate DOM element (e.g., #import-toast) specifically for the background import task. This ensures background imports do not clobber foreground blocking tasks.

app/eventyay/static/eventyay-common/js/asynctask.js
ViRUS-0-0: fixed

app/eventyay/static/eventyay-common/js/asynctask.js
ViRUS-0-0: fixed

Rachit7168 (approved):
LGTM , TESTED 

https://github.com/user-attachments/assets/ff66e219-f42b-49b1-ad74-5ffc40da7aa1

app/eventyay/static/eventyay-common/js/asynctask.js
```diff
@@ -0,0 +1,283 @@
...
+                    bigIcon.style.color = '#5cb85c'
+                }
+
+                sessionStorage.removeItem('eventyay_async_task_import')
+
+                setTimeout(() => {
+                    hide()
+                    // if (data.redirect) {
+                    //     location.href = data.redirect
+                    // }
```
Aqil-Ahmad: why this is commented out?

app/eventyay/static/eventyay-common/js/asynctask.js
```diff
@@ -0,0 +1,283 @@
...
+    fetch(checkUrl, { headers: { 'X-Requested-With': 'XMLHttpRequest' } })
+        .then(r => {
+            if (!r.ok) {
+                throw new Error('Network response was not ok')
+            }
+            return r.json()
+        })
+        .then(data => {
+            if (data.ready) {
+                const bar = document.querySelector('#import-loadingmodal .progress-bar')
```
Aqil-Ahmad, line 92: data.ready branch never checks data.success. Even a failed import shows the green checkmark / "Task completed" text, since success/error states aren't distinguished.

app/eventyay/static/eventyay-common/js/asynctask.js
```diff
@@ -0,0 +1,283 @@
...
+        body,
+        headers: { 'X-Requested-With': 'XMLHttpRequest' },
+    })
+        .then(async r => {
+            const contentType = r.headers.get('content-type')
+            if (contentType && contentType.includes('text/html')) {
+                const html = await r.text()
+                const newDoc = new DOMParser().parseFromString(html, 'text/html')
+                const newWrapper = newDoc.querySelector('#page-wrapper')
+                const oldWrapper = document.querySelector('#page-wrapper')
```
Aqil-Ahmad, line 169: when the server responds with text/html, the code swaps #page-wrapper but never calls hide(). isSubmitting stays true forever (blocking all future submits) and the fullscreen loading overlay never disappears

app/eventyay/static/eventyay-common/css/loading_modal.css
```diff
@@ -0,0 +1,442 @@
...
+    -o-animation: progress-bar-stripes 3s linear infinite;
+    animation: progress-bar-stripes 3s linear infinite;
+}
+
+body:not(.is-import-task) #loadingmodal .btn.loadingmodal-minimize {
+    display: none;
+}
+
+
+#import-loadingmodal, #ajaxerr {
```
Aqil-Ahmad: loadingmodal and import-loadingmodal seems almost similar. a lot of css is duplicated here.

app/eventyay/static/eventyay-common/js/asynctask.js
Aqil-Ahmad, line 1: we already have have asynctask.js in pretixbase/js, it seems we are reimplementing things here. can you confirm if we can reuse what is already there or maybe extract shared logic

Aqil-Ahmad (changes requested):

app/eventyay/static/eventyay-common/js/asynctask.js
ViRUS-0-0: it is for tickets imports and export. the asynctask.js introduced in this pr is unified for both tickets and talk imports only

app/eventyay/static/eventyay-common/css/loading_modal.css
ViRUS-0-0: fixed

app/eventyay/static/eventyay-common/js/asynctask.js
ViRUS-0-0: fixed

app/eventyay/static/eventyay-common/js/asynctask.js
ViRUS-0-0: fixed

app/eventyay/static/eventyay-common/js/asynctask.js
ViRUS-0-0: fixed

prayag78 (approved):

Aqil-Ahmad (approved):

Rachit7168 (approved):

Saksham-Sirohi (approved):

app/eventyay/static/eventyay-common/js/asynctask.js
```diff
@@ -0,0 +1,290 @@
...
+                    if (bigIcon) {
+                        bigIcon.className = 'fa fa-times big-rotating-icon'
+                        bigIcon.style.animation = 'none'
+                        bigIcon.style.color = '#d9534f'
+                    }
+                }
+
+                sessionStorage.removeItem('eventyay_async_task_import')
+
+                setTimeout(() => {
```
Rachit7168, line 124: When `data.ready` is true, this logic simply hides the modal after 2 seconds and returns. It completely ignores `data.redirect` which the backend `AsyncAction` tasks return upon successful completion. As a result, the user is left stranded on the import form page and has to manually refresh to see their imported data.

Please check for `data.redirect` inside this block so it correctly redirects to the success/summary page (similar to how the original `asynctask.js` did). For example:
```javascript
                setTimeout(() => {
                    hide()
                    if (data.redirect) {
                        location.href = data.redirect;
                    } else {
                        location.reload();
                    }
                }, 2000)
```

Rachit7168 (changes requested):
This PR does an excellent job centralizing the import polling logic, updating background tasks to send progressive updates, and refactoring the SCSS to standard CSS for a unified loading modal. However, I found a critical bug in the new JavaScript logic that will disrupt the user workflow after an import completes. Please address the redirect issue, and then this PR will be good to go!

Saksham-Sirohi (commented):

- Import tasks now report Celery progress and share a unified loading modal via `data-import-task` + ES module helper.
- CSV parse failures raise explicit import errors instead of failing later — good hardening.

app/eventyay/static/eventyay-common/js/asynctask.js
ViRUS-0-0: This was implemented to preserve continuity. If the modal is minimised while the user is working on something else, they shouldn't be redirected to the import page, as it would interrupt their current workflow.

Rachit7168 (approved):
LGTM , TESTED AGAIN ITS NOT BREAKING THE EXPORT LIKE LAST TIME

ArnavBallinCode (approved):
LGTM, Lets test on dev

PR #4282, opened 2026-07-07
fix: resolve bugs with team API tokens for submissions and answers

### LINKED ISSUE #4155: Answer POST crashes on create and upsert returns wrong response body
## Summary

The Talk **Answers API** (`POST .../answers/`) has two related bugs:

1. **POST always crashes (500)** — `AnswerCreateSerializer` references `self.request`, which does not exist.
2. **Upsert response is wrong (201)** — `perform_create` uses `update_or_create` but never sets `serializer.instance`, so the JSON response can have `id: null` even when the DB row was updated correctly.

Bug 1 blocks all answer creation via the API. Bug 2 affects duplicate POSTs (same question + submission) once Bug 1 is fixed.

---

Image

Image

### LINKED ISSUE #4156: POST /submissions/ fails with team token (wrong serializer, get_duration, missing event)

Rachit7168:
Fixes #4156 
Fixes #4155

**What was changed:**
1. Moved `create` and `update` logic from `SubmissionOrgaSerializer` to `SubmissionSerializer`.
2. Changed `self.request` to `request` in `AnswerCreateSerializer.__init__`.
3. Set `serializer.instance = answer` in `AnswerViewSet.perform_create`.
4. Updated `questions_for_user` in `talk_rules/submission.py` to check `request.eventpermset` in addition to `user.has_perm`.

**Why it was changed:**
1. **Submission Creation (#4156)**: Team API tokens authenticate as `AnonymousUser`, causing them to use the base `SubmissionSerializer` instead of the `SubmissionOrgaSerializer`. The base serializer lacked `create`/`update` methods to map `get_duration`, tags, images, and the event context, leading to 500 errors.
2. **Answer Creation Crash**: `AnswerCreateSerializer.__init__` was referencing `self.request` which didn't exist (it should extract it from `self.context.get("request")`). This caused a 500 error when team tokens tried to answer custom questions.
3. **Answer Upsert Null ID**: When upserting an answer, `update_or_create` successfully updated the row, but the DB ID was returning as `null` in the 201 response. Associating the updated `answer` object to `serializer.instance` fixes this.
4. **Answer Question Validation**: Because team tokens authenticate as `AnonymousUser`, `questions_for_user` filtered out all non-public questions because it relied on `user.has_perm`. Adding a check for `request.eventpermset` allows team tokens with event permissions to see and answer these questions, preventing the `Invalid pk - object does not exist` 400 Bad Request error.

https://github.com/user-attachments/assets/903a5e57-9ec6-4c97-9f2c-b1c6490ac4dc

https://github.com/user-attachments/assets/9dff11cd-080e-4b4b-ab29-8491277de12f

## Summary by Sourcery

Fix submission and answer handling to work correctly with team API tokens and event-scoped permissions.

Bug Fixes:
- Ensure submission create/update logic (duration, tags, images, event and locale) is applied for all submission serializers, including those used with anonymous or token-based requests.
- Fix answer creation serializer to use the request from the DRF context when resolving available questions.
- Return a non-null database ID in answer creation responses by associating the created or updated answer with the serializer instance.
- Allow users authenticated via event permission sets (e.g., team API tokens) to see and answer non-public talk questions when they have appropriate event-level permissions.


Saksham-Sirohi (changes requested):
duplicated permission branches in questions_for_user

Aqil-Ahmad (approved):

app/eventyay/api/serializers/submission.py
```diff
@@ -247,6 +247,55 @@ def to_representation(self, instance):
...
+        if 'get_duration' in validated_data:
+            validated_data['duration'] = validated_data.pop('get_duration')
+            duration_changed = validated_data['duration'] != instance.duration
+        slot_count_changed = 'slot_count' in validated_data and validated_data.get('slot_count') != instance.slot_count
+        track_changed = 'track' in validated_data and validated_data.get('track') != instance.track
+
+        submission = super().update(instance, validated_data)
+
+        if tags_data:
+            submission.tags.set(tags_data)
```
ArnavBallinCode: Since you've already checked `'slot_count' in validated_data`, you can directly access the value instead of using `.get()`.

```python
slot_count_changed = (
    'slot_count' in validated_data
    and validated_data['slot_count'] != instance.slot_count
)
```

This keeps the access pattern consistent and avoids an unnecessary `.get()`.

ArnavBallinCode (changes requested):

prayag78 (approved):

Saksham-Sirohi (commented):

- Fixes `questions_for_user(self.request, ...)` → `questions_for_user(request, ...)`, which would break Answer API queryset setup for team tokens.
- Team API permission expansion and submission create/update guards look intentional; uses specific `ValidationError` types.

Saksham-Sirohi (changes requested):
Thanks for tackling the team-token auth gap — the root cause analysis in the PR description is accurate, and the four fixes are directionally correct. Requesting a few changes before merge to make this complete and regression-safe.

## Required

### 1. Add `TeamAPIToken` regression tests

Existing submission/answer API tests use `UserApiToken` fixtures (`orga_user_write_token`), which authenticate as a real user. That path never hits the bugs this PR fixes (`AnonymousUser` + `request.eventpermset`).

Please add tests that exercise **`TeamAPIToken`** auth specifically:

- POST `/submissions/` with a team token that has `can_change_submissions` — assert 201, correct `duration`/tags/event assignment (covers #4156)
- POST `/answers/` against a **non-public** question with a team token — assert 201, not 400 "Invalid pk" (covers #4155)
- POST duplicate answer (upsert) — assert response `id` is non-null (covers the `serializer.instance` fix)
- Team token on a read-only team — assert 403 on write (confirms we didn't over-open permissions)

Without these, the fix relies entirely on manual verification.

### 2. Fix `SubmissionViewSet.is_orga` to respect `eventpermset`

Moving `create`/`update` to `SubmissionSerializer` fixes the 500, but team tokens still select the base serializer because `is_orga` only checks `request.user.has_perm`:

```python
def is_orga(self):
    return self.event and self.request.user.has_perm('base.orga_list_submission', self.event)
```

Team tokens are always `AnonymousUser`, so they still miss:
- Orga-only response fields (`notes`, `internal_notes`, `pending_state`, …)
- Orga-only validators (`validate_content_locale`, `validate_slot_count`)

Please update `is_orga` to also check `request.eventpermset`, e.g.:

```python
@cached_property
def is_orga(self):
    if not self.event:
        return False
    if self.request.user.has_perm('base.orga_list_submission', self.event):
        return True
    eventpermset = getattr(self.request, 'eventpermset', set())
    return 'can_change_submissions' in eventpermset
```

This aligns serializer selection with the `questions_for_user` / `EventPermission` model already used elsewhere (e.g. checkin, orders). Once `is_orga` is fixed, the hoisted `create`/`update` on the base class still helps as shared logic (DRY via inheritance), but team tokens get the full orga serializer surface.

## Optional (nice to have)

- If you prefer **not** to change `is_orga`, move `validate_content_locale` and `validate_slot_count` to `SubmissionSerializer` so team-token writes can't bypass them.
- Rebase to drop the ~35 merge commits — makes the diff much easier to review.

## What looks good

- `self.request` → `request` in `AnswerCreateSerializer.__init__` — clear bug fix
- `serializer.instance = answer` — correct DRF pattern for manual `update_or_create`
- `questions_for_user` + `eventpermset` — scoped correctly (`can_change_submissions` / `can_change_event_settings`), safe fallback via `getattr(..., set())`
- `if not self.event` guard on create/update — good defensive check
- No scheduling or unrelated areas touched

Happy to re-review once the tests and `is_orga` update are in.

Saksham-Sirohi (approved):

ArnavBallinCode (approved):

PR #4344, opened 2026-07-12
fix: correct review settings layout and validation for dynamic scores

### LINKED ISSUE #4158: UX/UI Flaws and Missing Validations in Review Score Categories
## Description

There are multiple user experience and validation issues on the **Review settings -> Review scores** page when adding, modifying, or saving score categories.

## Steps to Reproduce

1. Navigate to **Settings -> Review -> Review scores**.
2. Click on `+ Add another score category` (or use the `+` button next to existing score inputs).
3. Try to leave newly created score fields entirely blank or enter unexpected characters (like emojis/strings).
4. Try to find a way to delete a newly added or existing score row.
5. Click **Save**.

---

### 1. Missing Delete Button for Individual Score Rows

* **Issue:** When adding a new score value field using the `+` icon, there is no corresponding `-` or trash icon next to the fields to remove them.
* **Impact:** Users are forced to leave fields empty if they change their mind, which leads to the next validation bugs.

### 2. Missing Form Validation & Silent Drop on Empty Fields

* **Issue:** If a user adds a score row, leaves both the *Score* (numeric) and *Label* fields completely blank, and clicks **Save**, the system gives no validation error. However, the empty row is silently dropped and not saved.
* **Expected Behavior:** The form shouldn't silently discard rows; it should either prevent submission with a clear inline validation message or not generate entirely empty rows without data.

### 3. Non-Descriptive Global Error Banner

* **Issue:** When an actual validation error triggers (e.g., leaving a score empty but adding a text label like "12"), a generic flash message appears at the top: *"We had trouble saving your input - Please see below for details."*
* **Impact:** There are **no inline details** or highlighted fields below. The user has to guess what went wrong.

### 4. Lack of Strict Input Typing (Data Integrity)

* **Issue:** The "Scores" number fields accept invalid input formats. In the video, the field accepted an emoji (👍) and strings, which then triggered the broken validation workflow upon saving.
* **Expected Behavior:** The numeric score field should strictly enforce `type="number"` or equivalent validation to reject emojis and non-numeric characters before submission.

---

https://github.com/user-attachments/assets/95a9d409-3ae2-40d7-b935-287515122ea4

### LINKED ISSUE #4346: HTTP 500 Error when saving Review Settings with an empty "Review score category" name
## Description

When attempting to add or modify score categories in the **Review settings** dashboard, leaving the **Review score category** name field blank (even if the validation error "This field is required" triggers locally) can cause the entire application to crash with an **HTTP 500** internal server error upon hitting the "Save" button.

---

## Steps to Reproduce

1. Navigate to the **Open Source Connect** event dashboard -> **Settings** -> **Review** tab.
2. Click on the **Review phases** or **Review scores** configuration sub-tabs.
3. Scroll down to the score category section and click **"+ Add another score category"**.
4. Fill in standard options like `Weight`, `Scores` values, or check/uncheck status boxes.
5. **Leave the language-specific "Review score category" text input field completely empty** (e.g., under English, German, or other language tabs).
6. A red inline validation flag reading `"This field is required."` appears.
7. Disregard the validation text and click the **Save** button at the bottom of the form.

## Expected Behavior

The form should block submission client-side, highlight the missing required fields, and prevent the request from hitting the server until the mandatory "Review score category" strings are provided.

## Actual Behavior

The form submits, the application routes the incomplete payload, and the page crashes completely, displaying an **"HTTP 500: We have encountered an error"** page.

---

https://github.com/user-attachments/assets/6f9145c5-b112-4f56-b40a-b25714ed8c33

Rachit7168:
## Summary

This PR fixes UI and validation issues on the Review Settings page related to dynamic score rows, ensuring consistent layout and robust form handling.

## Why

When adding or modifying review score categories, validation errors on dynamically added score rows were silently dropped due to incomplete backend validation and missing template rendering. Additionally, the inline validation error messages artificially expanded the height of the row, disrupting the page layout. Finally, dynamically added score rows that were saved lost their delete button, preventing users from removing them later.

## Changes

* Implemented `clean()` on `ReviewScoreCategoryForm` to correctly validate dynamic and existing score fields.
* Ensured error messages for missing or invalid score and label pairs are passed to the template.
* Re-styled the inline validation errors using absolute positioning to act as tooltips, preventing the expanding box layout shift.
* Rebound the JavaScript delete button event handlers effectively so they attach to newly added rows dynamically.
* Removed hardcoded placeholders and removed the `required` attribute from dynamic inputs to allow empty row deletion logic to function correctly.
* Reverted the delete button logic to protect the first three default score rows (0, 1, 2) from being deleted while ensuring the delete button persists for any additional rows added and saved.
* Added a standard save success message when the Review Settings form is successfully submitted.

## Review feedback follow-ups

* Unified `clean()` so existing and dynamically added rows validate identically: a fully empty row is treated as removable, and a partially filled row consistently reports the missing field.
* Replaced the `:has()` CSS selector with a `.has-error` class (added in the template) for the invalid-input highlight, since `:has()` is not supported in some browsers.

## Fix: HTTP 500 when saving Review Settings (#4346)

While testing the above, saving the form could crash with an **HTTP 500** (`ValueError: The ReviewScoreCategory could not be changed because the data didn't validate`). Root causes and fixes:

* The extra-forms filters used `form.has_changed` (a bound method reference, always truthy) instead of `form.has_changed()`, so **empty added categories/phases were force-saved as blank rows** with a required-but-empty name, which then failed validation on the next save. Now empty added rows are correctly skipped.
* The initial-forms loop saved score categories flagged for deletion. Because the formset skips validating deleted forms, such a form can still carry errors and raise a `ValueError` in `save()`. Deleted forms are now skipped in the save loop (deletion is handled by the `deleted_forms` loop).
* `clearOldNewScores()` rebuilt the unsaved-row id list from the delete buttons, dropping a row's id whenever it re-rendered without one and silently skipping its validation on the next submit. It now derives the ids from the score inputs directly, keeping error highlighting and the delete button consistent across re-renders.

## Testing

* Verified dynamic score rows are successfully saved when complete.
* Verified appropriate error messages are shown for incomplete score rows (missing label or value) and block saving.
* Verified that error tooltips float cleanly over the page without expanding row height.
* Verified that the first three default scores lack delete buttons, while subsequent rows correctly show them.
* Confirmed the save success banner triggers upon successful form submission.
* Verified adding a category and saving twice no longer produces a 500, and empty added rows are no longer persisted as blank categories.

## Issue

Fixes #4158
Fixes #4346

https://github.com/user-attachments/assets/e5347914-ee71-42b8-9ff3-bcb53233d9d1


app/eventyay/orga/forms/event.py
```diff
@@ -421,13 +423,45 @@ def __init__(self, *args, event=None, **kwargs):
...
+    def clean(self):
+        data = super().clean()
+        for score_id in self.new_label_ids:
+            value = data.get(f'value_{score_id}')
+            label = data.get(f'label_{score_id}')
+
+            if (value is None or value == '') and not label:
+                self.add_error(f'value_{score_id}', _('Please provide a score and a label, or remove this row.'))
+            elif (value is None or value == '') and label:
+                self.add_error(f'value_{score_id}', _('Please provide a numeric score.'))
```
ArnavBallinCode: Should completely empty existing rows be treated as deletion, or should they be rejected exactly like new rows? behaviour here is inconsistent

app/eventyay/orga/templates/orga/settings/review.html
```diff
@@ -146,15 +145,28 @@ <h2>{% translate "Review settings" %}</h2>
...
                                             <div class="score-label">
                                                 {{ label_pair.1 }}
+                                                {% if label_pair.1.errors %}
+                                                    <div class="score-error-tooltip" role="alert">
+                                                        {% for error in label_pair.1.errors %}{{ error }} {% endfor %}
+                                                    </div>
+                                                {% endif %}
                                             </div>
+                                            {% if 'new' in label_pair.2|stringformat:"s" or forloop.counter > 3 %}
+                                                <div role="button" class="delete-score btn btn-danger flip ml-2 align-self-start" data-score="{{ label_pair.2 }}"><i class="fa fa-trash"></i></div>
```
ArnavBallinCode: delete button only appears if greater than 3?

app/eventyay/static/orga/css/_layout.css
```diff
@@ -890,6 +890,45 @@ table .color-square {
...
 }
 
+/* Score row inline validation: render the message as an overlay bubble so it
+   never grows the row height or expands the surrounding card. */
+.score-score,
+.score-label {
+  position: relative;
+}
+
+.score-score:has(.score-error-tooltip) input,
```
ArnavBallinCode: `:has() `is not supported in some browsers.

app/eventyay/orga/templates/orga/settings/review.html
Rachit7168: yes , the first 3 values should not be allowed to delete setted as default

Rachit7168: Needs to be discussed once

Rachit7168:
@codex review

Aqil-Ahmad (approved):

prayag78 (approved):

Rachit7168:
image

Issue #4383, opened 2026-07-12
Feature: Organizer Analytics Dashboard


prayag78:
## Goal
Provide organizers with a unified analytics page accessible from the organizer sidebar, so they can quickly understand how their events are performing across registrations, ticket sales, proposals, check-ins, email engagement, follower growth etc.

## Current Behavior
- Analytics in Eventyay are scattered across isolated, per-event locations.
- Ticket and order statistics live inside the per-event Statistics plugin and show orders-by-day, revenue-over-time, and orders-by-product using Morris.js charts.
- Talk and proposal statistics live inside the per-event Orga module and show proposal pipeline, state, type, and track breakdowns using ApexCharts.
- Admin-wide analytics exist only in the super-admin dashboard and are not visible to regular organizers.
- The organizer sidebar currently has no analytics entry.

## Expected Behavior
- Organizers can access a new "Analytics" section directly from their sidebar dashboard menu.
- The system presents a unified analytics view showing aggregated metrics across all permitted events belonging to the organizer.
- Charts load efficiently using a mixed library approach: Morris.js for line/area trends and ApexCharts for donuts/pie charts.
- Event permissions are strictly enforced so organizers only see aggregate metrics for events they actually have permission to view.

## Features
- Direct navigation entry in the organizer sidebar menu via the common navigation configuration.
- Scoped event permission check filtering events through the user's permitted event list.
- Ticket Sales & Revenue section showing orders over time, status breakdown, cumulative revenue and top events by sales.
- Proposals & Sessions section showing proposal pipeline states, submission timeline based on LogEntry and events with pending proposals.
- Check-In Summary section listing check-in rates per event using vertical bars and displaying entry check-ins over time.
- Email Engagement section surfacing outgoing and failed email counts.
- Follower Growth section tracking total follower counts and weekly/monthly growth rates.

PR #4384, opened 2026-07-12
feat: organizer analytics dashboard

prayag78:
Part of #4383 

## Summary

Introduces the Organizer Analytics feature, adding a unified analytics page to the organizer sidebar. It aggregates and visualizes cross-event statistics scoped strictly to the requesting user's permissions within the organizer.

### Navigation & routing

- Added an "Analytics" link (bar-chart icon) inside `get_organizer_navigation()` in `navigation.py`.
- Registered the view at `organizer/<str:organizer>/analytics` in `urls.py`.

### Scoped permissions & caching

- Each data section filters events independently using `get_events_with_permission` — orders data uses `can_view_orders`, proposals use `can_change_submissions`, check-ins use `can_checkin_orders`. Users only see data for events they have access to.
- Results cached for 10 minutes using a per-user cache key (`organizer_analytics_{organizer_id}_{user_id}`) to prevent cross-user data leakage. `?refresh=true` busts the cache.
- All queries on scoped models are wrapped with `scopes_disabled()` as required by django-scopes at the organizer context level.

### Frontend charts

- Consolidated all charts to use **ApexCharts** exclusively (loaded per-page, consistent with existing talk stats) for time-series area charts, status donut charts, and check-in bar charts.
- Configured ApexCharts area charts to preserve the classic layout styles (straight curve segments and data-point markers).
- Added library readiness polling (`initCharts`) to prevent silent load failures on slower networks.
- Deferred rendering by a 50ms layout tick so browser grid columns calculate correctly before drawing, preventing chart width overflow.

### Note:
 - Sections implemented in this PR: Ticket Sales & Revenue (orders over time, order status donut, cumulative revenue, top events table), Proposals & Sessions (proposals by state donut, proposals over time, pending proposals table), Check-In Summary (check-in rate per event bar chart, check-ins over time).
- Attendance Trends, Email Engagement, and Follower Growth are planned for a follow-up PR.
- Will make further changes based on review and discussions.

Screenshot 2026-07-16 183118
Screenshot 2026-07-15 181033
Screenshot 2026-07-13 230800


app/eventyay/static/eventyay-common/js/organizer-analytics.js
```diff
@@ -0,0 +1,264 @@
...
+        dataLabels: { enabled: false },
+        legend: { position: 'top' },
+        tooltip: {
+            shared: true,
+            x: {
+                formatter: (val, { dataPointIndex }) => {
+                    return series[dataPointIndex] ? series[dataPointIndex].event : val
+                }
+            }
+        },
```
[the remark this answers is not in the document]
prayag78: fixed

app/eventyay/eventyay_common/views/organizer_analytics.py
```diff
@@ -0,0 +1,381 @@
...
+                    'proposals_over_time_json': '[]',
+                    'pending_proposal_events': [],
+                }
+
+            content_type = ContentType.objects.get_for_model(Submission)
+            talk_ids = list(
+                Submission.objects.filter(event_id__in=event_ids)
+                .exclude(state=SubmissionStates.DELETED)
+                .values_list('id', flat=True)
+            )
```
ViRUS-0-0: In `_get_proposals_data`, `talk_ids` is extracted as a list of integers. However, `LogEntry.object_id` is a `JSONField`, and elsewhere in the codebase (such as `submission.py`), `talk_ids` are mapped to `str` before querying `LogEntry.objects.filter(..., object_id__in=talk_ids)`. If `LogEntry.object_id` stores IDs as strings, querying with integers will not match the records

Consider Mapping `talk_ids` to strings to align with the database storage type for GFK serialization in `LogEntry.object_id`

app/eventyay/eventyay_common/views/organizer_analytics.py
```diff
@@ -0,0 +1,381 @@
...
+
+            top_qs = list(
+                Order.objects.filter(event_id__in=event_ids)
+                .values('event', 'event__name', 'event__slug')
+                .annotate(
+                    total_orders=Count('pk'),
+                    paid_orders=Count('pk', filter=Q(status=Order.STATUS_PAID)),
+                )
+                .order_by('-total_orders')[:10]
+            )
```
ViRUS-0-0, line 128: The gross revenue in the "Top events by ticket sales" table is rendered as a raw float (e.g. `1050.0`) without currency symbols. Because organizers can have events with different currencies, displaying raw numbers is confusing

Consider Fetching `event__currency` in the `top_qs` values query, include it in `top_events`, load the `money` template tag library, and use the money filter

ViRUS-0-0 (changes requested):

app/eventyay/eventyay_common/views/organizer_analytics.py
prayag78: fixed

app/eventyay/eventyay_common/views/organizer_analytics.py
prayag78: `LogEntry.object_id` is stored as a numeric JSON value for Submission logs created through the generic relation. Converting these IDs to strings returns no proposal logs, so I retained the integer lookup.

app/eventyay/eventyay_common/templates/eventyay_common/organizers/analytics.html
```diff
@@ -0,0 +1,245 @@
...
+{% load static %}
+{% load compress %}
+{% load money %}
+
+{% block title %}{{ request.organizer.name }} - {% trans "Analytics" %}{% endblock %}
+
+{% block custom_header %}
+    <link rel="stylesheet" type="text/css" href="{% static 'vendored/apexcharts/apexcharts.css' %}" />
+    {% compress js %}
+        <script defer src="{% static 'vendored/apexcharts/apexcharts.min.js' %}"></script>
```
SxxAq, line 12: The PR uses Morris.js for area charts and ApexCharts for donuts/bar charts. Two full charting libraries loaded on one page increases bundle size and maintenance surface. can be considered consolidating to one library.

app/eventyay/eventyay_common/views/organizer_analytics.py
```diff
@@ -0,0 +1,396 @@
...
+    SubmissionStates,
+)
+from eventyay.control.permissions import OrganizerPermissionRequiredMixin
+from eventyay.control.views.organizer import OrganizerDetailViewMixin
+
+
+class OrganizerAnalyticsView(OrganizerDetailViewMixin, OrganizerPermissionRequiredMixin, TemplateView):
+
+    template_name = 'eventyay_common/organizers/analytics.html'
+    permission = None
```
SxxAq, line 34: `permission = None` means any organizer member can reach this page regardless
of event permissions. The per-section data scoping is correct (orders use
`can_view_orders`, proposals use `can_change_submissions`, checkins use
`can_checkin_orders`), but consider adding an early return or gate if the user
has zero permissions across all three. otherwise they get an empty dashboard
with no explanation.

app/eventyay/static/eventyay-common/js/organizer-analytics.js
```diff
@@ -0,0 +1,275 @@
...
+document.addEventListener('DOMContentLoaded', () => {
+    setTimeout(() => {
+        drawOrdersOverTime()
+        drawOrdersByStatus()
+        drawRevenueOverTime()
+        drawProposalsByState()
+        drawProposalsOverTime()
+        drawCheckinRate()
+        drawCheckinsOverTime()
+    }, 50)
```
SxxAq, line 321: If the page is slow (large JSON payloads, slow network), 50ms may not be enough and charts will silently not render. Better approach: use DOMContentLoaded + check for chart library readiness, or use MutationObserver/IntersectionObserver, or simply move the script to the bottom of the body after chart containers.

SxxAq (commented):
The PR works fine feature wise but can be improved in optimization and code quality.

app/eventyay/eventyay_common/templates/eventyay_common/organizers/analytics.html
prayag78: fixed

app/eventyay/eventyay_common/views/organizer_analytics.py
prayag78: Will handle the permissions in the follow-up PR after adding other sections, cause may need some data

app/eventyay/static/eventyay-common/js/organizer-analytics.js
prayag78: fixed

Aqil-Ahmad (changes requested):
the title says 'rate' but its not computed or shown

Image

the revenue chart seems to show lines based on the magnitude and not the real value, for example the value in PKR is less than the value in USD:

Image

can we have a single lib for consistency? for example see the date formatting for same chart type:

Image

Also i think some tests would be better here to add

prayag78:
> the revenue chart seems to show lines based on the magnitude and not the real value, for example the value in PKR is less than the value in USD:
> 

Since we don't have currency conversion fully set up yet, I will keep the current behavior and address the unified currency conversion for both the organizer and admin dashboards in a follow-up.

> Also i think some tests would be better here to add

Will add tests in the next PR along with the other sections.

Aqil-Ahmad (approved):

app/eventyay/eventyay_common/views/organizer_analytics.py
```diff
@@ -0,0 +1,396 @@
...
+
+        if 'refresh' in self.request.GET:
+            cached = None
+
+        if cached:
+            ctx.update(cached)
+            return ctx
+
+        data = {}
+        data.update(self._get_tickets_data())
```
Rachit7168, line 376: There is a significant bug here with how event names will display. In Django, the .values() method fetches raw database values and completely skips the field's from_db_value conversion. Because Event.name is an I18nCharField (which stores its data as JSON), row['event__name'] will be returned as a raw JSON string (e.g., {"en": "My Event"}). Calling str() on this will cause raw JSON to render on the analytics dashboard instead of the properly translated event name!

To fix this, you should avoid extracting event__name via .values(). Since these aggregations only return a small number of rows (like the top 10), it's much safer to fetch the actual Event objects to resolve the names.

app/eventyay/eventyay_common/views/organizer_analytics.py
```diff
@@ -0,0 +1,396 @@
...
+        return {
+            'orders_over_time_json': json.dumps(orders_over_time),
+            'orders_by_status_json': json.dumps(orders_by_status),
+            'revenue_over_time_json': json.dumps(revenue_over_time),
+            'currencies_json': json.dumps(sorted_currencies),
+            'top_events': top_events,
+            'has_orders': bool(orders_over_time or orders_by_status),
+        }
+
+    def _get_proposals_data(self):
```
Rachit7168, line 196: By casting talk_ids to a list() here, you are pulling every single submission ID into Python memory, and then passing them all into the LogEntry filter using object_id__in=talk_ids. For large organizers with thousands of submissions, passing a massive list to an __in filter can cause performance issues or exceed PostgreSQL's parameter limits.

You should remove the list() cast so Django handles it entirely in the database as an efficient SQL subquery

app/eventyay/eventyay_common/views/organizer_analytics.py
```diff
@@ -0,0 +1,396 @@
...
+                )
+                .values('order__event_id')
+                .annotate(total=Sum('amount'))
+            )
+            event_revenue = {row['order__event_id']: row['total'] for row in rev_qs}
+
+        orders_over_time = []
+        if ordered_by_day or paid_by_day:
+            all_keys = set(ordered_by_day.keys()) | set(paid_by_day.keys())
+            start = min(all_keys)
```
Rachit7168: By dynamically determining the start and end bounds of the charts based on the data (min(all_keys) / max(all_keys)), the timeline will behave unpredictably.

For example, if an organizer only sold tickets on a single day in the last 30 days, start will equal end and the graph will shrink down to a single point, rather than showing a full 30-day timeline with zeroes everywhere else. This breaks the promise of the panel title ('Orders over time (last 30 days)').

You should anchor the loop to the requested timeframe so the chart always spans the correct window

Rachit7168 (changes requested):
Thank you for working on this PR! Please address the following:

app/eventyay/eventyay_common/views/organizer_analytics.py
prayag78: fixed

app/eventyay/eventyay_common/views/organizer_analytics.py
prayag78: fixed

Rachit7168 (approved):
LGTM , TESTED , WORKS : 

https://github.com/user-attachments/assets/86b1153d-b9ce-47ad-bd4f-fa54c612ba9f

sarafarajnasardi (approved):

Saksham-Sirohi (approved):

MukundC25 (approved):

ArnavBallinCode (approved):
LGTM

PR #4389, opened 2026-07-13
fix(schedule): correct speaker track filter matching

Rachit7168:
## Summary

The Speakers page track filter returned **"No speakers found"** whenever a track was selected. The cause was a case-sensitive lookup: `resolvedSpeakers` mapped sessions with `sessionsBySpeaker[speaker.code]`, but `speaker.code` is uppercase while `sessionsBySpeaker` is keyed with lowercased codes, so speakers ended up with no sessions and were filtered out.

- Use the existing case-insensitive `sessionsForSpeaker(sessionsBySpeaker, speaker.code)` helper so speakers are correctly associated with their sessions regardless of code casing.
- Minor mobile refinements to the speakers toolbar/dropdown sizing (auto widths, sensible min-width, ellipsis for long items) and session card spacing.
- Show the full favourite count instead of capping the label at `99+`.

## Test plan
- [ ] Open the public Speakers page, select one or more tracks, and confirm matching speakers are listed (no longer "No speakers found").
- [ ] Verify language/other filters still work together with track filtering.
- [ ] Check the speakers toolbar/dropdown and session cards on a mobile-width viewport.

https://github.com/user-attachments/assets/1b15ae6c-6031-493d-b08f-a2f1bb5e35de

### DIFF
--- a/app/eventyay/webapp/schedule/src/components/Session.vue
+++ b/app/eventyay/webapp/schedule/src/components/Session.vue
@@ -222,7 +222,7 @@ export default {
 		},
 		favCountLabel () {
 			const count = normalizePopularityCount(this.session)
-			return count > 99 ? '99+' : String(count)
+			return String(count)
 		},
 		doNotRecordTooltip () {
 			const m = this.translationMessages || {}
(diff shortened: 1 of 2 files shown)


Rachit7168:
https://github.com/user-attachments/assets/07b4b71c-6cc7-4071-987b-a4e5a67f5230

Saksham-Sirohi (approved):

PR #4396, opened 2026-07-13
fix: use ajax for voucher tab switching

### LINKED ISSUE #4395: Use AJAX for tab switching on Vouchers page to prevent full page reload Description
The Vouchers page currently performs a full page refresh when switching between tabs such as All vouchers and Voucher tags. Other tabbed pages in the application (for example, Cancellation Settings) use AJAX-based tab switching, providing a smoother user experience without reloading the entire page.

The Vouchers page should follow the same behavior by loading tab content asynchronously.

## Steps to Reproduce

1)Start Eventyay locally.
2)Open an event dashboard.
3)Navigate to Vouchers.
4)Click between the All vouchers and Voucher tags tabs.
5)Observe that the entire page reloads during each tab switch.

## Actual Behavior

Switching between voucher tabs triggers a full page reload, causing unnecessary page refreshes and interrupting the user experience.

## Expected Behavior

Switching between voucher tabs should use AJAX to load only the tab content, similar to other tabbed pages in the application, without reloading the entire page.

## Related video

https://github.com/user-attachments/assets/f988025b-f893-4ad7-a3ef-03e29dc9c88c

arc12at:
Closes #4395 

Currently, switching between the tabs on the Vouchers page (All vouchers and Voucher tags) triggers a full page refresh, causing a visible white flash and reloading the entire layout (sidebar, headers, etc.).

This PR resolves the issue by extending the existing general `ajax-filter.js` framework to intercept link-based tab transitions and process them dynamically via AJAX, matching the smooth tab transitions seen in other parts of the application.

## Video

https://github.com/user-attachments/assets/60a82bf2-6507-424a-a300-469871d02ef9

> [!NOTE]
> **Note on Implementation Approach:**
>
>While `ajax-filter.js` is typically used for searching and pagination, we extended it for tab switching on the Vouchers page because the tabs ("All vouchers" and "Voucher Tags") require completely different data from the backend. 
>
>Fetching both datasets on initial page load would be inefficient and slow. By leveraging `ajax-filter.js`, a tab click acts as a URL filter (e.g., `?tab=tags`), allowing us to dynamically fetch and inject only the necessary data from the server without refreshing the entire page or duplicating logic for browser history and network requests.


app/eventyay/static/eventyay-common/js/ajax-filter.js
```diff
@@ -261,17 +268,75 @@ const handleSortClick = function(event) {
...
+    fetchAndReplace(url, true, formInTab || tabPane || defaultForm);
+};
+
+const setActiveTabUI = function(tabId) {
+    if (!tabId) return;
+
+    // Update the nav-tabs active state
+    document.querySelectorAll('.nav-tabs a').forEach(function(link) {
+        const linkUrl = new URL(link.getAttribute('href'), window.location.href);
+        const linkTab = linkUrl.searchParams.get('tab');
```
ViRUS-0-0, line 283: If a `.nav-tabs` a element is missing an `href` attribute, `link.getAttribute('href')` will return null. The `new URL()` constructor will then stringify it and parse the relative URL `"null"`, which could theoretically cause an unexpected mismatch. Adding an early return check prevents this

consider adding a null check for href before parsing it as a URL

ViRUS-0-0 (changes requested):
The changes done work as expected but left a comment please resolve.

https://github.com/user-attachments/assets/42c87fb3-f33e-4bb3-b0d8-3e7818a38255

app/eventyay/static/eventyay-common/js/ajax-filter.js
arc12at: Fixed

Rachit7168 (approved):
LGTM , TESTED :  

https://github.com/user-attachments/assets/f65f5fad-793a-448b-893a-956173f508af

ViRUS-0-0 (approved):
LGTM

prayag78 (approved):

ArnavBallinCode (approved):

Saksham-Sirohi (approved):

PR #2976, opened 2026-03-24
fix(settings): add missing ADMINS, VIDEO_SERVER_HOSTNAME, CACHE_TICKE…

Rachit7168:
Add ADMINS, VIDEO_SERVER_HOSTNAME, CACHE_TICKETS_HOURS, FETCH_ECB_RATES to settings. Code was trying to access these but they weren't defined.

Partial part of Issue : https://github.com/fossasia/eventyay/issues/1546
and inactive PR : https://github.com/fossasia/eventyay/pull/2154

## Summary by Sourcery

Define missing configuration settings and wire them into Django settings, including safer handling when the video server is not configured.

Bug Fixes:
- Expose ADMINS, VIDEO_SERVER_HOSTNAME, CACHE_TICKETS_HOURS, and FETCH_ECB_RATES via the settings config to prevent runtime access to undefined settings.
- Skip video world creation when VIDEO_SERVER_HOSTNAME is not configured, avoiding errors in background tasks.

PR #4422, opened 2026-07-16
fix: add missing settings keys to prevent AttributeError at runtime

### LINKED ISSUE #1546: Code refers to some undefined settings key
**Describe the bug**

Sometimes we see logs like this:

```
eventyay.eventyay_common.tasks.create_world[368ad1b4-e7c5-42a5-8b69-843fc1b91214] raised unexpected: AttributeError("'Settings' object has no attribute 'VIDEO_SERVER_HOSTNAME'")
```

It is because some parts of our code are trying to use some settings which don't exist. We need to review case by case and do either:
- Copy missing settings from old eventyay-xxx code, or pretix, pretalx.
- Delete the code which uses that settings, if that code is not relevant.

prajakta128:
## Problem

Closes #1546

Several settings were referenced across the codebase but were not defined in `eventyay/config/settings.py`. This could result in `AttributeError` exceptions when those settings are accessed.

## Changes

Added the following missing settings with appropriate default values:

| Setting | Default Value |
|---------|---------------|
| `FILE_UPLOAD_MAX_SIZE_OTHER` | `10 * 1024 * 1024` (10 MB) |
| `SENTRY_ENABLED` | `bool(SENTRY_DSN)` |
| `CACHE_TICKETS_HOURS` | `24 * 7` (1 week) |
| `FETCH_ECB_RATES` | `True` |
| `VIDEO_SERVER_HOSTNAME` | `os.getenv("VIDEO_SERVER_HOSTNAME", "")` |
| `HAS_GEOIP` | `False` |

## How This Was Verified

- Listed all defined Django settings.
- Scanned the codebase for `settings.<SETTING_NAME>` references.
- Compared both lists to identify missing settings.
- Added sensible default values for all missing settings.

## Result

This prevents `AttributeError` exceptions caused by undefined settings while maintaining backward compatibility.

## Summary by Sourcery

Define missing Django settings with sensible defaults to prevent runtime AttributeError when accessing undefined configuration options.

New Features:
- Add FILE_UPLOAD_MAX_SIZE_OTHER setting to control default limit for non-primary file uploads.
- Introduce SENTRY_ENABLED setting derived from SENTRY_DSN to toggle Sentry integration.
- Add CACHE_TICKETS_HOURS setting to control ticket caching duration in hours.
- Add FETCH_ECB_RATES setting to enable or disable fetching ECB exchange rates.
- Add VIDEO_SERVER_HOSTNAME setting sourced from environment for video server configuration.
- Add HAS_GEOIP setting indicating availability of GeoIP2 database.


prajakta128:
Hi @Rachit7168 @AmavBallinCode @Saksham-Sirohi @hongquan — just checking in on this PR, it's been open for about a week and all checks are passing. I've updated the branch against the latest dev.

Happy to make any changes needed — let me know if there's anything you'd like adjusted, or if this needs a different reviewer's attention.

app/eventyay/config/settings.py
```diff
@@ -1529,6 +1530,12 @@ def increase_redis_db(url: str, increment: int) -> str:
 
 CALL_FOR_SPEAKER_LOGIN_BUTTON_LABEL = conf.call_for_speaker_login_button_label
 
+# Default values for settings used across the codebase
+CACHE_TICKETS_HOURS = 24 * 7  # Cache tickets for 1 week by default
+FETCH_ECB_RATES = True  # Fetch ECB exchange rates for foreign currency invoices
+VIDEO_SERVER_HOSTNAME = os.getenv('VIDEO_SERVER_HOSTNAME', '')  # Hostname of the video server
```
hongquan: Why needing `VIDEO_SERVER_HOSTNAME`?

app/eventyay/config/settings.py
```diff
@@ -1529,6 +1530,12 @@ def increase_redis_db(url: str, increment: int) -> str:
 
 CALL_FOR_SPEAKER_LOGIN_BUTTON_LABEL = conf.call_for_speaker_login_button_label
 
+# Default values for settings used across the codebase
+CACHE_TICKETS_HOURS = 24 * 7  # Cache tickets for 1 week by default
```
hongquan: What is nice about Python is that it has `timedelta` type, then why do the dumb way with `_HOURS` suffix, integer value and cost a comment to explain "1 week".

app/eventyay/config/settings.py
```diff
@@ -1529,6 +1530,12 @@ def increase_redis_db(url: str, increment: int) -> str:
...
+FETCH_ECB_RATES = True  # Fetch ECB exchange rates for foreign currency invoices
+VIDEO_SERVER_HOSTNAME = os.getenv('VIDEO_SERVER_HOSTNAME', '')  # Hostname of the video server
+HAS_GEOIP = False  # Set to True if GeoIP2 database is available
```
hongquan: Does this need to be in setting?
Can the system detect if GoeIP2 database is available and act accordingly?

hongquan (changes requested):
Almost low-quality AI-generated.

Saksham-Sirohi (changes requested):
Agree with @hongquan — hardcoding a pile of defaults in `settings.py` isn't the right fix.

Pull these through `conf` like everything else (`VIDEO_SERVER_HOSTNAME = os.getenv(...)` shouldn't bypass that). Don't hardcode `HAS_GEOIP = False` either, that just silently turns GeoIP off.

Only add keys that are actually blowing up, and note the traceback for each. `SENTRY_ENABLED = bool(SENTRY_DSN)` is fine if it lives next to the existing Sentry setup.

prajakta128:
Hi @hongquan @Saksham-Sirohi — thanks for the detailed feedback, you were right that the original version just patched over the AttributeErrors without addressing the underlying issue. Reworked it:

- VIDEO_SERVER_HOSTNAME, CACHE_TICKETS_MAX_AGE, FETCH_ECB_RATES, and a new geoip_path are now proper typed fields on BaseSettings, pulled through `conf` like everything else in this file — not os.getenv() or hardcoded constants.
- CACHE_TICKETS_HOURS is now CACHE_TICKETS_MAX_AGE, stored as a real timedelta (matches how it's actually used in cleanup.py — no more wrapping it in timedelta(hours=...) at the call site).
- HAS_GEOIP is no longer hardcoded False. It's derived from whether GEOIP_PATH is actually configured and the database file exists, so it reflects real availability instead of silently disabling the feature for everyone.
- Left SENTRY_ENABLED and FILE_UPLOAD_MAX_SIZE_OTHER as-is per your note that those are fine.

Verified locally: app starts cleanly with no errors, and all four settings resolve to correct values (confirmed via manage.py shell).

Let me know if you'd like anything adjusted further.

app/eventyay/config/settings.py
```diff
@@ -1529,6 +1535,14 @@ def increase_redis_db(url: str, increment: int) -> str:
 
 CALL_FOR_SPEAKER_LOGIN_BUTTON_LABEL = conf.call_for_speaker_login_button_label
 
+VIDEO_SERVER_HOSTNAME = conf.video_server_hostname
+CACHE_TICKETS_MAX_AGE = conf.cache_tickets_max_age
+FETCH_ECB_RATES = conf.fetch_ecb_rates
+# GeoIP is only enabled if a database path was actually configured and exists,
```
Rachit7168: Remove unnessary comments no need

Saksham-Sirohi (commented):

No additional findings beyond the existing review thread.

prajakta128:
Hi all — just following up, this has been passing checks for a couple of days now with no further concerns raised. Would appreciate a review/approval when you have a moment. Thanks for the feedback so far, it's genuinely made the fix better.

Rachit7168:
Thank you for contributing to the FOSSASIA Eventyay project! We sincerely apologize for the delay in reviewing your contribution. We truly appreciate your patience and don't want this to discourage you from contributing to Eventyay. Your contribution is valuable to the project, and we will review it as soon as possible. Thank you for your understanding and continued support!

hongquan:
You should look into the previous PRs that try to work on the same issue.
They all try to add `VIDEO_SERVER_HOSTNAME` settings when the video server is actually this Eventyay instance.
Please check this [comment](https://github.com/fossasia/eventyay/issues/1546#issuecomment-5091629073).

prajakta128:
Hi @hongquan — I looked into this and found your earlier comment on #2154 asking the same question. I also see #2154 and #2976 both stalled on exactly this point without a final answer, and the code + app/eventyay.cfg still reference VIDEO_SERVER_HOSTNAME as of current dev.

Since this seems like an architecture decision rather than something I can determine from outside, could you confirm: is VIDEO_SERVER_HOSTNAME fully obsolete now that video lives in the same instance, or is it still valid for certain self-hosted/external video deployments?

In the meantime, to avoid this blocking the rest of the fix, I can apply the same safe-guard #2976 proposed: skip video world creation with a clear warning log if VIDEO_SERVER_HOSTNAME isn't configured, instead of crashing. That way the setting stays harmless either way until the architecture question is settled. Let me know if that approach works, or if you'd rather I remove it outright.

hongquan:
It has been confirmed since February:

image

> I also see https://github.com/fossasia/eventyay/pull/2154 and https://github.com/fossasia/eventyay/pull/2976 both stalled on exactly this point without a final answer

The missing answer is the answer from the PR creator. I keep reminding them that the "video server" is actually the same Eventyay system.

> the code + app/eventyay.cfg still reference VIDEO_SERVER_HOSTNAME as of current dev.

That's why I pointed you to this [comment](https://github.com/fossasia/eventyay/issues/1546#issuecomment-5091629073).

app/eventyay/eventyay_common/tasks.py
```diff
@@ -632,4 +551,4 @@ def billing_invoice_send_email(subject, content, invoice, organizer_billing):
             'attach_file_base64': pdf_base64,
             'attach_file_name': pdf_buffer.filename,
         }
-    )
+    )
```
Rachit7168, line 554: Unnecessary change

prajakta128: Looks like this line was already fixed in my commit — checked HEAD and the file ends cleanly with no trailing newline issue. Might have been comparing against an older revision. Let me know if you're still seeing it on your end.

hongquan (changes requested):
I'm not patient with the PR that is AI-generated.
This is the last chance for you to check all the previous reviews. If you seem to still depend on AI, I will close the PR.

prajakta128:
Thanks for the feedback. I understand your concern.
I went back through each review comment and made sure I understood the reasoning behind the changes before updating the PR. I verified the relevant code paths and checked that the changes were correct rather than just applying the suggested fixes. 
If you think there's any part of the PR that needs more explanation or another review, I'm happy to discuss it.

Saksham-Sirohi:
Closing in favor of #4645, which better matches the intended direction for #1546 (remove obsolete `VIDEO_SERVER_HOSTNAME` / `create_world` plumbing rather than adding the setting, and wire the remaining keys through existing config patterns).

Thank you for your contribution and for iterating on the review feedback — the work helped push this issue forward.

hongquan:
> If you think there's any part of the PR that needs more explanation or another review, I'm happy to discuss it.

Line 148:

```py
video_server_hostname: str = ''
```

prajakta128:
> > If you think there's any part of the PR that needs more explanation or another review, I'm happy to discuss it.
> 
> Line 148:
> 
> ```python
> video_server_hostname: str = ''
> ```

Checked — video_server_hostname is still present because neither this PR nor #4645 was merged; both were closed unmerged, so #1546 is still open in dev. My branch (before it was closed) did remove it, along with create_world/_create_world, verified via repo-wide search that neither was called anywhere outside that one file.

I understand the concern about AI dependency, and I want to be upfront: I used AI assistance heavily while working through this, especially to verify claims (like checking whether create_world was actually dead code) rather than just guessing. I recognize that's not the same as being able to independently defend every line without support. If it would help, I'm glad to answer specific questions about the reasoning directly, without pasting anything in first.

PR #4434, opened 2026-07-16
fix: handle AnonymousUser in talk rules predicates

Rachit7168:
This PR fixes a TypeError that occurs when an unauthenticated user accesses the submissions API. The AnonymousUser object does not have an ID, which causes an error when passed into the `teams.filter(members__in=[user])` query inside the talk_rules predicates. Fixed by immediately returning False for unauthenticated users.

## Summary by Sourcery

Bug Fixes:
- Prevent TypeError in talk rules predicates when an AnonymousUser or unauthenticated user is evaluated by team membership queries.


Saksham-Sirohi (approved):
enforcement fix

PR #4454, opened 2026-07-18
fix(badges): keep autofit text inside container width

Saksham-Sirohi:
## Summary
- Fix badge/PDF editor "Adapt content to container width" so text stays within the defined box width instead of expanding the container or wrapping outside it
- Measure full line width when computing autofit font size in both the editor preview and PDF renderer
- Prevent Fabric.js text areas from auto-expanding width beyond the user-set container

## Test plan
- [x] `pytest tests/tickets/plugins/badges/test_utils.py -q -k fit_fontsize`
- [ ] Open badge layout editor, add a text field with a narrow width, enable **Adapt content to container width**, and verify long sample text shrinks inside the box
- [ ] Use **Preview** with long attendee names and confirm PDF output matches the editor bounds

### DIFF
--- a/app/eventyay/base/pdf.py
+++ b/app/eventyay/base/pdf.py
@@ -1017,10 +1017,7 @@ class Renderer:
         while size > min_size:
             fits = True
             for line in lines:
-                parts = line.split()
-                if not parts:
-                    continue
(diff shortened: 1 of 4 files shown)


ArnavBallinCode (approved):
LGTM

PR #4501, opened 2026-07-19
Fix room scoped audio translation

### LINKED ISSUE #4470: [Video] Selected Audio Translation Language / Video Stream Overwrites Other Unconfigured Rooms Globally
### Description

When multiple rooms are configured with completely different stream sources, switching the **Audio Translation** language in one room mistakenly synchronizes and overrides the video source across all other rooms, even if those rooms do not have that language interpretation track configured.

---

### Steps to Reproduce

1. Navigate to the event home page containing multiple stages/rooms (e.g., *Minimal Studio*, *Sabi, MIA BOYKA...*, *love nwantinti...*).
2. Enter **Room 1 (Minimal Studio)** and verify it plays its unique video stream (e.g., Jonathan Young's English cover of "Unravel").
3. Go back and check **Room 2** and **Room 3** to verify they are playing distinct, separate video streams.
4. Return to **Room 1** and open the **Audio Translation** settings menu at the bottom right.
5. Select a different language track (e.g., **Japanese**). The video stream updates correctly in Room 1 to the Japanese source ("Tokyo Ghoul Opening - Unravel by TK").
6. Navigate back to the main stage list and open **Room 2** or **Room 3**.

### Expected Behavior

* **Room 2** and **Room 3** should continue to play their originally assigned video streams.
* Audio translation track selections made in Room 1 should remain scoped exclusively to Room 1.

### Actual Behavior

* Changing the language track in Room 1 forces the translated video stream ("Tokyo Ghoul Opening - Unravel by TK") globally onto **Room 2** and **Room 3**, completely overriding their distinct native streams.

---

https://github.com/user-attachments/assets/458f4cf7-bc83-4e4e-8087-72342dd100a9

sarafarajnasardi:
Fixes #4470 

## Summary

  Fixes audio interpretation behavior for YouTube rooms by making translation selection room-scoped and more robust.

  ## Changes

  - Store selected YouTube/audio translation per room instead of globally.
  - Reset translation state only for the room being entered, avoiding leakage across rooms.
  - Filter invalid interpretation language entries before showing them in the dropdown.
  - Normalize YouTube IDs and URL-based audio sources consistently.
  - Reapply selected translation after media iframe rebuilds, including stream/module changes.
  - Fix switching between two video-based interpretation channels so the player updates to the newly selected interpretation video.
  - Clean up stale translation iframe/WHEP clients when the main iframe is destroyed.
  
  

https://github.com/user-attachments/assets/022e5a32-396b-408c-bf36-a31624b6be21


app/eventyay/webapp/video/src/components/AudioTranslationDropdown.vue
```diff
@@ -39,25 +44,28 @@ export default {
 		}
 	},
 	methods: {
+		isUsableLanguageEntry(entry) {
```
Rachit7168: THis method duplicates isUsableLanguageEntry/normalizeAudioSource in AudioTranslationDropdown.vue (same "valid YouTube ID or parseable URL" logic, ~15 lines each). Please extract a single shared helper into lib/validators.js next to normalizeYoutubeVideoId and import it in both places, so the two copies can't drift apart.

app/eventyay/webapp/video/src/components/AudioTranslationDropdown.vue
```diff
@@ -29,7 +29,12 @@ export default {
 		languages: {
 			immediate: true,
 			handler(newLanguages) {
-				this.languageOptions = newLanguages.map(entry => entry.language) // Directly assigning the list of languages
+				this.languageOptions = newLanguages
+					.filter(entry => this.isUsableLanguageEntry(entry))
```
Rachit7168: The dropdown already filters unusable entries in its languages watcher, so this list is filtered twice. Filtering here is still needed for the v-if="languages.length > 1" visibility check — but once the helper is shared (see comment on line 163), consider filtering only here and dropping the dropdown-side filter, since the dropdown always receives this pre-filtered array

app/eventyay/webapp/video/src/views/rooms/item.vue
```diff
@@ -136,22 +137,40 @@ export default {
...
 			}
 			if (!this.languages.find(lang => lang.language === 'Original')) {
 				this.languages.unshift({language: 'Original', youtube_id: null, use_video: false})
 			}
 			// Reset translation only when actually changing rooms, not on component remount
 			const currentRoomId = this.room?.id
 			if (this.previousRoomId !== null && this.previousRoomId !== currentRoomId) {
-				this.$store.commit('updateYoutubeTransAudio', null)
+				this.$store.commit('updateYoutubeTransAudio', {
+					roomId: currentRoomId,
```
Rachit7168: This branch is effectively dead code: the Room view remounts on every navigation (the router-view in App.vue is keyed by $route.fullPath), so previousRoomId is always null when initializeLanguages runs and this reset never fires. The state still gets reset in practice, but only as a side effect of the remounted dropdown auto-selecting "Original" and emitting languageChanged.

Two consequences worth addressing:

Re-entering a room with a saved translation can briefly play the translated video (the shared MediaSource applies the stale store entry on mount) before the dropdown's "Original" commit rebuilds the iframe — two iframe loads and a visible flash.
If a room's dropdown stops rendering (languages.length <= 1 after a config change), a stale entry for that room is applied with no UI to undo it.
Suggest either clearing the room's entry in beforeUnmount (deliberate reset), or initializing the dropdown's selection from the store (real per-room persistence). Either is cleaner than relying on the remount side effect.

Rachit7168 (changes requested):
Please address

Rachit7168 (changes requested):
Image

Why is the the selection is changed to dropdown, In recent it is changed to radio button and the dropdown is not working

sarafarajnasardi:
> Image
> Why is the the selection is changed to dropdown, In recent it is changed to radio button and the dropdown is not working

the branch was not updated...that's why

Rachit7168 (approved):
LGTM , WORKING TESTED 

https://github.com/user-attachments/assets/57bdb570-9ab0-4206-aa56-73e63365401f

ArnavBallinCode (approved):
LGTM

PR #4503, opened 2026-07-19
Make exhibition requests discoverable from the account dropdown

### LINKED ISSUE #76: Allow submitters to view and track their exhibition proposals after submission
### Description
When an attendee submits a talk proposal or places a ticket order in eventyay, they can view and track it from their account dashboard. Exhibition proposals submitted through the Call for Exhibitors or Call for Sponsors have no equivalent. A submitter has no clear way to find their submission, check its status, or edit it after submitting, even though they are logged in with an eventyay account.

**To verify** if the `UserProposalListView` is currently linked from the submitter's account dashboard or any post-submission confirmation page?

### Requirements
- After submitting a proposal, the submitter can view their submission and its current state (submitted, accepted, rejected) from their eventyay account
- The experience should be consistent with how eventyay handles order placement or talk submissions, submitters find their exhibition proposals in the same place they find their other eventyay activity
- A submitter can edit their proposal (even after it is accepted)
- Submitting without an eventyay account is not supported, an account is required to apply

Aqil-Ahmad:
fixes https://github.com/fossasia/eventyay-exhibition/issues/76 and related to https://github.com/fossasia/eventyay-exhibition/pull/97

image

## Summary by Sourcery

New Features:
- Add a user_menu_items event plugin signal to allow extensions to the user account dropdown menu.


app/eventyay/common/signals.py
```diff
@@ -175,6 +175,9 @@ def wrapper(*args, **kwargs):
 than expected.
 """
 
+user_menu_items = EventPluginSignal()
```
Rachit7168, line 178: Because the account dropdown is a global user menu, this needs to be a standard Django signal, not an `EventPluginSignal`. When a user navigates to their global profile page (`/user/`), `request.event` is `None`. For an `EventPluginSignal`, a `None` sender silently disables all plugin receivers, meaning the menu item will vanish. Please change this to: `user_menu_items = django.dispatch.Signal()` (and ensure `django.dispatch` is imported if not already).

app/eventyay/common/templates/common/base.html
```diff
@@ -173,6 +174,7 @@ <h1 class="public-event-header">
                                             </a>
                                         {% endif %}
                                     {% endif %}
+                                    {% eventsignal request.event "eventyay.common.signals.user_menu_items" request=request icon_class="mr-2 mb-1" %}
```
Rachit7168, line 177: Since the signal must be a global Django signal, replace `{% eventsignal %}` with the global HTML signal tag: 
`{% html_signal "eventyay.common.signals.user_menu_items" sender=request.event request=request icon_class="mr-2 mb-1" %}`

Also make sure to replace `{% load eventsignal %}` with `{% load html_signal %}` at the top of the file.

app/eventyay/common/templates/common/fragment_header_user.html
```diff
@@ -121,6 +122,7 @@
                             </a>
                         {% endif %}
                     {% endif %}
+                    {% eventsignal current_event "eventyay.common.signals.user_menu_items" request=request %}
```
Rachit7168, line 125: Same as in `base.html`, please replace this with:
`{% html_signal "eventyay.common.signals.user_menu_items" sender=current_event request=request %}`

And replace `{% load eventsignal %}` with `{% load html_signal %}` at the top of the file.

Rachit7168 (changes requested):
This PR's goal is straightforward, but the implementation uses an event-bound signal instead of a global signal. This completely breaks the feature when the user is navigating global account pages because `request.event` will be `None`, causing `EventPluginSignal` to disable the receivers. As a result, the exhibition menu item will vanish from the user dropdown randomly depending on what page the user is currently viewing. 

Please address this blocker to ensure a stable user experience.

Aqil-Ahmad:
> This PR's goal is straightforward, but the implementation uses an event-bound signal instead of a global signal. This completely breaks the feature when the user is navigating global account pages because `request.event` will be `None`, causing `EventPluginSignal` to disable the receivers. As a result, the exhibition menu item will vanish from the user dropdown randomly depending on what page the user is currently viewing.

your comments are valid and they mention a single thing, but if the actual goal is to surface exhibition proposals on the global account page (not the event dropdown), that's a worthwhile but separate design change. CfP doesn't do that today either, so I kept this consistent with CfP.

Rachit7168 (approved):
That makes sense. If the intent is for the dropdown to be contextual to the current event (consistent with how CfP and other event-bound items work), then using `EventPluginSignal` is indeed the correct approach. Thanks for clarifying! Approving.

Saksham-Sirohi (dismissed):
Verified. Adding `user_menu_items` as an `EventPluginSignal` and wiring it into both account dropdowns is the right hook for event-contextual exhibition entries. Aligns with how other event-bound menu items work.

Saksham-Sirohi (changes requested):

1. Templates gain a `user_menu_items` hook, but this PR never registers a receiver that emits an exhibition/proposal menu entry.
2. As shipped, the account dropdown will not show the discoverability link the PR describes.

MukundC25 (approved):
Tested, flow works! 

https://github.com/user-attachments/assets/52fd4569-3ee7-405e-962e-671bfa6526e3

Saksham-Sirohi (approved):

ArnavBallinCode (approved):

PR #4526, opened 2026-07-21
Add alt text to image preview widgets across the control panel

### LINKED ISSUE #4525: Accessibility: ImageInput widget renders preview images without alt text
While going through the codebase I noticed `common/templates/common/widgets/image_input.html` renders the uploaded-image preview with a bare `` tag and no `alt` attribute:
 
Image

This is the `ImageInput` widget defined in `common/forms/widgets.py`. It's used as the preview for pretty much every image upload field in the organizer/control panel — organizer header image, event logo, ticket layout background, badge design, presale banner, etc. I counted around 20 `ImageField` usages that go through this same widget. So one missing attribute means a screen reader user gets zero context for the image preview across all of those forms.

Reproduced this locally on a fresh dev instance:
1. Organizer settings → Organizer page tab → upload a Header image → Save
2. Right-click the preview thumbnail that appears → Inspect
3. DevTools shows the `` with a `src` but no `alt`

Screenshot attached, taken from that exact flow.

The same gap also exists separately in `control/templates/pretixcontrol/order/index.html` (line ~490), for the uploaded-file-answer image preview on the order detail page — different template, same missing-alt issue.

For comparison, `common/templates/common/avatar.html` handles this correctly for the user avatar preview — it passes a translated `alt="Your avatar"`. That's the pattern I'd follow for a fix.

**Proposed fix**
- Give `ImageInput.get_context()` a way to pass a meaningful alt text (e.g. derived from the field label, or an explicit `alt_text` kwarg on the widget) and use it in `image_input.html`.
- Fall back to something generic like "Image preview" if no better label is available — this is user-uploaded content, not decoration, so an empty alt isn't appropriate here.
- Apply the same fix to the order detail template (can reuse `q.answer.file_name` for the alt there).

Happy to work on this if it's not already assigned — seems like a good scoped first PR: one widget template, one Python file, one order template, no migrations, no behavior change beyond markup.

prajakta128:
Fixes #4525

## What changed

Found and fixed three places where preview `` elements had no `alt` attribute:
Screenshot 2026-07-21 101540

1. **`ImageInput`** (`common/forms/widgets.py` / `image_input.html`) — the widget backing ~20 `ImageField` usages across the control panel (event/organizer logos, badge layouts, banners, etc). Added a `get_context` override providing an explicit alt when set via `widget.attrs`, falling back to a generic translated string ("Image preview") otherwise.

2. **`thumbnailed_file_input.html`** — while testing the fix above on the organizer header image field, I found it doesn't actually use `ImageInput`. It uses a *separate, identically-named* `ClearableBasenameFileInput` class defined in `control/forms/__init__.py`, which renders through this different template. One of its two preview branches already had `alt="{% trans 'Preview' %}"`; the other (used when `event_settings_image_tools` isn't set) didn't. Fixed to match.

3. **Order detail file-answer preview** (`control/templates/pretixcontrol/order/index.html`) — same gap, unrelated widget. Fixed using the uploaded file's name as alt text.

## Note for reviewers

There are two separate classes both named `ClearableBasenameFileInput` (one in `common/forms/widgets.py`, one in `control/forms/__init__.py`) with different templates. That's how this bug ended up in two different templates despite looking like "one widget." Worth being aware of for anyone touching either of these in the future — happy to open a follow-up naming-cleanup issue if that seems useful.

## Testing

Reproduced and verified locally: uploaded an organizer header image, confirmed the preview `` now renders with `alt="Preview"` (previously empty — see #4525).

No migrations, no behavior change beyond markup.

## Summary by Sourcery

Ensure all image preview widgets in the control panel render meaningful alt text for improved accessibility.

New Features:
- Provide configurable alt text for the shared ImageInput widget, defaulting to a generic translated image preview label.

Bug Fixes:
- Add missing alt attributes to image previews in thumbnailed file inputs and order detail file-answer previews to avoid empty alt text in rendered markup.

Enhancements:
- Update image preview templates to pass through and display alt text consistently across control panel views.


ViRUS-0-0 (approved):
Please Resolve Conflicts

ViRUS-0-0 (dismissed):
Please Resolve Conflicts

prajakta128:
Hi @ViRUS-0-0 — resolved the conflicts and rebased onto the latest dev. Should be clean and ready for review now. Let me know if anything else needs changing!

Saksham-Sirohi (dismissed):
Small accessibility win — alt text on image previews is welcome. Please restore the trailing newline in `thumbnailed_file_input.html` before merge (file currently ends without one). Not otherwise blocking.

prajakta128:
Hi @Saksham-Sirohi @ViRUS-0-0 — restored the trailing newline in thumbnailed_file_input.html and rebased to resolve the conflicts (dev had also added SVG handling to this template in the meantime, merged that in too). Should be clean and ready for review now.

Saksham-Sirohi (commented):

Alt text wiring on the thumbnail widgets looks correct; nothing else stood out in the diff.

prajakta128:
Hi all — just following up, this has been passing checks for a couple of days now with no further concerns raised. Would appreciate a review/approval when you have a moment. Thanks for the feedback so far, it's genuinely made the fix better.

Saksham-Sirohi:
Pushed a small follow-up on top:

1. Restored the missing trailing newline in `thumbnailed_file_input.html` (had regressed after the latest merge).
2. Hardened `ImageInput.get_context` so an explicit `alt` is used for the preview image but popped from widget attrs, avoiding `alt` leaking onto the `<input type="file">`.

Smoke-checked `ImageInput.get_context` locally (default / widget attrs / per-render attrs) and verified the three templates still render the expected alt attributes.

ArnavBallinCode (approved):

Saksham-Sirohi (approved):

PR #4544, opened 2026-07-22
fix(video): require JWT for video access unless public link is enabled

Saksham-Sirohi:
## Summary
- Require a ticket/organizer JWT for video access by default; anonymous `client_id` login is denied unless the public video link is enabled.
- Include the event attendee trait in organizer JWTs so signed-in organizers keep access.
- Harden token decoding for events with missing JWT config so clients get `auth.missing_token` / `auth.invalid_token` instead of `server.fatal`.

## Test plan
- [ ] Open `/{org}/{event}/video/` without a token (or with only a random `client_id`) and confirm access is denied with the existing token error UI
- [ ] Join via attendee ticket JWT and confirm attendee-level access works
- [ ] Join via organizer video access flow and confirm elevated organizer permissions work
- [ ] Enable “show public video link” and confirm anonymous access works again
- [ ] On an event with `config=None`, send a garbage token and confirm `auth.invalid_token` (not fatal server error)
- [ ] Run `pytest tests/stable/test_video_jwt_required.py tests/stable/test_attendee_direct_message_permission.py`

### DIFF
--- a/app/eventyay/base/models/event.py
+++ b/app/eventyay/base/models/event.py
@@ -62,7 +62,10 @@ from eventyay.core.permissions import (
     traits_match_required,
 )
 from eventyay.core.utils.json import CustomJSONEncoder
-from eventyay.eventyay_common.video.permissions import VIDEO_TRAIT_ROLE_MAP
+from eventyay.eventyay_common.video.permissions import (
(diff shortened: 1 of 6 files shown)


ArnavBallinCode (approved):

Rachit7168:
@codex review

PR #4393, opened 2026-07-13
feat: add teamshifts permission fields and roles management to core team model

MukundC25:
Closes #4304 | Paired with fossasia/eventyay-teamshifts#57

#### Summary

Adds a simplified 2-role TeamShifts permission model to the core Team model, replacing the previous 5 granular boolean fields per reviewer feedback. Organizers assign either **Event Coordinator** (full TeamShifts access) or **Team Lead** (shifts + emails only) with optional role scoping and email visibility control.

#### What Changed

**Model & Migration (0054)**
- `teamshifts_role` CharField: `''` (no access), `'coordinator'`, or `'lead'`
- `all_teamshifts_roles` / `limit_teamshifts_roles`: role scoping (Team Lead only)
- `hide_teamshifts_emails`: email visibility toggle (Team Lead only)

**Form (`team_form.py`)**
- 2-radio role selector replaces 5 radio
- Conditional Team Lead options (role access + hide emails) shown via JS toggle
- Validation: "select at least one role" fires only for Team Lead with "Selected roles only"

**Template (`_permissions_fieldsets.html`)**
- Role selector with inline descriptions
- Per-event role picker with dropdown (groups hidden by default)
- Inline validation error display

**JS (`teamshiftsRoleToggle.js`)**
- Toggles Team Lead options based on role radio selection
- Toggles role selector based on All/Selected radio

PR #4557, opened 2026-07-24
feat: limit registrations for meetup events

### LINKED ISSUE #4560: Limit registrations for meetup events
## Description
There is currently no way to set registration limits or stop RSVPs once a Meetup fills up.

## Current Behavior
- Meetups do not enforce a registration limit.
- The meetup RSVP page keeps accepting registrations even after maximum capacity is reached.

## Expected Behavior
- Organizers can set a registration limit when creating or managing a Meetup event.
- Once a Meetup hits its registration limit, the RSVP page shows a "Registration is closed" message and stops accepting new RSVPs.

prayag78:
## Summary

Adds registration limit management for Meetup events to restrict user RSVPs once an event reaches its maximum capacity.

## Key Changes

- Added a registration limit setting on Meetup creation and settings to control event capacity.
- Displays a "Registration is closed" alert banner on the presale page when maximum registrations are reached.
- Adds row-level locking on quotas during RSVP order creation to prevent overbooking from simultaneous registrations.

Screenshot 2026-08-11 071916
Screenshot 2026-08-11 072016

Fixes #4560


Saksham-Sirohi (changes requested):

1. Migration `0047_team_can_create_meetups_...` conflicts with current `dev` and with #4393’s 0047 — renumber.
2. `_create_rsvp_order()` has no quota check under `select_for_update()`, so concurrent RSVPs can oversell.
3. `JoinOnlineVideoView.validate_access()` only matches `request.user.email`; guest RSVPs stored in session cannot join video while logged out.

Saksham-Sirohi (changes requested):
## Review

### High — Video join IDOR via session order code
In `JoinOnlineVideoView.validate_access`, authenticated users filter with:

`(user_email_match | Q(code=session_order_code))`

Guest RSVP stores another person’s order code in session. A later login as a different user can join video for that order.

**Fix:** If authenticated, authorize **only** by email/attendee match. Use session order code only when anonymous (or require session order email ≡ current user).

### Medium — Inconsistent quota selection
Registration limit / closed UI uses `event.quotas.first()`, while RSVP uses `product.quotas.first()` / `get_rsvp_product_and_quota` (prefers `size__isnull=True`). Multi-quota events can show the wrong closed state.

**Fix:** Always use the same helper for settings, UI, and RSVP.

`select_for_update` inside `transaction.atomic()` for RSVP is a good overbooking fix. Please fix the authz issue before merge.

Rachit7168 (approved):
LGTM , Tested

https://github.com/user-attachments/assets/1b45e448-02e7-4b4a-8a27-05edfc7ab71e

Saksham-Sirohi (changes requested):
## Changes requested

### 1. Inconsistent quota selection (settings initial)
In `eventyay_common/forms/event.py`, registration-limit **initial** still uses `self.event.quotas.first()`, while save / closed UI / RSVP use `get_rsvp_product_and_quota()` (`product.quotas.first()`).

Please load the initial value via the same helper so create, settings, UI, and RSVP always touch the same quota.

### 2. Dead code: unused `is_meetup`
`EventWizardFoundationForm.__init__` pops `is_meetup` and never uses it. Remove the unused kwarg (and stop passing `is_meetup=...` from the view) unless it is wired up for real behavior.

### 3. Missing tests
Please add coverage for the new capacity behavior, for example:
- Setting a registration limit on meetup create/settings updates the RSVP quota size
- Presale shows “Registration is closed” / blocks RSVP when the quota is full
- Concurrent/full RSVP path returns failure when availability is exhausted (at least the locked check path)
- `JoinOnlineVideoView` authz: authenticated users match by email only; anonymous users may use session order code

Saksham-Sirohi (changes requested):
## Review ([PR #4557](https://github.com/fossasia/eventyay/pull/4557))

Capacity UI, settings/create wiring, video-join authz, and tests are in much better shape. A few remaining issues should be fixed before merge.

### Medium — RSVP lock still does not use the shared quota helper

`get_rsvp_product_and_quota()` is now used for settings initial/save and the closed banner, but `_create_rsvp_order()` still does:

```python
quota = product.quotas.select_for_update().first()
```

`.first()` has no guaranteed order. If the meetup product has more than one quota, settings can update quota A while RSVP locks quota B, so the limit can be ignored or the wrong pool is filled.

Use the same helper (then `select_for_update()` on that quota’s PK) inside the `transaction.atomic()` block.

### Medium — Nested `scope(organizer=...)`

`get_rsvp_product_and_quota` / `ensure_rsvp_product` now enter `scope(organizer=event.organizer)` themselves, and callers wrap the same scope again (settings save, create, presale context). Nested organizer scopes often raise at runtime. Keep scoping in one place: either only inside the helper, or only at the call site.

Also: project policy is `scope(event=event)` for event-owned data. Products/quotas are organizer-scoped in the ORM, so organizer scope may be required in the settings form — please document why, and do not nest.

### Medium — Weak settings-form test

`test_settings_form_initial_and_save_updates_quota_size` calls `save()` with hand-built `cleaned_data` when `is_valid()` is false. That hides real validation failures. Assert `bound_form.is_valid()` and only then save.

### Low — Meetup create permission is not organizer-scoped

```python
request.user.teams.filter(can_create_events=True).exists()
```

This is true if the user can create events for *any* organizer. Rely on the wizard’s organizer queryset / a team on the selected organizer.

### Low — Extra create-event behavior

`team.limit_events.add(event)` for limited teams is reasonable, but it is outside the PR title (registration limits). Fine if intentional; call it out in the description.

---

**Looks good:** authenticated video join is email-only; anonymous guests use session order code; closed banner is an `elif` so registered attendees still see success; RSVP POST still has `select_for_update` + availability check.

Saksham-Sirohi (approved):

ArnavBallinCode (approved):

PR #4589, opened 2026-07-26
Fix room webhook autofill handling

### LINKED ISSUE #4559: Webhook configuration fields in room settings are vulnerable to browser autofill
## Observed behavior
 
In the room settings panel where chat can be enabled, the optional webhook configuration section includes a URL field and a secret field. The secret field is rendered as a password-style input, which browsers and password managers commonly recognize as a login credential field regardless of its actual purpose. This makes it a candidate for autofill.
 
In practice, this means a saved password from an unrelated site or form can be silently offered or inserted into the webhook secret field, rather than the field staying empty or reflecting the actual stored webhook value. Neither the URL nor the secret field currently has any explicit signal telling the browser to exclude it from autofill.

Image

 
## Why it matters
 
An organizer could save room settings without noticing the secret field was populated by an autofilled password rather than their intended webhook secret. This would produce a webhook configuration that looks filled in but is actually wrong, causing webhook delivery to fail silently or send to/authenticate against the wrong target.
 
## Expected behavior
 
- The webhook URL and secret fields are excluded from browser and password-manager autofill.
- New rooms with chat enabled show genuinely empty webhook fields until the organizer enters values.
- Existing saved webhook values continue to load and display correctly for a room.
- Saving room settings persists only what the organizer actually entered or what was already saved — never an autofilled or unrelated value.

sarafarajnasardi:
## Summary

  Fixes #4559.

  This PR prevents browser and password-manager autofill from targeting the chat webhook configuration
  fields in room settings.

  ## Changes

  - Replaces the webhook URL and HMAC secret `bunt-input` fields with native inputs inside `bunt-input-
  outline-container` so autofill attributes are applied to the actual browser input.
  - Adds autofill/password-manager suppression attributes to the webhook URL and HMAC secret fields.
  - Keeps existing saved webhook values bound to the same room module config keys.
  - Adds backend validation to reject `webhook_hmac_secret` when no `webhook_url` is provided.
  - Adds a regression test for rejecting orphan webhook secrets.

https://github.com/user-attachments/assets/f4331081-ebfe-4cbd-9d6b-5ec81dd0fbf3


app/eventyay/features/live/modules/room.py
```diff
@@ -107,6 +107,11 @@ async def _verify_webhook_challenges(self, old_module_config, new_module_config)
             cfg = m.get("config", {})
             url = cfg.get("webhook_url")
             secret = cfg.get("webhook_hmac_secret")
+            if secret and not url:
+                raise ConsumerException(
+                    "webhook.missing_url",
+                    "webhook_url is required when webhook_hmac_secret is set.",
```
Rachit7168: Instead of raising a `ConsumerException` here, it would be much better for UX to simply discard the secret if the URL is empty. 

If a user decides to disable a webhook by clearing the URL field, they shouldn't be blocked from saving just because they forgot to clear the obfuscated secret field as well. Additionally, if a rogue password manager still manages to autofill the secret field (despite your excellent frontend mitigations), this exception would completely block the user from saving the room.

We can silently clean up the orphaned secret instead:
```python
            if secret and not url:
                cfg.pop("webhook_hmac_secret", None)
                secret = None
                
            if not url:
                continue
```
This achieves the same data consistency goal without any risk of blocking legitimate user workflows. *(Note: You will also need to update the newly added test in `test_room.py` to assert that the secret is stripped rather than expecting an exception).*

Rachit7168 (changes requested):
This PR does a great job at mitigating the password manager autofill issue on the frontend using standard best practices (`data-1p-ignore`, etc). However, the new backend validation logic in `room.py` introduces a frustrating UX issue that we should improve before merging. Please address the comment below.

app/eventyay/features/live/modules/room.py
sarafarajnasardi: > Instead of raising a `ConsumerException` here, it would be much better for UX to simply discard the secret if the URL is empty.
> 
> If a user decides to disable a webhook by clearing the URL field, they shouldn't be blocked from saving just because they forgot to clear the obfuscated secret field as well. Additionally, if a rogue password manager still manages to autofill the secret field (despite your excellent frontend mitigations), this exception would completely block the user from saving the room.
> 
> We can silently clean up the orphaned secret instead:
> 
> ```python
>             if secret and not url:
>                 cfg.pop("webhook_hmac_secret", None)
>                 secret = None
>                 
>             if not url:
>                 continue
> ```
> 
> This achieves the same data consistency goal without any risk of blocking legitimate user workflows. _(Note: You will also need to update the newly added test in `test_room.py` to assert that the secret is stripped rather than expecting an exception)._

 good idea...thanks ...

ArnavBallinCode (approved):
LGTM

Rachit7168 (approved):
LGTM , TESTED 

https://github.com/user-attachments/assets/acf63161-62b5-4e61-89a9-32b93fb052a4

Saksham-Sirohi (approved):

PR #4774, opened 2026-08-07
Add caching, invalidation, and ETag support to streams/current endpoint

### LINKED ISSUE #4601: `streams/current` endpoint has no cache — every request hits the database
**Component:** `app/eventyay/api/views/room.py`, `app/eventyay/base/models/room.py`  
**Severity:** High — combined with polling (sub-issue 01) this is the direct cause of database saturation during the incident

---

### View: `app/eventyay/api/views/room.py` lines 68–77

```python
@action(detail=True, methods=["get"], url_path="streams/current")
def current_stream(self, request, pk=None, **kwargs):
    room = self.get_object()
    current = room.get_current_stream()
    if current:
        serializer = StreamScheduleSerializer(current)
        return Response(serializer.data)
    return Response(status=404)
```

### ORM query: `app/eventyay/base/models/room.py` lines 392–407

```python
def get_current_stream(self, at_time=None):
    return (
        StreamSchedule.objects.filter(
            room=self, start_time__lte=at_time, end_time__gt=at_time
        )
        .order_by('start_time')
        .first()
    )
```

**Every single call to this endpoint runs a `StreamSchedule` database query.** There is no `cache_page` decorator, no `Cache-Control` header, no ETags, no Redis lookup before hitting the database.

### What the response looks like

A stream schedule does not change mid-second. During an event, the current stream for a room is fixed for the duration of the session (typically 30–90 minutes). Yet every 10-second poll from every client rebuilds this response from the database.

### Comparison with other endpoints

`app/eventyay/agenda/views/schedule.py` uses `@cache_page(60 * 60 * 24)` — 24-hour cache.  
`app/eventyay/features/live/views.py` uses `@method_decorator(cache_page(60), name="dispatch")` — 60-second cache.

The streams endpoint, which is polled far more frequently than those endpoints, has no cache at all.

---

### 1. Add a short Redis cache on `current_stream` (primary fix)

```python
from django.utils.timezone import now

@action(detail=True, methods=["get"], url_path="streams/current")
def current_stream(self, request, pk=None, **kwargs):
    room = self.get_object()
    cache_key = f"stream:current:{room.pk}"
    cached = cache.get(cache_key)
    if cached is not None:
        return Response(cached) if cached != "none" else Response(status=404)
    
    current = room.get_current_stream()
    if current:
        data = StreamScheduleSerializer(current).data
        # Cache until the stream ends (or max 60s)
        ttl = min(60, max(1, int((current.end_time - now()).total_seconds())))
        cache.set(cache_key, data, ttl)
        return Response(data)
    
    # Cache the "no stream" result for 30s to avoid repeated DB misses
    cache.set(cache_key, "none", 30)
    return Response(status=404)
```

### 2. Invalidate cache on stream schedule changes

In `StreamSchedule.save()` and `StreamSchedule.delete()`:

```python
def save(self, *args, **kwargs):
    super().save(*args, **kwargs)
    cache.delete(f"stream:current:{self.room_id}")

def delete(self, *args, **kwargs):
    room_id = self.room_id
    super().delete(*args, **kwargs)
    cache.delete(f"stream:current:{room_id}")
```

### 3. Add HTTP cache headers for anonymous responses

```python
from django.views.decorators.cache import never_cache
from django.utils.cache import patch_cache_control

# For unauthenticated requests (public events):
response = Response(data)
patch_cache_control(response, max_age=30, public=True, stale_while_revalidate=60)
return response
```

For authenticated/restricted events, set `Cache-Control: no-store`.

### 4. Add `Last-Modified` and `ETag` support

If the stream has not changed since the client's last request, return `304 Not Modified` without re-serializing:

```python
from django.utils.http import http_date
import hashlib

if current:
    etag = hashlib.md5(f"{current.pk}:{current.updated_at}".encode()).hexdigest()
    if request.META.get('HTTP_IF_NONE_MATCH') == f'"{etag}"':
        return Response(status=304)
    response = Response(StreamScheduleSerializer(current).data)
    response['ETag'] = f'"{etag}"'
    response['Last-Modified'] = http_date(current.updated_at.timestamp())
    return response
```

### 5. Verify database index on `StreamSchedule`

The query filters on `(room, start_time, end_time)`. Confirm a composite index exists:

```python
class Meta:
    indexes = [
        models.Index(fields=['room', 'start_time', 'end_time']),
    ]
```

If absent, add it and generate a migration.

---

## Expected impact

- Reduces DB queries for `streams/current` from O(requests/second) to at most 1 query per room per 30–60 seconds.
- 10 rooms × 30-second cache = maximum 10 DB queries per 30 seconds instead of potentially thousands.
- Cache miss on stream change is immediate (invalidated on save).

---

## Acceptance criteria

- [ ] `streams/current` for an active stream returns a cached response within 60 seconds of the first request without a DB query.
- [ ] The cache is invalidated within 1 second when a `StreamSchedule` is saved or deleted.
- [ ] `Cache-Control: max-age=30, public` is set on anonymous public event responses.
- [ ] `304 Not Modified` is returned when `ETag` matches.
- [ ] No DB query is made for unauthenticated requests to a public event's stream endpoint after cache warm.
- [ ] Load test shows >90% reduction in DB queries to `StreamSchedule` table during polling traffic.

---

## Related

- Sub-issue 01: Frontend polls this endpoint every 10 seconds with no error backoff
- Sub-issue 07: ORM N+1 queries across high-traffic API endpoints

prajakta128:
Fixes #4601

## Problem

`streams/current` had no caching at all. Every poll (clients hit this endpoint every ~10s) went straight to the database, even though a stream schedule is typically fixed for 30-90 minutes at a time. During high-traffic events this caused unnecessary database load — the same query, repeated by every connected client, every 10 seconds, for an answer that almost never changes.

## What changed

**`api/views/room.py` — `current_stream`**
- Cache the serialized response in Redis under `stream:current:{room.pk}`, with a TTL capped at either the stream's remaining duration or 60 seconds (whichever is shorter), so the cache never outlives the actual stream.
- A "no active stream" result is also cached (30s) to avoid repeated DB misses when no stream is running.
- `Cache-Control: max-age=30, public, stale-while-revalidate=60` is set for anonymous/unauthenticated requests; authenticated requests get `no-store`.
- ETag / `Last-Modified` support: on a cache miss, if the client's `If-None-Match` matches, a `304 Not Modified` is returned instead of resending the body.
- Note: ETag/304 support only applies on cache misses. On a cache hit, `get_current_stream()` is not called again (avoiding a second DB query just to compute the ETag), so cached responses don't get 304 support. This trade-off keeps the cache-hit path to zero DB queries, which is the primary goal of the issue.

**`base/models/stream_schedule.py` — `StreamSchedule.save()` / `delete()`**
- `save()`: added cache invalidation (`cache.delete(f'stream:current:{self.room_id}')`) after the existing save logic — the original validation (`full_clean()`) is untouched, this is purely additive.
- `delete()`: added new, since none existed before. Captures `room_id` before calling `super().delete()` since the instance's fields aren't reliable to read after deletion.

**DB index (required change #5 in the issue)**
- Already present: `StreamSchedule.Meta.indexes` already has `models.Index(fields=['room', 'start_time', 'end_time'])`. No migration needed — confirmed by reading the current model rather than assuming.

## Verification

- `git diff` reviewed line-by-line for both files; no unrelated changes.
- Cache key naming and local-import style (`from django.core.cache import cache` inside the function, not top-of-file) matches existing conventions found elsewhere in the codebase (e.g. `base/services/room.py`, `agenda/apps.py`).
- [ ] Still need to verify locally: first request hits DB, subsequent requests within TTL don't; cache clears within 1s of a StreamSchedule save/delete; response headers show `Cache-Control`/`ETag` correctly. Will update this checklist once tested against a running instance.

## Acceptance criteria from #4601

- [x] `streams/current` returns a cached response within 60s without a DB query (pending local verification)
- [x] Cache invalidated on `StreamSchedule` save/delete
- [x] `Cache-Control: max-age=30, public` set on anonymous responses
- [x] `304 Not Modified` returned when ETag matches (cache-miss path only — see note above)
- [x] Composite index confirmed to already exist
- [ ] Load test showing >90% DB query reduction — not performed; would need a load-testing setup

## Summary by Sourcery

Cache current stream responses and invalidate them when schedules change to reduce database load while supporting HTTP freshness validation.

New Features:
- Add Redis caching for the current stream response, including short-lived caching when no stream is active.
- Add ETag and Last-Modified response support for current stream requests.

Bug Fixes:
- Reduce repeated database queries caused by polling the current stream endpoint.
- Invalidate cached current-stream data when stream schedules are saved or deleted.

Enhancements:
- Apply public browser/proxy caching headers for anonymous requests and disable storage for authenticated requests.
- Add coverage for response caching, cache headers, conditional requests, and schedule invalidation.

Tests:
- Add API tests covering current stream responses, cache population and hits, ETag handling, and save/delete invalidation.


app/eventyay/api/views/room.py
```diff
@@ -69,57 +64,53 @@ def perform_destroy(self, instance):
...
+            patch_cache_control(response, max_age=30, public=True, stale_while_revalidate=60)
+        else:
+            patch_cache_control(response, no_store=True)
+
+        
         if current:
-            serializer = StreamScheduleSerializer(current)
-            return Response(serializer.data)
-        return Response(status=404)
+            etag = hashlib.md5(f"{current.pk}:{current.updated_at}".encode()).hexdigest()
```
Aqil-Ahmad: On cache hits, the ETag never gets set  so 304 only ever works right after a cache miss, not on the fast/cached path that matters most. see acceptance criteria "304 Not Modified is returned when ETag matches," since it's written as an unconditional requirement, not one that only applies right after a miss.

app/eventyay/api/views/room.py
```diff
@@ -69,57 +64,53 @@ def perform_destroy(self, instance):
...
+            else:
+                cache.set(cache_key, "none", 30)
+                return Response(status=404)
+
+        if not request.user.is_authenticated:
+            patch_cache_control(response, max_age=30, public=True, stale_while_revalidate=60)
+        else:
+            patch_cache_control(response, no_store=True)
+
+
```
Aqil-Ahmad: extra blank line over here

Aqil-Ahmad (changes requested):
please add tests

app/eventyay/api/views/room.py
```diff
@@ -69,57 +64,53 @@ def perform_destroy(self, instance):
...
-        return response
-
     @extend_schema(
         summary="Get Current Stream",
         description="Returns the currently active stream schedule for this room, if any.",
         responses={200: StreamScheduleSerializer, 404: None},
     )
     @action(detail=True, methods=["get"], url_path="streams/current")
     def current_stream(self, request, pk=None, **kwargs):
+        from django.core.cache import cache
```
Rachit7168: Thanks for working on the issue!

Please move the imports to above

app/eventyay/api/views/room.py
prajakta128: Fixed — the etag and last_modified are now computed once and cached alongside the response data, so 304 works on cache hits too, not just misses. Thanks for catching that, you were right that the acceptance criteria means this unconditionally.

app/eventyay/api/views/room.py
prajakta128: Fixed, thanks.

app/eventyay/api/views/room.py
prajakta128: Done — moved all five imports to the top of the file.

ArnavBallinCode (changes requested):
Please resolve merge conflicts

ArnavBallinCode (changes requested):

prajakta128:
This is a backend/API change with no visible UI, so the "before/after" shows up in the API response rather than a page. Tested locally: created a test room with an active stream schedule, hit the streams/current endpoint, and confirmed it returns the correct cached response (200, with the stream data) from the updated view.
Screenshot 2026-08-08 005108

Aqil-Ahmad (changes requested):
You're still missing the tests.

Rachit7168 (changes requested):
@prajakta128, for the screen recording and screenshots, please add a larger amount of data and check the response time at each Endpoint. This will help demonstrate how the changes perform with a more realistic dataset.

Saksham-Sirohi:
## Review — not quite complete yet

The caching approach looks sound and matches #4601 (Redis cache + TTL, negative cache, invalidation on save/delete, Cache-Control, ETag on cache hits). A few items still need to land before this is merge-ready.

### Required: tests
Please add automated tests covering at least:

1. **Cache hit** — first request calls `get_current_stream()`, second request within TTL does not.
2. **Invalidation** — after `StreamSchedule.save()` / `.delete()`, the next request refreshes from the DB (updated title / 404).
3. **ETag / 304** — request with matching `If-None-Match` returns `304` on both miss and hit paths.
4. **Cache-Control** — anonymous → `public, max-age=30` (+ `stale-while-revalidate` if kept); authenticated → `no-store`.
5. **Negative cache** — no active stream → `404`, and a follow-up request does not hit the DB again within 30s.

### Suggested hardening (nice to have in this PR)
- **404 responses** currently return early without `Cache-Control` / `ETag`. Consider setting the same headers on the `"none"` path for consistency.
- Note (docs or PR description): natural stream transitions (schedule A ends, B starts with no save) can be stale for up to the TTL (~60s). That matches the issue design, but worth stating explicitly.
- Note: `QuerySet.update()` / CASCADE deletes bypass model `save()`/`delete()`, so they won’t invalidate; TTL is the backstop. Fine if intentional — call it out.
- The video client (`fetchCurrentStream`) does not send `If-None-Match` yet, so 304 won’t help the main poller until that is wired (can be a follow-up). Authenticated polls also get `no-store`, so the real win here is Redis server-side caching — which is still valuable.

### Process
- Please check off the local verification items in the PR description once you’ve confirmed cache hit / invalidation / headers against a running instance.
- Resolve/re-request review after tests are added so we can clear the outstanding “changes requested” reviews.

Happy to re-review once the tests are in.

Saksham-Sirohi (changes requested):
PTAL

prajakta128:
Added tests covering all 5 cases outlined in review — cache hit, invalidation on save/delete, ETag/304 on both miss and hit paths, Cache-Control headers, and negative caching for no-active-stream. See test_api_stream_schedule_cache.py, pushed just now.

Apologies for the delay — I was away from this PR for the past few days due to exams.

Regarding local verification: I was not able to get these tests fully passing in my local environment, but this appears to be due to two pre-existing issues unrelated to this PR's changes:

1. A missing pg_trgm Postgres extension in my local dev database, required by an unrelated index elsewhere in the codebase.
2. Token authentication (Authorization: Token via UserApiToken) is returning 401 locally. To rule out an issue with my new tests specifically, I ran an existing, unrelated test (test_orga_can_see_rooms in test_api_rooms.py) using the same authentication pattern, and it failed the same way. This suggests an environment-specific configuration issue on my end rather than a problem introduced by this branch.

CI should provide a cleaner signal on both fronts. Happy to investigate further if CI results differ from what I'm seeing locally, or if either issue turns out to be something I should address as part of this PR.

Aqil-Ahmad (approved):

app/eventyay/api/views/room.py
```diff
@@ -115,11 +121,41 @@ def export_broadcast_configuration(self, request, **kwargs):
...
+            return Response(status=304)
+
+        response = Response(data)
+        response['ETag'] = f'"{etag}"'
+        response['Last-Modified'] = http_date(last_modified)
+
+        if not request.user.is_authenticated:
+            patch_cache_control(response, max_age=30, public=True, stale_while_revalidate=60)
+        else:
+            patch_cache_control(response, no_store=True)
```
Rachit7168, line 158: Because `no_store=True` tells the browser to *never* store the response locally, the browser will never save the `ETag` and will therefore never send the `If-None-Match` header on its next polling request. This means authenticated users will always receive full 200 OK payloads instead of the optimized 304 Not Modified responses you carefully built!

Consider changing this to:
```python
patch_cache_control(response, no_cache=True, private=True)

prajakta128: Good catch — you're right, no_store was defeating the whole ETag mechanism for authenticated users. Fixed to use no_cache + private, so revalidation (and 304) now works for them too.

Rachit7168:
Thanks for your contribution. 

Closing this in favor of #4992. Please go ahead and feel free to check out other issues and raise a PR. Your contributions are appreciated!

PR #4780, opened 2026-08-07
feat: dual-mode plugin dependency management (dev vs production)

### LINKED ISSUE #4680: Improve plugin dependency workflow for dev/staging and production builds
Eventyay plugins are currently pinned to specific Git commit hashes in `pyproject.toml` and `uv.lock`. This gives reproducible builds, but it also creates manual overhead when plugins are under active development.

We need different behavior for different branches/environments:

* `dev` / staging builds should automatically use the latest plugin commits from the configured development branches.
* `main` / production builds should remain pinned and reproducible through `uv.lock`.

## Problem

* Plugin updates require manual hash bumps and lockfile regeneration.
* This slows down deployment to dev/staging.
* At the same time, production builds must remain reproducible and should not silently pull changing branch heads.
* Docker caching can also cause stale plugin code to be reused if branch-based dependencies are used without cache invalidation.

### 1. Dev/staging builds

For the `dev` branch and staging deployments:

* Use the latest plugin code from the configured development branches.
* Refresh plugin dependencies during Docker builds.
* Prevent Docker from reusing stale dependency layers.
* Make the installed plugin commit versions visible in the build logs.
* This mode may use branch references and cache busting because it is intended for active testing.

### 2. Main/production builds

For the `main` branch and production deployments:

* Use pinned plugin versions from `uv.lock`.
* Do not silently pull latest plugin branch commits.
* Keep builds reproducible.
* Plugin updates for production should happen through an explicit dependency update PR.
* The PR should update plugin refs, regenerate `uv.lock`, run tests, and show which plugin commits changed.

### Dev/staging

* Configure plugin dependencies for dev/staging builds to resolve from the latest configured plugin branches.
* Add Docker cache busting before dependency installation for dev/staging builds.
* Use a dynamic build argument in CI/CD, for example a timestamp or commit SHA.
* Ensure `uv` refreshes branch-based plugin dependencies during dev/staging builds.
* Do not rely on stale Docker cache layers.

Example approach:

```dockerfile
ARG CACHE_BUST=1
RUN uv sync --upgrade
```

The exact command should be adjusted to the project’s `uv` setup. Avoid `--frozen` when the intention is to fetch newer branch commits.

### Main/production

* Keep production Docker builds using the checked-in lockfile.
* Use locked dependencies for production builds.
* Add a CI/manual workflow to update plugin pins for production:

  * fetch latest plugin commits from configured branches
  * update dependency references
  * regenerate `uv.lock`
  * run tests
  * open or update a PR

## Implementation tasks

* Identify all internal plugin dependencies.
* Define which plugin branch is used for dev/staging.
* Update CI/CD so dev/staging builds refresh plugin branches.
* Add Docker cache-busting only for dev/staging builds.
* Ensure main/production builds continue to use pinned lockfile dependencies.
* Add a production plugin update workflow that creates a PR with updated pins and lockfile.
* Log installed plugin commit versions in both dev/staging and production builds.
* Document the difference between dev/staging and production plugin dependency handling.

## Acceptance criteria

* Dev/staging builds use the latest plugin commits from configured development branches.
* Docker cache does not cause stale plugin code in dev/staging builds.
* Main/production builds remain pinned and reproducible.
* Production does not silently pull new plugin branch commits.
* Plugin updates for production happen through an explicit PR.
* `uv.lock` remains the source of truth for production builds.
* Build logs show which plugin commits were installed.
* The workflow is documented for developers and deployment maintainers.

ArnavBallinCode:
Fixes #4680 

Dockerfile.prod: add BUILD_MODE and CACHE_BUST build args. Dev mode refreshes plugins via --upgrade-package, production mode uses --locked. Both modes log installed plugin versions.

Dockerfile: add plugin version logging for local dev builds.

enext-docker.yml: pass BUILD_MODE=dev with CACHE_BUST for dev branch pushes, production mode for main branch pushes.

enext-docker-pr.yml: set BUILD_MODE based on PR target branch.

update-plugin-pins.yml: new workflow to explicitly update production plugin pins, compare commit hashes, validate, and open a PR.

doc/developer/plugin-dependencies.rst: document the dual-mode strategy.

## Summary by Sourcery

Enable branch-aware plugin dependency management so development builds refresh plugin versions while production builds remain reproducibly pinned.

New Features:
- Support separate development and production modes for plugin dependency resolution in Docker builds.
- Add a manually triggered workflow to update pinned plugin commits, validate the resulting lockfile, and create a pull request.

Enhancements:
- Log installed plugin versions during Docker builds.
- Select Docker dependency mode based on the target branch and invalidate development build caches when refreshing plugins.

Build:
- Add build arguments for dependency mode and cache invalidation to production Docker builds.

CI:
- Pass branch-specific plugin dependency settings through development, production, and pull-request Docker workflows.

Documentation:
- Document the dual-mode plugin dependency management strategy.

Tests:
- Validate updated plugin dependencies with environment synchronization and test collection before opening an update pull request.


Rachit7168 (approved):
LGTM

app/Dockerfile.prod
```diff
@@ -23,12 +23,39 @@ RUN wget -qO- https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --
...
-    uv sync --locked --no-install-project --no-editable
+    if [ "$BUILD_MODE" = "dev" ]; then \
+      echo "=== Dev build: refreshing plugins (CACHE_BUST=$CACHE_BUST) ===" && \
+      uv sync --no-install-project --no-editable \
+        --upgrade-package eventyay-paypal \
+        --upgrade-package eventyay-stripe \
+        --upgrade-package eventyay-bitpay \
+        --upgrade-package eventyay-socialmedia \
+        --upgrade-package exhibition \
+        --upgrade-package eventyay-teamshifts; \
```
Saksham-Sirohi: `app/pyproject.toml` currently has eight git plugins. This list is missing **`eventyay-interpretation`** and **`eventyay-hubspot`** (both `@dev`). Dev/staging images will keep those two on the `uv.lock` SHAs and never pick up new commits.

Please add them here (and in the `pip list` grep below), or better: generate `--upgrade-package` args from `uv.lock` / `pyproject.toml` so a new plugin cannot be forgotten.

Also pass `--refresh-package` for each plugin (or `--refresh`). `CACHE_BUST` invalidates the Docker layer, but this `RUN` still uses `--mount=type=cache,target=/root/.cache/uv`, and PR builds enable GHA cache. Without a refresh, uv can reuse a stale git clone.

app/Dockerfile.prod
```diff
@@ -23,12 +23,39 @@ RUN wget -qO- https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --
...
+    else \
+      echo "=== Production build: using locked dependencies ===" && \
+      uv sync --locked --no-install-project --no-editable; \
+    fi
+
+# Log installed plugin versions for build traceability
+RUN echo "=== Installed plugin versions ===" && \
+    .venv/bin/pip list 2>/dev/null \
+      | grep -iE 'eventyay-paypal|eventyay-stripe|eventyay-bitpay|eventyay-socialmedia|exhibition|eventyay-teamshifts' \
+    || echo "(no plugins found — check build output above)"
```
Saksham-Sirohi, line 60: `pip list` only prints the package version from the plugin's own metadata (e.g. `1.0.0`), not the installed git commit. #4680 asks for commit versions in the build log.

Please log the resolved git SHAs, for example `uv pip freeze` (shows `pkg @ git+https://…@<sha>`) or the `source = { git = "…#<sha>" }` lines from `uv.lock` after sync.

.github/workflows/update-plugin-pins.yml
```diff
@@ -0,0 +1,182 @@
...
+# These correspond to the git-based dependencies in pyproject.toml.
+env:
+  PLUGINS: >-
+    eventyay-paypal
+    eventyay-stripe
+    eventyay-bitpay
+    eventyay-socialmedia
+    exhibition
+    eventyay-teamshifts
+
```
Saksham-Sirohi, line 31: Same gap as `Dockerfile.prod`: `eventyay-interpretation` and `eventyay-hubspot` are in `app/pyproject.toml` / `uv.lock` but not here, so production pin PRs will never bump them.

Please keep this list in sync with the Docker build (ideally one source of truth derived from the lockfile).

.github/workflows/update-plugin-pins.yml
```diff
@@ -0,0 +1,182 @@
...
+        if: inputs.dry_run == 'true'
+        run: |
+          echo "${{ steps.changelog.outputs.body }}"
+          echo ""
+          echo "Dry run complete — no PR will be created."
+
+      - name: Create or update PR
+        if: steps.changelog.outputs.has_changes == 'true' && inputs.dry_run != 'true'
+        uses: peter-evans/create-pull-request@v7
+        with:
```
Saksham-Sirohi, line 162: `|| true` means test collection can fail and this workflow will still open the pin-update PR. Please drop `|| true` so validation actually gates PR creation (or move PR creation behind a job that requires this step to succeed).

.github/workflows/update-plugin-pins.yml
```diff
@@ -0,0 +1,182 @@
...
+          body: |
+            ${{ steps.changelog.outputs.body }}
+
+            ---
+
+            This PR was automatically generated by the `update-plugin-pins` workflow.
+
+            **Review checklist:**
+            - [ ] Verify the plugin commit changes are expected
+            - [ ] Confirm CI tests pass
```
Saksham-Sirohi, line 176: This repo's integration branch is `dev`, then `main`. Opening pin PRs directly against `main` skips staging unless that is an explicit exception.

Please either target `dev`, or document why production pins bypass `dev`. Also check out the same branch you open the PR against (`ref: main` above will otherwise be wrong if the base changes).

doc/developer/plugin-dependencies.rst
```diff
@@ -0,0 +1,128 @@
...
+
+======================  ====================================  =========
+Plugin package          Repository                            Branch
+======================  ====================================  =========
+``eventyay-paypal``     ``fossasia/eventyay-paypal``           ``main``
+``eventyay-stripe``     ``fossasia/eventyay-stripe``           ``main``
+``eventyay-bitpay``     ``fossasia/eventyay-bitpay``           ``main``
+``eventyay-socialmedia``  ``fossasia/eventyay-socialmedia``    ``main``
+``exhibition``          ``fossasia/eventyay-exhibition``       ``main``
+``eventyay-teamshifts`` ``fossasia/eventyay-teamshifts``       ``main``
```
Saksham-Sirohi: This table is incomplete and the branches are wrong vs `app/pyproject.toml`:

- `eventyay-socialmedia`, `exhibition`, `eventyay-teamshifts` track **`dev`**, not `main`
- missing **`eventyay-interpretation`** (`@dev`) and **`eventyay-hubspot`** (`@dev`)

Please match the real refs so operators know which branch `--upgrade-package` will follow.

Saksham-Sirohi (changes requested):
Thanks for this — the dual-mode approach (`BUILD_MODE=dev` + `--upgrade-package` vs production `--locked`) is the right design for #4680. It is not complete enough to merge yet: a `dev` image will still silently keep two plugins on lockfile pins, and a few other acceptance criteria from the issue are not met.

Please address the inline comments. Highest priority:

1. Include **all** git plugin deps from `app/pyproject.toml` (currently missing `eventyay-interpretation` and `eventyay-hubspot`), or derive the list from the lockfile so it cannot drift.
2. Log **git commit SHAs**, not `pip list` versions.
3. Do not swallow validation failures (`|| true`) before opening the pin-update PR.

Happy to re-review once those are in.

Saksham-Sirohi (changes requested):
There is no detector. After a dev image is deployed, the containers keep the plugin commits that were baked in at build time. Nothing on the server watches plugin repos, nothing re-runs uv sync, and Watchtower only pulls a new Eventyay image from Docker Hub (every 10 minutes). It does not notice that eventyay-stripe got a new git commit.

--upgrade-package and CACHE_BUST run only when Eventyay’s image is built (push to Eventyay dev/main, or a manual “Docker” workflow run). CACHE_BUST stops Docker from reusing an old build layer. It does not invalidate a container that is already live.

Saksham-Sirohi (approved):
Let's do the auto change detection and update in a follow-up LGTM as auto updater plugins on a PR merge

mariobehling (approved):

PR #4784, opened 2026-08-07
perf: optimize start page event loading and image delivery

### LINKED ISSUE #4781: Optimize start page event loading and image delivery
The public start page can be improved in two main areas:

1. event-card images are loaded more aggressively and at larger dimensions than necessary; and
2. the backend loads more event records than required before limiting each section to the displayed number of events.

Example:

`https://wikimedia.eventyay.com/`

The start page architecture is otherwise relatively lightweight and should remain server-rendered.

## Current behavior

The start page displays sections such as:

* Featured events
* Upcoming events
* Upcoming events from followed organizers
* Past events

Each section is limited to a small number of events in the final rendered page.

However, the backend currently retrieves all matching visible events, converts the queryset to a list, categorizes the events in Python, and only afterwards limits the sections to eight events.

Conceptually:

```python
events = list(qs.order_by('date_from'))

for event in events:
    ...
    featured_events.append(event)
    upcoming_events.append(event)
    past_events.append(event)

featured_events = featured_events[:8]
upcoming_events = upcoming_events[:8]
past_events = past_events[-8:]
```

This means that as the installation accumulates more historical and upcoming events, the amount of work required to render the start page can grow even though only a small fixed number of events are displayed.

Event-card images also currently use a basic `` element without:

* lazy loading;
* asynchronous decoding;
* explicit width and height;
* responsive `srcset`;
* responsive `sizes`.

## Goals

The start page should:

* remain fast as the number of events on the installation grows;
* query only events that are actually required for the page;
* avoid downloading below-the-fold event images immediately;
* serve images close to their rendered size;
* avoid layout shifts caused by images;
* use modern compressed image formats where practical; and
* remain server-rendered and accessible.

### 1. Limit event queries in the database

Do not retrieve all visible events before deciding which ones belong on the homepage.

Use separate bounded queries for the homepage sections.

For example:

### Featured events

Retrieve only the next eight eligible featured events:

```python
featured_events = (
    Event.objects
    .filter(
        live=True,
        startpage_featured=True,
        ...
    )
    .order_by('date_from')[:8]
)
```

### Upcoming events

Retrieve only the next eight non-featured visible events:

```python
upcoming_events = (
    Event.objects
    .filter(
        live=True,
        startpage_visible=True,
        startpage_featured=False,
        ...
    )
    .order_by('date_from')[:8]
)
```

### Past events

Retrieve only the eight most recent past events:

```python
past_events = (
    Event.objects
    .filter(
        live=True,
        startpage_visible=True,
        ...
    )
    .order_by('-date_from')[:8]
)
```

Avoid loading old event records that cannot appear on the start page.

The amount of database and Python work for rendering the homepage should remain approximately constant as historical event count grows.

### 2. Move test-mode filtering into the queryset

The homepage currently retrieves matching events and then excludes component test-mode events in Python.

Where possible, apply the same database-side filtering already used by the dedicated upcoming and past event pages.

For example:

```python
.exclude(
    _settings_objects__key='talks_testmode',
    _settings_objects__value='True'
)
```

Avoid loading event objects that are guaranteed not to be displayed.

### 3. Lazy-load event-card images

Images below the initial viewport should not be requested immediately.

Use native image loading:

```html
<img
  src="..."
  alt="..."
  loading="lazy"
  decoding="async"
  width="800"
  height="450"
>
```

Do not necessarily lazy-load the image that becomes the page's Largest Contentful Paint element.

For the first visible event image, use normal eager loading or `fetchpriority="high"` only if performance measurements show this improves LCP.

All event cards further down the page should use lazy loading.

### 4. Add explicit image dimensions

Event previews use a 16:9 ratio and locally generated previews currently use approximately 800 × 450 px.

Add explicit dimensions to the `` element so that the browser can reserve layout space before the image is loaded.

For example:

```html
width="800"
height="450"
```

The responsive CSS can continue to determine the displayed width.

This should reduce Cumulative Layout Shift.

### 5. Add responsive preview image variants

The existing preview-image implementation already creates a locally cached 800 × 450 thumbnail.

Add at least one smaller variant for event cards.

For example:

* 400 × 225
* 800 × 450

Then use:

```html
<img
  src="/preview/400x225.webp"
  srcset="
    /preview/400x225.webp 400w,
    /preview/800x450.webp 800w
  "
  sizes="(max-width: 600px) 100vw, 400px"
  ...
>
```

The browser should not download an 800 px image when a significantly smaller image is sufficient.

### 6. Use modern image formats

Generated start-page preview thumbnails should use efficient image formats where practical.

Prefer:

* WebP;
* optionally AVIF; and
* appropriate fallback behavior.

Image processing should:

* strip unnecessary metadata;
* resize before delivery;
* preserve transparency only when required;
* avoid large PNG files for photographic event imagery; and
* use a sensible quality setting.

### 7. Keep the page server-rendered

Do not convert the homepage into an API-driven SPA as part of this optimization.

The existing server-rendered design is appropriate for:

* accessibility;
* SEO;
* fast first render;
* event discovery; and
* simple navigation.

The optimization should reduce server work and asset transfer while preserving the existing architecture.

## Search

The existing event search already uses limited results and should remain separate from homepage section loading.

No major search redesign is required as part of this issue.

If event counts become very large in the future, search performance can be evaluated separately.

## Performance investigation

Before implementing changes, measure the current Wikimedia eventyay start page.

Record:

* server response time;
* database query count;
* number of event objects retrieved;
* number of event cards rendered;
* HTML transfer size;
* number of image requests started immediately;
* image transfer size;
* largest individual event image;
* total initial transfer size;
* Largest Contentful Paint;
* Cumulative Layout Shift; and
* number of event records evaluated in Python.

Also test with a database containing a large number of past events.

The homepage render cost should not grow linearly with the total historical event count.

### Backend

* [ ] Featured events are limited in the database.
* [ ] Upcoming events are limited in the database.
* [ ] Past events are limited in the database.
* [ ] Rendering the start page does not load all historical events.
* [ ] Component test-mode events are filtered before unnecessary model objects are instantiated where possible.
* [ ] Existing start-page visibility rules remain unchanged.
* [ ] Existing featured-event behavior remains unchanged.
* [ ] Followed-organizer events continue to work.
* [ ] Search results continue to work.

### Images

* [ ] Event-card images below the initial viewport use lazy loading.
* [ ] Images use `decoding="async"` where appropriate.
* [ ] Event preview images have explicit width and height.
* [ ] At least one smaller responsive preview size is available.
* [ ] Event cards use `srcset` and `sizes` where supported.
* [ ] Modern compressed thumbnail formats are used where practical.
* [ ] The first/LCP image is not unnecessarily delayed by lazy loading.
* [ ] Missing event images continue to use the existing placeholder.

### Performance

* [ ] The PR documents the number of event objects retrieved before and after the change.
* [ ] The PR documents start-page response time before and after the change.
* [ ] The PR documents initial image request count before and after the change.
* [ ] The PR documents initial image transfer size before and after the change.
* [ ] The PR documents LCP and CLS before and after the change.
* [ ] Performance remains stable when the number of historical events is increased substantially.

## Test scenarios

Test at minimum:

1. No featured events.
2. More than eight featured events.
3. More than eight upcoming events.
4. Hundreds of past events.
5. Event with preview image.
6. Event using header image as preview fallback.
7. Event using logo as preview fallback.
8. Event without any image.
9. Component test-mode event.
10. Authenticated user following organizers.
11. Mobile viewport.
12. High-DPI display.
13. Slow network.
14. Cached image revisit.
15. Search results page.

## Scope

The main objective is to keep the public start page lightweight as an installation grows.

The start page should:

> query only the event records it intends to display and download only the image assets needed for the current viewport.

This issue should remain separate from the larger schedule and speaker-directory optimization work.

SxxAq:
Fixes #4781

## Overview

This PR improves the performance and scalability of the public start page (`/`) by optimizing database queries, reducing unnecessary data retrieval, and improving image delivery for better Core Web Vitals.

### 1. Optimized database queries

Refactored `StartPageView.get_context_data` in `app/eventyay/presale/views/startpage.py` to avoid loading all live events into Python memory using `list(qs.order_by('date_from'))`.

Instead, the page now uses four independent, database-bounded queries (each limited with `[:12]`) for:

* Featured events
* Upcoming events
* Past events
* Followed upcoming events

Additional improvements include:

* Moved test mode filtering (`testmode=False` and `talks_testmode=False`) into the database query instead of filtering in Python.
* Used `.only(...)` to fetch only the fields required by the start page cards:

  * `id`
  * `organizer_id`
  * `name`
  * `slug`
  * `date_from`
  * `date_to`
  * `location`
  * `live`
  * `is_public`
  * `testmode`
  * `startpage_visible`
  * `startpage_featured`

This reduces unnecessary data transfer and memory usage while keeping query performance constant as the number of events grows.

### 2. Improved image delivery

Added a new `preview_image_url_small` property to the `Event` model to serve optimized 400×225 thumbnails.

Updated `startpage_event_card.html` to improve loading performance by adding:

* `loading="lazy"` for images below the fold
* `fetchpriority="high"` for the primary (LCP) event image
* `decoding="async"`
* Explicit `width="800"` and `height="450"` attributes to prevent layout shifts
* Responsive `srcset` (`400w`, `800w`) and `sizes` attributes so smaller images are served on mobile devices

These changes improve image loading efficiency and help reduce CLS and improve LCP.

## Testing

Added a dedicated test suite in `app/tests/tickets/presale/test_startpage_optimization.py` covering:

* `test_startpage_sections_rendering`
* `test_startpage_event_card_image_attributes`
* `test_startpage_bounded_query_scaling`

The scaling test verifies that query bounds remain constant even when the database contains 30+ past events.

### Test Results

```text
tests/tickets/presale/test_startpage_optimization.py
...                                                             [100%]

3 passed in 15.21s
```

## Summary by Sourcery

Optimize start-page event retrieval and image delivery to improve scalability and Core Web Vitals.

New Features:
- Add responsive small preview image delivery for start-page event cards.

Enhancements:
- Bound start-page event sections with database-level filtering and limits to reduce memory use and improve scalability.
- Improve start-page image loading with responsive sources, lazy loading, prioritization, and layout-stability attributes.
- Use consistent timezone-aware day-boundary filtering across event listing views.

Tests:
- Add coverage for start-page section rendering, image performance attributes, and bounded query scaling.


Rachit7168 (changes requested):
For the screen recording and screenshots, please add a larger amount of data and check the response time at each Endpoint. This will help demonstrate how the changes perform with a more realistic dataset.

app/eventyay/presale/views/startpage.py
```diff
@@ -73,54 +73,65 @@ def get_context_data(self, **kwargs):
...
+                    'date_from',
+                    'date_to',
+                    'location',
+                    'live',
+                    'is_public',
+                    'testmode',
+                    'startpage_visible',
+                    'startpage_featured',
+                )
+            )
```
ViRUS-0-0, line 82: N+1 query regression caused by deferred `header_image` field in `.only(...)`. Each event card in `startpage_event_card.html`
 evaluates `event.social_image_signature`, `preview_image_url_with_fallback`, and `preview_image_url_small`. These methods check `self._visible_header_image_path`, which accesses the model field `self.header_image`. Because `header_image` is missing from `.only(...)`, accessing it triggers an extra SQL query per event to fetch the deferred field.

Consider adding 'header_image' to the .only(...) field list to prevent N+1 queries.

app/eventyay/presale/views/startpage.py
```diff
@@ -73,54 +73,65 @@ def get_context_data(self, **kwargs):
...
+            future_filter = Q(date_to__gte=today) | Q(date_to__isnull=True, date_from__gte=today)
+            past_filter = Q(date_to__lt=today) | Q(date_to__isnull=True, date_from__lt=today)
```
ViRUS-0-0: Comparing `DateTimeField` attributes (`date_from` and `date_to`) against a `datetime.date` object (`today`) without using the `__date` lookup transform.

Resolution: Use the `__date` field lookup transform for `DateTimeField` date comparisons.

app/eventyay/presale/views/startpage.py
```diff
@@ -73,54 +73,65 @@ def get_context_data(self, **kwargs):
...
+
+            featured_qs = base_qs.filter(startpage_featured=True).filter(future_filter).order_by('date_from')[:12]
```
ViRUS-0-0: Redundant DB slicing `[:12]` and post-processing list comprehension. 
Since `base_qs` already excludes `testmode=False` and `talks_testmode='True'`, no elements are filtered by `[e for e in qs if not e.has_component_testmode]`. 
Fetching 12 items from DB only to slice `[:8]` in Python executes unnecessary database work on every request.

Resolution: Slice `[:8]` directly in the ORM queryset.

ViRUS-0-0 (changes requested):

app/eventyay/presale/views/startpage.py
SxxAq: fixed

app/eventyay/presale/views/startpage.py
```diff
@@ -73,54 +73,66 @@ def get_context_data(self, **kwargs):
...
-            ctx['upcoming_events'] = upcoming_events[:8]
-            ctx['past_events'] = list(reversed(past_events))[:8]
+            base_qs = (
+                Event.objects.select_related('organizer')
+                .prefetch_related('_settings_objects')
+                .filter(live=True, testmode=False)
+                .exclude(_settings_objects__key='talks_testmode', _settings_objects__value='True')
+                .only(
+                    'id',
+                    'organizer_id',
```
ArnavBallinCode: The `.only()` optimisation looks potentially risky here. These Event objects are subsequently consumed by the template and image-related properties, so any deferred field access could turn this into per-event queries and offset the intended performance improvement. Could we avoid .only() here unless we're certain all fields accessed during rendering are covered?

app/eventyay/presale/views/startpage.py
```diff
@@ -73,54 +73,66 @@ def get_context_data(self, **kwargs):
...
-            qs = Event.objects.select_related('organizer').prefetch_related('_settings_objects').filter(live=True)
-            qs = qs.filter(Q(startpage_visible=True) | Q(startpage_featured=True))
-            events = list(qs.order_by('date_from'))
-            visible_events = [event for event in events if not event.has_component_testmode]
-
             today = timezone.localdate()
-            featured_events = []
-            upcoming_events = []
-            past_events = []
-
```
ArnavBallinCode, line 86: One concern with the new `base_qs` is the combination of `prefetch_related('_settings_objects')` with the `_settings_objects` exclusion. Since we're now filtering through this relation at the DB level, could we make sure we're not adding unnecessary JOIN/prefetch work that offsets the query optimisation?

app/eventyay/presale/views/startpage.py
```diff
@@ -73,54 +73,66 @@ def get_context_data(self, **kwargs):
...
+                    'is_public',
+                    'testmode',
+                    'startpage_visible',
+                    'startpage_featured',
+                    'header_image',
+                )
+            )
+            future_filter = Q(date_to__date__gte=today) | Q(date_to__isnull=True, date_from__date__gte=today)
+            past_filter = Q(date_to__date__lt=today) | Q(date_to__isnull=True, date_from__date__lt=today)
+
```
ArnavBallinCode: The switch from direct `datetime` comparisons to `__date` is worth reconsidering from a DB-performance perspective. Applying a date conversion to the column can prevent effective use of an index on date_from/date_to. Can we keep the comparison against a datetime boundary where possible?

Rachit7168 (approved):
LGTM , Changes look fine

MukundC25 (commented):
@SxxAq PR looks approvable, please resolve branch conflicts once

ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):

PR #4785, opened 2026-08-07
perf: replace schedule-JSON speakers page with server-rendered paginated view

### LINKED ISSUE #4776: Optimize public speakers page payload, images, and loading performance
Public speaker overview pages can become very large and slow for events with many speakers.

Example:

`https://wikimedia.eventyay.com/wm/wikimania2025/speakers/`

The current implementation renders the speaker overview through the schedule web component and embeds schedule data into the page before initializing the speakers view.

For large events, this can result in unnecessarily large:

* HTML and embedded JSON payloads;
* image transfers;
* JavaScript parsing and rendering work;
* browser memory usage; and
* server-side serialization work.

The speaker directory should not need to load the complete event schedule and full speaker data merely to display a grid or list of speakers.

## Current behavior

The speakers page embeds `schedule_json` into the HTML and initializes the schedule component using the speakers view.

The schedule data can contain information such as:

* speakers;
* biographies;
* original avatar URLs;
* default avatar thumbnail URLs;
* tiny avatar thumbnail URLs;
* talks;
* rooms;
* tracks;
* feature flags; and
* other schedule metadata.

For a speaker overview, much of this information is not required for the initial render.

Speaker images also currently have a large size gap between the available thumbnail variants:

* `tiny`: 64 × 64 px
* `default`: 460 × 460 px

For speaker cards that are typically displayed around 100–200 px wide, 64 px may be too small while 460 px is unnecessarily large.

Generated thumbnails also preserve the original image format. As a result, photographic PNG uploads can remain substantially larger than necessary.

## Goals

The public speakers page should:

* load quickly even for events with hundreds of speakers;
* transfer only data needed for the current view;
* avoid loading images outside or far beyond the viewport;
* use appropriately sized and compressed speaker images;
* avoid embedding full biographies and schedule data unnecessarily; and
* remain searchable, accessible, crawlable, and shareable.

### 1. Introduce a lightweight speaker-list payload

The speakers overview should use a dedicated payload rather than the complete schedule payload.

The initial speaker entry should contain only information required by the speaker card, for example:

```json
{
  "code": "ABC123",
  "name": "Speaker Name",
  "avatar": "/path/to/list-thumbnail.webp",
  "is_featured": false
}
```

Depending on the design, limited additional fields such as affiliation or a short subtitle may be included.

The initial payload should not include information that is not displayed on the overview page, such as:

* full biographies;
* complete session descriptions;
* resources;
* all room data;
* all track data;
* multiple unused avatar URLs; or
* unrelated schedule metadata.

Full speaker details should be loaded only when the speaker profile is opened.

### 2. Do not load the complete speaker directory initially

Load speakers in batches.

Suggested initial size:

* approximately 40–60 speakers.

Additional speakers can be loaded using:

* infinite scrolling; or
* a **Load more speakers** action.

Pagination must be applied at the database/query level rather than retrieving every speaker and only hiding them in the frontend.

The server response time should therefore depend primarily on the requested page size and not on the total number of speakers.

### 3. Add server-side search

For large speaker directories, search should not depend on every speaker already being loaded in the browser.

Support server-side search by relevant fields such as:

* speaker name;
* affiliation, if available;
* Wikimedia username or other configured public speaker identifiers, where applicable.

Changing the search should reset pagination.

Search URLs should preferably remain shareable, for example:

```text
/speakers/?q=smith
```

### 4. Lazy-load speaker images

Images outside the initial viewport should not be downloaded immediately.

Speaker images should use native lazy loading at minimum:

```html
<img
  loading="lazy"
  decoding="async"
  ...
>
```

Opening a page containing hundreds of speakers must not immediately trigger requests for hundreds of avatar images.

Images should load as the visitor approaches the corresponding cards.

### 5. Add an appropriately sized speaker-list thumbnail

Introduce an intermediate thumbnail size optimized for speaker cards.

For example:

* tiny: 64 × 64 px
* list: 192 × 192 px or 256 × 256 px
* default: 460 × 460 px

The exact dimensions should be based on the maximum rendered card size and expected device pixel ratio.

The speaker overview should use the new list thumbnail instead of the original image or 460 px thumbnail wherever possible.

### 6. Generate modern image formats

Generated speaker thumbnails should use efficient delivery formats.

Prefer:

* WebP; and/or
* AVIF where supported and operationally suitable.

For photographic images without transparency, avoid retaining PNG merely because the original upload was PNG.

Image processing should:

* remove unnecessary metadata;
* resize before delivery;
* use appropriate quality settings;
* preserve transparency only where needed; and
* avoid unnecessarily large files.

### 7. Use responsive images

Speaker cards should provide appropriately sized image variants using `srcset` and `sizes`.

Example:

```html
<img
  src="/avatars/ABC/list.webp"
  srcset="
    /avatars/ABC/tiny.webp 64w,
    /avatars/ABC/list.webp 192w,
    /avatars/ABC/default.webp 460w
  "
  sizes="(max-width: 600px) 96px, 160px"
  loading="lazy"
  decoding="async"
  width="160"
  height="160"
  alt="Speaker Name"
>
```

The browser should not download a 460 px image when the rendered card only needs approximately 100–160 px.

### 8. Avoid biographies in the speaker-directory payload

Speaker biographies can contain multiple paragraphs and substantially increase embedded JSON size when an event has many speakers.

The speaker overview should not download biographies unless they are actually displayed.

If the overview later requires a short description, consider a separately defined short speaker summary rather than transferring the complete biography for every speaker.

### 9. Avoid full schedule data on the speakers page

Review `build_speakers_list_schedule_json()` and the schedule web component.

The speakers page should not require the complete released schedule merely because the same frontend component also supports schedule views.

Specifically measure whether `/speakers/` currently transfers:

* all talks;
* all tracks;
* all rooms;
* all speakers;
* all biographies; and
* other schedule-only metadata.

Split the speaker directory data model from the full schedule data where possible.

### 10. Improve caching

Speaker thumbnails and other immutable or versioned assets should use effective browser and CDN caching.

Where URLs already contain cache-busting/version identifiers, use long-lived cache headers.

The page should not repeatedly transfer unchanged speaker images on subsequent visits.

## Performance measurements

Before implementing changes, record a baseline for a large event such as Wikimania 2025.

Measure:

* HTML document size;
* compressed and uncompressed embedded schedule JSON size;
* number of speakers in the initial payload;
* number of image requests triggered during initial load;
* total image transfer size;
* largest speaker images;
* total page transfer size;
* number of network requests;
* server response time;
* Time to First Byte;
* Largest Contentful Paint;
* time until speaker cards become interactive; and
* browser memory usage where practical.

Separately measure the size of:

```html
<script id="pretalx-schedule-data" type="application/json">
```

This will show how much of the page size originates from schedule serialization rather than image assets.

## Suggested performance targets

For a large speaker directory:

* initial HTML + speaker data: preferably below 300–500 KB compressed;
* only the first batch of speaker data should be transferred initially;
* initial speaker-image transfer should preferably stay below approximately 1 MB;
* individual list thumbnails should generally remain in the tens of KB rather than hundreds of KB or several MB;
* images well below the viewport should not be downloaded during initial load;
* full biographies should not be included in the initial directory payload;
* initial request duration should not grow linearly with the total number of speakers.

These values are targets rather than strict limits and should be validated against the actual design and device requirements.

### Data payload

* [ ] The speakers overview no longer requires the complete event schedule payload.
* [ ] Only fields required for speaker cards are included in the initial speaker response.
* [ ] Full biographies are not transferred unless they are required by the displayed view.
* [ ] Unrelated talk, room, and track data is excluded from the speaker-directory payload.
* [ ] Only the first page/batch of speakers is loaded initially.
* [ ] Additional speakers can be loaded without refreshing the page.
* [ ] Pagination is performed on the backend.

### Images

* [ ] Speaker images use an appropriately sized list thumbnail.
* [ ] Speaker images below the viewport are lazy-loaded.
* [ ] Speaker cards provide explicit image dimensions to reduce layout shifts.
* [ ] Responsive image variants are available through `srcset`/`sizes`.
* [ ] Modern compressed thumbnail formats are used where practical.
* [ ] Photographic PNG uploads are not unnecessarily delivered as large PNG thumbnails.
* [ ] Image metadata is removed during processing where applicable.
* [ ] Browser/CDN caching is configured effectively.

### Search and navigation

* [ ] Speaker search works without loading every speaker first.
* [ ] Search operates on the backend.
* [ ] Pagination resets when search criteria change.
* [ ] Search state can be represented in the URL.
* [ ] Speaker profile URLs remain directly accessible and crawlable.

### Performance

* [ ] Initial network requests do not include every speaker image.
* [ ] Initial response size is substantially smaller for large events.
* [ ] Page response time does not scale linearly with the total speaker count.
* [ ] The Wikimania 2025 speaker page is tested before and after the change.
* [ ] HTML, JSON, image, and total transfer sizes are recorded in the PR.
* [ ] Lighthouse/Web Vitals or equivalent measurements are included in the PR for comparison.

## Test scenarios

Test at minimum:

1. Small event with fewer than one page of speakers.
2. Large event with hundreds of speakers.
3. Speakers with JPEG portraits.
4. Speakers with PNG portraits.
5. Speakers without portraits.
6. Speakers with very long biographies.
7. Speaker search returning many results.
8. Speaker search returning no results.
9. Scrolling through multiple batches of speakers.
10. Reloading after images are cached.
11. Mobile viewport.
12. High-DPI display.
13. Slow network connection.
14. JavaScript-disabled or fallback behavior where supported.
15. Speaker cards with missing or failed image requests.

## Scope

The main objective of this issue is not simply to compress existing images.

The speaker directory should be changed so that **data and assets are loaded according to what the visitor actually needs**, rather than treating the speaker overview as another rendering mode of the complete schedule dataset.

ArnavBallinCode:
Fixes #4776

## Description

This PR resolves the critical performance bottlenecks on the public speaker overview pages for large events. By replacing the heavy full-schedule JSON injection with a lightweight, paginated JSON API, we preserve the original UI while drastically improving load times and reducing payload sizes.

## What was optimized?

The previous implementation forced the backend to serialize the *entire* event schedule (all talks, rooms, tracks, and full speaker profiles) into a massive `schedule_json` blob injected into the HTML. This PR optimizes several key areas:

1. **HTML and embedded JSON payloads**: The massive `<script id="pretalx-schedule-data">` blob is completely removed from the DOM on the speakers page. The page HTML is now tiny and loads instantly.
2. **Server-side serialization work**: The backend no longer has to recursively query and serialize the entire schedule graph just to show a list of speakers. We now use a targeted `build_speaker_cards()` utility that queries only the necessary fields for a single page of speakers.
3. **Image transfers**: Instead of the browser eagerly trying to load all speaker avatars at once (or failing to lazy-load them efficiently from a massive DOM), the avatars are now fetched incrementally as the user scrolls.
4. **JavaScript parsing and rendering work**: The frontend Vue component (`SpeakersList.vue`) no longer freezes the main thread parsing a multi-megabyte JSON object. It now parses tiny, paginated API responses.
5. **Browser memory usage**: The browser only holds a fraction of the event data in memory at any given time, gracefully expanding the list via infinite scroll using `IntersectionObserver`.

## Implementation Details
- **Backend**: Added a `?format=json` response handler to the `SpeakerList` Django view, utilizing the existing `Filterable` ListView for native database-level pagination (48 speakers per page).
- **Frontend**: Preserved the original `<pretalx-schedule view="speakers">` component and its styling, but updated the inner `SpeakersList.vue` to fetch data asynchronously via the new API endpoint.
- **Cleanup**: Reverted previous experimental commits that attempted to switch to a pure Django server-rendered UI. This ensures the Vue layout, search, and filtering behaviors remain 100% intact.

https://github.com/user-attachments/assets/b0032d42-96d7-4c62-b740-71148a5549d0

## Summary by Sourcery

Replace full schedule injection on the public speakers page with a lightweight, server-paginated data flow while retaining the existing UI and interactions.

New Features:
- Add a paginated JSON endpoint for public speaker data, including profile, session, filtering, and pagination metadata.
- Support asynchronous speaker loading with infinite scroll while preserving the existing Vue speakers interface and filtering behavior.

Bug Fixes:
- Prevent private speaker information, unpublished sessions, and data from other events from appearing in public speaker results.

Enhancements:
- Reduce speakers-page HTML payload, server-side serialization, browser parsing, and memory usage by replacing full schedule data with targeted speaker metadata and cards.
- Improve speaker search, sorting, language and track filtering, avatar loading, and session display for API-backed results.

Tests:
- Add coverage for paginated speaker responses, public-field handling, search and filter behavior, localization, pagination links, and HTML payload contents.


Saksham-Sirohi (commented):
Please add a screencast to this even though this is a draft

ArnavBallinCode:
@sourcery-ai re-review

app/eventyay/agenda/views/utils.py
```diff
@@ -110,6 +110,66 @@ def speaker_profile_display_order():
...
+        user = profile.user
+        
+        # find sessions for this speaker
+        speaker_sessions = []
+        for talk in talks:
+            if user in talk.submission.speakers.all():
+                speaker_sessions.append({
+                    'id': talk.submission.code,
+                    'title': talk.submission.title,
+                })
```
SxxAq: Currently, build_speaker_cards() iterates over all schedule talks for every speaker profile on the page, leading to $O(M \times N)$ complexity.

We can optimize this to $O(N + M)$ linear time by constructing a speaker_id -> sessions lookup dictionary first and then iterating over profiles.

app/eventyay/agenda/views/utils.py
```diff
@@ -110,6 +110,66 @@ def speaker_profile_display_order():
...
+    """Return responsive avatar variants for a speaker card, or ``None``.
+
+    Uploaded avatars expose tiny/list/default thumbnails so the browser can pick
+    an appropriately sized image via ``srcset``. External avatars fall back to a
+    single URL.
+    """
+    if user.avatar and user.avatar != 'False':
+        tiny = user.get_avatar_url(event=event, thumbnail='tiny')
+        default = user.get_avatar_url(event=event, thumbnail='default')
+        return default or tiny
```
SxxAq: calls get_avatar_url() runs twice for both tiny and default, triggering extra disk I/O. 
should fetch default first and only fall back to tiny if default is missing.

SxxAq (changes requested):

Saksham-Sirohi (changes requested):
## Request changes

Nice direction overall — dropping the full `schedule_json` embed and loading speakers via a paginated `?format=json` API is the right fix for large events.

Please address these before merge:

### 1. Featured sort is broken without schedule data
`featuredSortAvailable` still depends on `this.scheduleData?.schedule…`, which is no longer present on `/speakers/`. That always evaluates false, so `mounted` forces `sortBy = 'a-z'` and the backend’s featured-first default never runs.

Derive featured availability from the API (or omit `sort` by default so backend featured order applies).

### 2. `build_speaker_cards` still loads the full talk set on every page
Each JSON page request still pulls essentially all schedule talks, so work still scales with total schedule size. Filter talks to the speaker user IDs on the current page (or query sessions from those users) so cost tracks page size, not event size.

### 3. Harden `fetchSpeakers`
- Check `res.ok` before parsing JSON and surface `loadError` on failure.
- Guard against search races (AbortController or ignore stale responses) so an older request can’t overwrite a newer search.

Happy to re-review once these are in.

Saksham-Sirohi:
### Additional requests (should-fix)

On top of the earlier must-fix items, please also address:

1. **Dead Vue imports/helpers in `SpeakersList.vue`**  
   After the rewrite these appear unused and can be removed:
   - imports: `compareFeaturedSpeakers`, `sessionsForSpeaker`
   - local helpers: `normalizeLocaleCode`, `localePrimary`, `localesMatch`
   - (and `formatLanguageLabel` if nothing calls it anymore)

2. **Shareable search URL**  
   `clearAllFilters` / `searchQuery` currently only drive the API fetch via the watcher. Please sync `q` to the browser URL (`history.replaceState` / read `?q=` on mount) so `/speakers/?q=smith` is shareable and reload-safe, as called for in #4776.

3. **Return a usable `next` from the JSON API**  
   Backend currently returns `'next': bool`, and the client invents the next page URL by incrementing `page`. Prefer returning a full next URL (or at least the next page number) from the backend so pagination stays correct when other query params (`q`, `sort`) are present and we don’t rely on client-side page arithmetic.

SxxAq (approved):
LGTM

Saksham-Sirohi (changes requested):
## Request changes

Thanks for addressing the earlier must-fixes (featured sort, page-scoped talks query, `res.ok` + AbortController, usable `next` URL, `?q=` sync, dead imports). Those look good.

Please address these before merge — they are regressions / incomplete performance wins in the current tip:

### 1. `SpeakersList` ignores the `speakers` prop
`filteredSpeakers` always returns `speakersFromApi`, and `mounted` always fetches `?format=json` from `window.location`. That breaks shared usages that pass speakers in (notably video landing: `speakers-list(:speakers="featuredSpeakers", …)`), which will show empty / hit the wrong URL.

Keep a prop/schedule-data path for embed usages; only use the paginated API on the public `/speakers/` overview (or when no `speakers` prop / schedule speakers are available).

### 2. Full schedule still loads before the list mounts
Removing `schedule_json` from the speakers page only lightens the HTML. `App.vue` still gates `speakers-list` on `schedule`, so it falls back to `fetchWidgetScheduleData` (`widgets/schedule.json` / full widget payload) **before** `SpeakersList` can mount and call the paginated API.

So large events still download the full schedule client-side, plus an extra API round-trip — which undercuts the PR’s performance goal. Speakers overview should be able to render from the paginated API without waiting on the full widget schedule.

### 3. Empty state always shows
```pug
.empty(v-if="loadError")
.empty(v-else)
```
The `v-else` branch is not gated on `!filteredSpeakers.length`, so “No speakers found” still renders under a successful list. Gate it on empty results (and ideally not while the first load is in progress).

### 4. Details view is hollow with the new card payload
API cards only include `{code, name, is_featured, avatar, sessions:[{id,title}]}`, but details mode still expects biography, social links, and session start/end/track styling. Those fields will be blank.

Either enrich the JSON card payload for details mode, or temporarily default/force list mode on the API-driven speakers page until that data is available.

Happy to re-review once these are in.

Rachit7168 (approved):
LGTM 

https://github.com/user-attachments/assets/06ad7866-36a2-44c9-a43d-5dfd76b3a975

ViRUS-0-0 (approved):

Saksham-Sirohi (changes requested):
## Request changes

Direction is still right (drop full `schedule_json` on `/speakers/`, paginated `?format=json`), but this is **not** yet a no-regression swap. A few session/schedule data gaps and UI regressions remain on the current tip.

### 1. Session times go blank (timezone missing)

On the speakers overview, `App.vue` stubs:

```js
this.schedule = { talks: [], rooms: [], schedule_unavailable: false }
```

That stub has **no `timezone`**, so:

```js
this.currentTimezone = … : this.schedule.timezone  // → undefined
```

`formatSessionSlot` / `formatSessionDateTime` then early-return when `!tz`, so details-view session times render empty.

**Fix:** include `timezone` (and preferably `has_ampm`) in `speakers_meta_json`, and set `currentTimezone` from that meta when the schedule stub is used.

### 2. Session payload is incomplete vs the old schedule sessions

`build_speaker_cards()` currently returns sessions with only:

`id`, `title`, `start`, `end`, `track: { color }`

Missing fields the previous schedule-backed UI could use / that details view expects for parity:

- `room` (name/id)
- `content_locale`
- `track.id` / `track.name` (not only color)

Please restore enough session fields that the speakers details view stays functionally equivalent.

### 3. `clearAllFilters` is broken

`SpeakersList.vue` defines `clearAllFilters` (and `toggleDropdown`) twice. The later definition only clears `searchQuery` / closes the dropdown and **does not** clear `selectedLanguages` / `selectedTracks`.

Remove the duplicate and keep one implementation that resets search + filters.

### 4. Language labels regressed

`formatLanguageLabel` now returns raw locale codes instead of `Intl.DisplayNames` labels. Please restore human-readable language names.

### 5. Confirm embed / prop path still intact

Video landing still does `speakers-list(:speakers="featuredSpeakers", …)`. Prop path looks partially restored — please re-verify it does **not** hit `/speakers/?format=json` or show empty when `speakers` is passed.

---

Happy to re-review once timezone/session fields and the filter/label regressions are fixed. Pagination itself is fine as a perf tradeoff; the blockers above are data/UX parity issues.

Saksham-Sirohi (changes requested):
JsonResponse without I18nJSONEncoder (already flagged) — title / track.name / room.name can 500 or serialize badly on multilingual events.
Biography ignores public_biography (already flagged) — always included in cards.
Search can hit private / non-schedule data — default_filters includes biography__icontains and user__submissions__title__icontains with no “visible on current schedule” constraint, so people can be found via hidden bios or non-public submission titles.
Language filter lost locale normalization — old UI used normalizeLocaleCode / fallback matching; backend now does exact content_locale__in=…. en vs en-us (etc.) will miss speakers.
event.cfp can throw in build_speaker_cards when reading social-link flags (unlike speaker_public_field_flags, which guards ObjectDoesNotExist).
Track checkbox type mismatch — meta track.id is int; URL/selectedTracks are strings → selected state / toggles can break after reload.
Sessions not sorted by start — appended in query order; details times can look shuffled vs old schedule UI.
Duplicate Vue keys — session id is submission code; multi-slot talks collide on :key="session.id".
Avatar payload incomplete — cards only set avatar (often “default” size); list UI prefers avatar_thumbnail_tiny / _default (always missing) → heavier images than the old schedule thumbnails.
Social-link prefetch likely never runs — object_list is usually a list, so prefetch_related is skipped → N+1 on social links per page.
Room is in JSON but not shown in the details session row (parity gap vs richer schedule data).
Duplicate toggleDropdown method; clearAllFilters doesn’t reset sort.
Accept: application/json sniffing beside ?format=json is a bit brittle.

Saksham-Sirohi:
`user__submissions__title__icontains` is not limited to this event. On the public speakers page, searching for a talk title from another event the same user spoke at can list that speaker, including unpublished/private titles from other events.

Rachit7168 (approved):
LGTM , Tested again

Saksham-Sirohi (approved):

PR #4831, opened 2026-08-10
fix(schedule): restore multimd-table under markdown-it 15

Saksham-Sirohi:
## Summary
- Video production builds crash with `e.utils.assign is not a function` because the router pulls schedule rich text, which loads `markdown-it-multimd-table`.
- `markdown-it` 15 removed legacy `md.utils.assign`; wrap multimd-table with a small compat shim that falls back to `Object.assign` instead of pinning markdown-it.

## Test plan
- [x] With `EVY_NPM_DEV=0`, rebuild video assets (`make npminstall` / collectstatic) and open an event video page
- [x] Confirm console no longer shows `utils.assign is not a function` from `vendor-markdown` / `router`
- [x] Confirm schedule session/talk rich text (including markdown tables) still renders

## Summary by Sourcery

Bug Fixes:
- Prevent crashes in video event pages caused by markdown-it-multimd-table calling the removed md.utils.assign function under markdown-it v15.

### DIFF
--- a/app/eventyay/webapp/schedule/src/utils/eventyayRichText.js
+++ b/app/eventyay/webapp/schedule/src/utils/eventyayRichText.js
@@ -7,6 +7,17 @@ import MarkdownIt from 'markdown-it'
 import createDOMPurify from 'dompurify'
 import multimdTable from 'markdown-it-multimd-table'
 
+/**
+ * markdown-it v15 removed legacy md.utils.assign (use Object.assign).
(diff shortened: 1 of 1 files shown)


Rachit7168 (approved):
LGTM , Working Now 

https://github.com/user-attachments/assets/bc9d4f24-2a82-4787-8b83-c53c6cc227fb

PR #4837, opened 2026-08-11
Add request throttling, custom stream throttle and 404 guard middleware

### LINKED ISSUE #4602: No API rate limiting — unauthenticated clients can poll without restriction
**Component:** `app/eventyay/config/settings.py` (REST_FRAMEWORK), DRF viewsets  
**Severity:** High — absence of throttling means a single misbehaving or unauthorized client can consume unlimited server capacity

---

## Problem

`app/eventyay/config/settings.py` lines 1335–1349:

```python
REST_FRAMEWORK = {
    'DEFAULT_PERMISSION_CLASSES': [
        'eventyay.api.auth.permission.EventPermission',
    ],
    'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.PageNumberPagination',
    'DEFAULT_VERSIONING_CLASS': 'rest_framework.versioning.NamespaceVersioning',
    'PAGE_SIZE': 50,
    'DEFAULT_AUTHENTICATION_CLASSES': (
        'eventyay.api.auth.device.DeviceTokenAuthentication',
        'rest_framework.authentication.SessionAuthentication',
        'eventyay.api.auth.token.TeamTokenAuthentication',
        'oauth2_provider.contrib.rest_framework.OAuth2Authentication',
    ),
    'DEFAULT_RENDERER_CLASSES': ('rest_framework.renderers.JSONRenderer',),
    'UNICODE_JSON': False,
}
```

**There is no `DEFAULT_THROTTLE_CLASSES` or `DEFAULT_THROTTLE_RATES`.** Any client — authenticated or not — can call any API endpoint without restriction.

The logs show dozens of 403/401/404 responses per second to `streams/current`. With no throttling:
- The server cannot tell the difference between a legitimate burst and an attack.
- A client that keeps receiving 403 but keeps polling anyway incurs full Django processing cost on every request.
- Rate limiting only exists in the presale contact form (`presale/views/contact.py`: 5 requests per 10 minutes) — not on any API endpoint.

---

### 1. Add default throttle classes to `REST_FRAMEWORK` settings

```python
# app/eventyay/config/settings.py

REST_FRAMEWORK = {
    # ... existing config ...
    'DEFAULT_THROTTLE_CLASSES': [
        'rest_framework.throttling.AnonRateThrottle',
        'rest_framework.throttling.UserRateThrottle',
    ],
    'DEFAULT_THROTTLE_RATES': {
        'anon': '60/minute',    # 1 request/second average for unauthenticated clients
        'user': '300/minute',   # 5 requests/second for authenticated users
    },
}
```

### 2. Add a dedicated throttle class for public stream/schedule endpoints

Create `app/eventyay/api/throttles.py`:

```python
from rest_framework.throttling import AnonRateThrottle, UserRateThrottle

class PublicStreamThrottle(AnonRateThrottle):
    """
    Stricter throttle for the streams/current endpoint.
    This endpoint is polled by frontends; the client should self-limit,
    but server-side throttling is the backstop.
    """
    scope = 'public_stream'

class PublicScheduleThrottle(AnonRateThrottle):
    scope = 'public_schedule'
```

Add rates to settings:

```python
'DEFAULT_THROTTLE_RATES': {
    'anon': '60/minute',
    'user': '300/minute',
    'public_stream': '10/minute',    # 1 per 6s per IP — matches 10s frontend poll
    'public_schedule': '30/minute',
},
```

Apply to the room viewset:

```python
# app/eventyay/api/views/room.py

from eventyay.api.throttles import PublicStreamThrottle

class RoomViewSet(...):
    @action(detail=True, methods=["get"], url_path="streams/current")
    def current_stream(self, request, pk=None, **kwargs):
        self.throttle_classes = [PublicStreamThrottle]
        ...
```

### 3. Return `Retry-After` on 429 responses

DRF's built-in throttle already sets `Retry-After`. Verify it is present in the response and document expected client behavior (see sub-issue 01).

### 4. Add rate limiting for excessive 404 generators

Clients generating more than 30 `404` responses per minute (scanner pattern) should be rate-limited at the application level or flagged for edge-level blocking:

```python
# In a custom middleware or DRF exception handler:
# Track 404 count per IP in Redis with a sliding window.
# After threshold, return 429 for all subsequent requests from that IP.
```

Consider also configuring Nginx or Cloudflare WAF rules to block IPs that generate more than N 4xx responses per minute.

### 5. Set `Retry-After` on temporary overload responses

When returning 503 or 429 due to application overload, include a `Retry-After` header:

```python
response = Response(
    {"detail": "Service temporarily unavailable. Try again later."},
    status=503
)
response['Retry-After'] = '30'
return response
```

---

## Acceptance criteria

- [ ] All unauthenticated API requests are throttled at no more than 60/minute per IP by default.
- [ ] `streams/current` is throttled at 10 requests/minute per IP for anonymous clients.
- [ ] 429 responses include a `Retry-After` header.
- [ ] Clients generating >30 404 responses/minute from a single IP are rate-limited.
- [ ] Throttle rates are configurable via settings without code changes.
- [ ] Throttle state is stored in Redis (not in-memory) so it works across all Gunicorn workers.
- [ ] Tests confirm that throttle is applied and that 429 is returned when limits are exceeded.

---

## Notes

- DRF `AnonRateThrottle` uses the client IP address (via `X-Forwarded-For` when `USE_X_FORWARDED_HOST = True`). Ensure `TRUSTED_PROXIES` or `SECURE_PROXY_SSL_HEADER` is correctly configured so the real client IP is used, not the Cloudflare or Nginx proxy IP.

---

## Related

- Sub-issue 01: Frontend polling — the client-side fix that reduces the number of requests reaching the server
- Sub-issue 06: Invalid URL rejection before Django
- Sub-issue 09: Protecting check-in endpoints from public traffic

Rachit7168:
Close: #4602

**Description**  

 This PR introduces robust rate‑limiting for the public API and protects the server from abusive 404 scans.  
 
 ### What’s new
 1. **DRF throttling defaults** – `DEFAULT_THROTTLE_CLASSES` and `DEFAULT_THROTTLE_RATES` are now defined in `REST_FRAMEWORK`.  
    - `anon`: 60 requests/minute (≈ 1 rps)  
    - `user`: 300 requests/minute (≈ 5 rps)  
    - `public_stream`: 10 requests/minute (the front‑end poll endpoint)  
   - `public_schedule`: 30 requests/minute  
    - `excessive_404`: 1 request/minute (used by custom middleware)  
 
 2. **Custom throttles** (`app/eventyay/api/throttles.py`)  
    - `PublicStreamThrottle` – applied to `GET /rooms/{id}/streams/current/`.  
    - `PublicScheduleThrottle` – placeholder for schedule‑related public endpoints.  
    - `Excessive404Throttle` – used by the 404‑guard middleware.  
 
 3. **ViewSet update** (`app/eventyay/api/views/room.py`) – the `current_stream` action now sets `self.throttle_classes = [PublicStreamThrottle]`.  
 
 4. **404‑rate‑limit middleware** (`app/eventyay/middleware/block_404.py`)  
    - Counts 404 responses per client IP using the Redis cache.  
    - After >30 404s in a minute the client receives a **429 Too Many Requests** with a `Retry‑After` header.  
    - Integrated into the middleware stack (`eventyay.middleware.block_404.Block404Middleware`).  
 
 5. **Redis cache already configured** – the project uses `django.core.cache.backends.redis.RedisCache`; the new middleware re‑uses the same cache, so throttling state is shared across all Gunicorn workers.  
 
 ### Why this matters
 * Prevents a single misbehaving or malicious client from exhausting server resources.  
 * Provides a graceful back‑off mechanism (`Retry‑After`) for both legit polling clients and abusive scanners.  
 * All limits are configurable from `settings.py`, no code changes needed to adjust rates.  
 
### Verification steps
1. Run the test suite – all existing tests pass.  
2. Manually test the anonymous stream endpoint (see below) – the 11‑th request gets a 429 with `Retry‑After`.  
3. Test authenticated users – after ~300 requests a 429 appears.  
4. Hit a non‑existent endpoint repeatedly – after the 30‑th 404 a 429 with `Retry‑After` is returned.

## Summary by Sourcery

Introduce API-wide rate limiting and 404 abuse protection using DRF throttling and custom middleware.

New Features:
- Add default DRF throttle classes and rates for anonymous, authenticated, and specific public API scopes.
- Introduce custom throttles for public stream, public schedule, and excessive 404 traffic.
- Apply a stricter throttle to the public current stream endpoint.
- Add middleware to detect excessive 404 responses per client IP and return 429 with Retry-After when limits are exceeded.


app/eventyay/api/views/room.py
```diff
@@ -114,6 +115,8 @@ def export_broadcast_configuration(self, request, **kwargs):
     )
     @action(detail=True, methods=["get"], url_path="streams/current")
     def current_stream(self, request, pk=None, **kwargs):
+        # Apply stricter throttling for public stream polling
+        self.throttle_classes = [PublicStreamThrottle]
```
sarafarajnasardi: This does not actually apply the stricter throttle to this action. DRF runs `check_throttles()` during `initial()`, before `current_stream()` is called, so setting `self.throttle_classes` here is too late for the request being handled. Please move this to the action declaration, 
e.g. `@action(..., throttle_classes[PublicStreamThrottle])`, or override `get_throttles()` based on `self.action`.

app/eventyay/middleware/block_404.py
```diff
@@ -0,0 +1,49 @@
...
+        if response.status_code != 404:
+            return response
+
+        ip = self._get_client_ip(request)
+        cache = caches[self.CACHE_ALIAS]
+        key = f'404_counter:{ip}'
+        count = cache.get(key, 0) + 1
+        # Store the count for 60 seconds (sliding window).
+        cache.set(key, count, timeout=60)
+
```
sarafarajnasardi: The 404 counter is updated with a non-atomic `get()` + `set()`. Under concurrent 404 traffic, multiple requests can read the same old value and overwrite each other, so the limit can be bypassed or delayed exactly when this middleware is needed. Since the default cache is Redis, please use an atomic pattern such as `cache.add(key, 0, timeout=60)` followed by `cache.incr(key)`, preserving the TTL.

sarafarajnasardi (changes requested):

sarafarajnasardi (approved):

app/eventyay/middleware/block_404.py
```diff
@@ -0,0 +1,50 @@
...
+        # Atomically increment counter, setting initial value if it doesn't exist
+        cache.add(key, 0, timeout=60)
+        count = cache.incr(key)
+
+        if count > self.MAX_404_PER_MINUTE:
+            # Apply the custom throttling class to produce a 429 with Retry-After.
+            throttle = Excessive404Throttle()
+            if not throttle.allow_request(request, view=None):
+                # throttle.wait() returns the remaining wait time in seconds.
+                retry_after = int(throttle.wait())
```
Aqil-Ahmad: DRF explicitly defines wait() as potentially returning None; it is only guaranteed to return a number when a wait period can be determined.

app/eventyay/middleware/block_404.py
```diff
@@ -0,0 +1,50 @@
...
+                return HttpResponse(
+                    content='Too many 404 responses – request throttled.',
+                    status=429,
+                    headers={'Retry-After': str(retry_after)}
+                )
+        return response
+
+    def _get_client_ip(self, request):
+        # Respect X-Forwarded-For when behind a proxy.
+        xff = request.META.get('HTTP_X_FORWARDED_FOR')
```
Aqil-Ahmad: this can have security issues. please confirm and fix if needed

Aqil-Ahmad (changes requested):

app/eventyay/middleware/block_404.py
```diff
@@ -0,0 +1,46 @@
...
+from django.utils.deprecation import MiddlewareMixin
+from django.core.cache import caches
+from django.http import HttpResponse
+from eventyay.api.throttles import Excessive404Throttle
+from eventyay.helpers.http import get_client_ip
+
+logger = logging.getLogger(__name__)
+
+
+class Block404Middleware(MiddlewareMixin):
```
ViRUS-0-0: [`Block404Middleware`] only implements `process_response`. When a client IP exceeds `MAX_404_PER_MINUTE`, subsequent requests from that IP still execute full URL routing, database queries, template rendering, and view logic before `process_response` replaces the 404 status with a 429 response. This leaves the server vulnerable to resource exhaustion / DoS attacks from automated scanners, as the heavy backend processing still runs for every request.

**Resolution:** Intercept incoming requests in `process_request` (or before `get_response` in modern middleware style) for IPs that are already flagged as throttled, returning HTTP 429 immediately without executing any view or database operations.

app/eventyay/api/throttles.py
```diff
@@ -0,0 +1,26 @@
...
+    scope = 'public_schedule'
+
+
+class Excessive404Throttle(AnonRateThrottle):
+    """
+    Generic throttle used by the ``Block404Middleware`` to
+    limit clients that generate a large number of 404 responses in a short
+    period.  The ``excessive_404`` scope is defined in ``DEFAULT_THROTTLE_RATES``.
+    """
+    scope = 'excessive_404'
```
ViRUS-0-0: [`Excessive404Throttle`] inherits from `AnonRateThrottle`. In DRF, `AnonRateThrottle.get_cache_key()` returns `None` for authenticated users. 

When `get_cache_key()` returns `None`, DRF's `allow_request()` unconditionally returns `True`. As a result, authenticated users generating excess 404s will never be throttled by `if not throttle.allow_request(request, view=None)`.

**Resolution:** Inherit [`Excessive404Throttle`] from `SimpleRateThrottle` so that IP-based throttling applies consistently regardless of whether the user is authenticated or anonymous.

app/eventyay/api/throttles.py
```diff
@@ -0,0 +1,26 @@
...
+    back‑stop against misbehaving or malicious clients.
+    """
+    scope = 'public_stream'
+
+
+class PublicScheduleThrottle(AnonRateThrottle):
+    """
+    Throttle for schedule‑related public endpoints (e.g. ``/schedule``).
+    """
+    scope = 'public_schedule'
```
ViRUS-0-0: `PublicScheduleThrottle` is defined with scope `'public_schedule'` and assigned `'30/minute'` in `REST_FRAMEWORK['DEFAULT_THROTTLE_RATES']`. However, it is never attached to `ScheduleViewSet`, `TalkSlotViewSet`, or any schedule view in [`schedule.py`].

**Resolution:** Apply `PublicScheduleThrottle` to `ScheduleViewSet` and `TalkSlotViewSet` in `app/eventyay/api/views/schedule.py`.

ViRUS-0-0 (changes requested):
Hi Rachit, I've left some comments. Please take a look at them and also add the tests


There are zero unit or integration tests for `Block404Middleware`, `PublicStreamThrottle`, `PublicScheduleThrottle`, or `Excessive404Throttle`.

**Resolution:** Add a dedicated test module (e.g. `app/tests/api/test_throttling.py`) covering:
1. `Block404Middleware` triggering HTTP 429 after 30 404 requests and asserting `Retry-After` headers.
2. Rate-limit enforcement on `GET /api/v1/rel/events/{event}/rooms/{id}/streams/current`.
3. Anonymous vs authenticated throttling behavior.

Saksham-Sirohi (changes requested):
## Review

### High — Team/Device tokens throttled as anonymous
`TeamTokenAuthentication` / `DeviceTokenAuthentication` return `AnonymousUser()`, so they hit the global **anon 60/min per IP** bucket, not user limits. Real organizer/device integrations can break under shared NAT.

**Fix:** Key throttles on `TeamAPIToken` / `Device` (or a synthetic authenticated principal), not IP-as-anon.

### High — 404 guard off-by-one / dual counters
After `count > 30`, middleware still calls `Excessive404Throttle` at `1/minute`, so the first over-threshold response often stays **404**; 429 comes later. Cache counter and DRF history disagree.

**Fix:** If `count > MAX_404_PER_MINUTE`, return 429 immediately with `Retry-After`. Drop the nested 1/min throttle.

### Medium — `streams/current` drops user throttle
`throttle_classes=[PublicStreamThrottle]` replaces defaults, so session users get no user rate limit on a polled endpoint.

**Fix:** Include `UserRateThrottle` (and token-aware throttle) alongside `PublicStreamThrottle`.

### Medium — Tests not in CI path
`tests/api/test_throttling.py` is outside `tests/stable` (what Video/Backend CI runs). Also needs a real cache (not `DummyCache`) and assertions matching the fixed 429 behavior.

### Medium — `DummyCache` + `cache.incr`
`cache.add` + `cache.incr` raises under DummyCache → 404 handling can 500 in that settings path. Fail open on cache errors.

Please address the high items before merge.

Saksham-Sirohi (changes requested):
## Review ([PR #4837](https://github.com/fossasia/eventyay/pull/4837))

Token-aware anon/user throttles and tests are an improvement. A few security and correctness issues remain.

### High — Public read on stream endpoints

`RoomViewSet.initial()` sets `allow_public_read = True` for `current_stream` and `next_stream`. That is a permission change, not just throttling. `StreamScheduleSerializer` includes `url` and `config` (HLS/stream keys). Unauthenticated clients can poll private stream URLs.

Keep auth as it was (or gate on a real public-event flag). Tests should use a token/session, not open the endpoint to the internet.

`next_stream` is made public but has **no** `throttle_classes`; only `current_stream` is limited to 10/min.

### High — 404 guard blocks the whole site per IP

`Block404Middleware` sits first in `MIDDLEWARE`. After 30 **any** 404s in a minute (static, HTML, API, crawlers), **every** later request from that IP is 429, including valid pages.

That is too broad for shared NAT / offices / misconfigured frontends, and it is a DoS lever if `TRUST_X_FORWARDED_FOR` is on (`get_client_ip` uses the first `X-Forwarded-For` hop).

Prefer: API-only (or scan-like paths), do not 429 non-404 traffic, and/or key authenticated users by user id instead of IP.

### Medium — Authenticated stream polling is barely limited

`PublicStreamThrottle` subclasses `EventyayAnonRateThrottle`, so logged-in users skip the 10/min cap and only hit `EventyayUserRateThrottle` at **300/min**. The authenticated test even asserts 15 requests are never 429.

If the frontend poll is the abuse case, apply the same `public_stream` scope to authenticated users (a `SimpleRateThrottle` keyed on user or IP).

### Medium — Schedule views replace default throttles

`ScheduleViewSet` / `TalkSlotViewSet` set `throttle_classes = [PublicScheduleThrottle, UserRateThrottle]`. That **replaces** `DEFAULT_THROTTLE_CLASSES`, so they do not use `EventyayUserRateThrottle`. Team/Device tokens fall through to DRF `UserRateThrottle` keyed by **IP**. Use `EventyayUserRateThrottle` (or append, don’t replace).

### Medium — Unused / dead config

`excessive_404` in `DEFAULT_THROTTLE_RATES` and leftover `SimpleRateThrottle` import are unused after dropping the nested 404 throttle.

### Medium — Cache increment and exceptions

`cache.incr` + `except ValueError: cache.set(..., 1)` races under concurrent 404s. Broad `except Exception` around `cache.get` / increment violates project exception rules and can hide real failures. Catch `ValueError` / cache backend errors only.

### Low

- `PublicScheduleThrottle` is unused except as a class on those viewsets — fine, but document that 30/min is the anon schedule cap.
- LocMem override in tests is good; add a test that `next_stream` is throttled if it stays public (it should not).

**Looks good:** `EventyayAnonRateThrottle` skipping `request.auth`, 429 + `Retry-After` on the 404 path, fail-open on increment errors, LocMem isolation in tests.

Saksham-Sirohi (changes requested):
## Re-review

Thanks — the last commits address the previous high items:

- `allow_public_read` on stream endpoints is gone
- `next_stream` has the same throttle classes as `current_stream`
- 404 guard is `/api/`-only
- Schedule viewsets use `EventyayUserRateThrottle`
- unused `excessive_404` / `SimpleRateThrottle` removed
- tests in `tests/stable` cover path scoping and identity keys

Two leftovers still matter:

### Medium — User/token 404 keys never apply in production

`Block404Middleware` is still first in `_LIBRARY_MIDDLEWARES`, **before** `AuthenticationMiddleware`. At that point `request.user` is unset (hence the `getattr` fallback), so every request is keyed as `404_counter:ip:…`. Authenticated NAT isolation in `_cache_key` only works in unit tests where `request.user` is assigned by hand.

**Fix:** Place `Block404Middleware` **after** `django.contrib.auth.middleware.AuthenticationMiddleware` (and session middleware). Keep the `getattr` fallback for safety.

### Low — Dead `RoomViewSet.initial`

```python
def initial(self, request, *args, **kwargs):
    super().initial(request, *args, **kwargs)
```

This does nothing now that public-read was reverted. Please delete it.

### Low — `except Exception` on increment

`cache.get` is narrowed to `ValueError`/`TypeError`; `_increment` still uses `except Exception`. Catch `ValueError` (missing key) plus the cache backend’s error type, or keep fail-open but don’t swallow everything.

The `incr` + `set(1)` race is acceptable for this guard.

After the middleware is moved past auth, this looks mergeable.

Saksham-Sirohi (approved):
Nit (non-blocking): `_cache_key`’s docstring still talks about running before auth. Team/Device tokens are still IP-keyed here because DRF sets `request.auth` in the view, not in Django middleware — that’s expected. The outer `except Exception` around `_increment()` is a reasonable fail-open if the cache backend errors.

ArnavBallinCode (approved):
LGTM

PR #4647, opened 2026-07-31
Re-introduce Tiptap editor with sanitization fixes

Rachit7168:
This PR re-introduces the Tiptap editor from #4147 which was reverted in #4520. It also includes the HTML sanitization and sourcemap fixes from #4511.

https://github.com/user-attachments/assets/22f1e52b-c579-41ba-ae84-cf52a6583b9f

## Summary by Sourcery

Reintroduce a JS-based Tiptap editor for rich text and email bodies, backed by new server-side HTML sanitization and preview mechanisms, and integrate it into existing email workflows and build infrastructure.

New Features:
- Add Tiptap-based rich text and email body editor widgets, including placeholder insertion and preview support.
- Introduce server-side HTML sanitization profiles for rich text and email editor output.

Bug Fixes:
- Ensure email HTML sanitization preserves mailto/tel links and avoids injecting rel attributes that can confuse email clients.
- Fix rich text link rendering to avoid altering existing link labels and adjust absolute link behaviour.

Enhancements:
- Refine snippet rendering to unwrap single outer paragraph tags for cleaner inline display.
- Update email compilation logic to distinguish between legacy Markdown/plain text and already-HTML Tiptap content.

Build:
- Add Vite-based build pipeline for the new editor webapp and integrate it into the Makefile.
- Register the editor JS/CSS bundles with the Django templates via a lightweight loader script.

Tests:
- Add tests for rich text linkification behaviour and email body compilation logic.
- Add tests covering HTML sanitizers for rich text and email profiles, including edge cases for schemes and attributes.
- Add tests for i18n email body form field sanitization behaviour.

Chores:
- Wire up new email editor fields in sendmail and control order forms, including placeholder lists and preview endpoints.
- Introduce a server-side email preview AJAX endpoint used by the editor toolbar.

Issue #3684, opened 2026-05-28
Implement a unified Tiptap-based editor system for eventyay


mariobehling:
eventyay currently uses different editor implementations in different areas:

* Talk submission forms use Toast UI Editor through `MarkdownWidget`.
* The video component uses Quill through a custom Vue rich text editor.
* Ticket forms and email creation areas still need simple rich text editing similar to Gmail.

As part of the current redesign work, we should implement a unified editor system instead of expanding the existing Quill or Toast UI usage into more areas.

The goal is to introduce **Tiptap** as the future shared editor framework for eventyay and provide reusable editor profiles for different use cases.

## Goals

* Provide a shared editor implementation for eventyay.
* Support simple rich text editing in ticket-related forms.
* Support Gmail-like email body editing in Tickets and Talk.
* Support placeholder and variable insertion for email templates.
* Keep the current Toast UI Markdown editor for Talk submission forms for now.
* Avoid adding another unrelated editor stack.
* Prepare a long-term path to migrate existing Quill and possibly Toast UI usage later.

## Current situation

* Talk submission fields such as abstract, description and notes use `MarkdownWidget`, which enhances textareas with Toast UI Editor.
* Video rich text editing currently uses Quill.
* Tickets and email creation areas need a simple rich text editor with a limited toolbar.
* Email editing should not use Markdown as the canonical format, because email output needs reliable HTML.

## Proposed approach

Introduce a shared Tiptap-based editor layer with reusable editor profiles.

Suggested structure:

```text
eventyay/common/editors/
  core/
  profiles/
  widgets/
  serializers/
  sanitizers/
```

or, depending on the existing frontend structure:

```text
eventyay/static/common/js/editor/
  tiptap/
  profiles/
  widgets/
```

The implementation should be reusable from Tickets, Talk and other components.

### 1. Simple rich text editor

Use for ticket forms and simple content fields.

Features:

* Bold
* Italic
* Underline
* Bullet list
* Numbered list
* Link
* Clear formatting
* Undo / redo

Storage:

* Sanitized HTML

### 2. Email editor

Use for email creation and email template areas in Tickets and Talk.

Features:

* Bold
* Italic
* Underline
* Bullet list
* Numbered list
* Link
* Clear formatting
* Undo / redo
* Insert placeholder / variable
* Preview rendered email
* Send test email, where applicable

Storage:

* Sanitized email-compatible HTML

The email editor should support placeholders such as:

```text
{{ attendee_name }}
[...]

PR #4858, opened 2026-08-12
feat(video): replace Quill with Tiptap rich text editor

Rachit7168:
Part 2 : #3684

## Summary

This PR replaces the Quill editor with Tiptap in the video webapp, consistent with the Tiptap adoption in #4647 (tickets/email). No new webapp or Dockerfile changes.

## Changes

- **`RichTextEditor.vue`** — Rewrite using `@tiptap/core` + `StarterKit` + `Image` + `Link` + `Underline` + class-based text alignment. Same toolbar UI; `v-model` accepts HTML (and legacy Quill Delta for loading) and emits HTML.
- **`RichTextContent.vue`** — Sanitized HTML via DOMPurify with an explicit allowlist. Legacy Quill Delta falls back to escaped plain text.
- **`ChatInput.vue`** — Migrated from Quill to Tiptap while preserving emoji insertion, `@` mentions/autocomplete, file attachments, and Enter-to-send. Chat still serializes mentions as `@<id>` for `ChatContent`.
- **`LandingPage.vue`** — Detects Tiptap HTML strings and routes them through `RichTextContent` (Markdown strings still use `MarkdownContent`).
- **`main.js`** — Removed Quill CSS imports.

## Affected Areas

| Location | Used for |
|---|---|
| Exhibitor manager | Long description (`text_content`) |
| Poster manager | Abstract (`abstract`) |
| Admin → Rooms → Static page | Page content |
| Admin → Rooms → Landing page | Main content |
| Chat composer | Message body with emoji/mentions/files |

## Intentional feature drops vs Quill

Video embed, full-width blocks, and H4 are **not** restored in the Tiptap toolbar. That is intentional for this migration; legacy Delta video/image embeds render as empty in the plain-text fallback. Organizers can still upload images via the image button.

## Out of Scope

No backend model or migration changes. Existing JSON fields transparently store HTML strings going forward.

## Test plan

- [ ] Landing page: edit HTML content, save, confirm alignment classes and links render after reload
- [ ] Legacy Quill Delta landing content still displays as plain text (escaped)
- [ ] Chat: send text, emoji, mention, attachment; Shift+Enter keeps line breaks; empty Enter does not send
- [ ] Image upload failure: selecting the same file again retries cleanly


app/eventyay/webapp/video/src/components/RichTextContent.vue
```diff
@@ -1,49 +1,110 @@
...
+						.join('')
+					return DOMPurify.sanitize('<p>' + plain.replace(/\n/g, '<br>') + '</p>')
+				}
+				return ''
+			}
+
+			// Regular HTML string
+			return DOMPurify.sanitize(this.content, {
+				ADD_ATTR: ['target'],
+			})
```
Saksham-Sirohi, line 53: Tiptap `TextAlign` writes `style="text-align: center|right|left"` onto `<p>` / headings. DOMPurify’s default allowlist **strips `style`**, so alignment works in the editor and disappears on the landing page.

Do not add a blanket `style` allow — that is an XSS footgun. Two safer options:

1. Configure `TextAlign` to emit classes (`text-center` / `text-right` / `text-left`). This file already styles those classes.
2. If you must keep inline styles, allow only `text-align` via `uponSanitizeAttribute` (same pattern as `eventyayRichText.js`).

Also add `rel` to the allowlist so `noopener noreferrer` survives sanitize (`ADD_ATTR: ['target', 'rel']` is not enough by itself if you later set `ALLOWED_ATTR`).

app/eventyay/webapp/video/src/components/RichTextContent.vue
```diff
@@ -1,49 +1,110 @@
...
-			// don't redirect on right click
+			// Don't intercept right-click
 			if (event.button !== undefined && event.button !== 0) return
-			// don't handle same page links/anchors or external links
+			// Don't intercept external or same-page links
 			const url = new URL(a.href)
 			if (window.location.pathname === url.pathname) return
 			if (window.location.hostname !== url.hostname) return
 			event.preventDefault()
 			router.push(url.pathname + url.hash)
```
Saksham-Sirohi: `new URL(a.href)` throws on an empty/invalid href and will break the whole click handler. Wrap it.

`router.push(url.pathname + url.hash)` drops `url.search`, so a same-host link like `/control/event/?foo=1` loses the query. Use `url.pathname + url.search + url.hash`.

app/eventyay/webapp/video/src/components/RichTextEditor.vue
```diff
@@ -37,121 +41,155 @@ bunt-input-outline-container.c-rich-text-editor(ref="outline", :label="label")
...
-				}
-			}
+	editorInstance = markRaw(new Editor({
+		element: editorMount.value,
+		extensions: [
+			StarterKit,
+			Image.configure({ inline: false, allowBase64: false }),
+			Link.configure({ openOnClick: false, autolink: true }),
+			Underline,
+			TextAlign.configure({ types: ['heading', 'paragraph'] }),
```
Saksham-Sirohi: `Link.configure` should set `HTMLAttributes: { rel: 'noopener noreferrer', target: '_blank' }` (and keep `openOnClick: false`). `setLink({ href, target: '_blank' })` currently omits `rel`, which is tabnabbing.

`TextAlign.configure({ types: ['heading', 'paragraph'] })` currently serializes alignment as a `style` attribute. Pair this with class-based alignment (see `RichTextContent.vue`) so it survives sanitization.

Please also confirm dropping `ql-video`, full-width, and H4 is intentional — those were on the Quill toolbar and are used on landing-page copy.

app/eventyay/webapp/video/src/components/ChatInput.vue
```diff
@@ -60,135 +116,238 @@ function getMentionMatch(text) {
...
+			if (!this.autocomplete?.fromPos) return null
+			try {
+				const coords = this.editor.view.coordsAtPos(this.autocomplete.fromPos)
+				const editorRect = this.$refs.editorRef.getBoundingClientRect()
+				return {
+					left: coords.left - Math.max(0, 240 - (editorRect.width + 60 - (coords.left - editorRect.x))) + 'px',
+					bottom: window.innerHeight - coords.top + 8 + 'px',
+				}
+			} catch {
+				return null
```
Saksham-Sirohi, line 155: Empty `catch` is disallowed (`.github/instructions/js.instructions.md`). `coordsAtPos` can fail when the mention pos is stale; log it:

```js
} catch (error) {
  console.warn('ChatInput: mention dropdown position failed', error)
  return null
}
```

app/eventyay/webapp/video/src/components/ChatInput.vue
```diff
@@ -60,135 +116,238 @@ function getMentionMatch(text) {
...
+			element: this.$refs.editorRef,
+			extensions: [
+				Document,
+				Paragraph,
+				Text,
+				History,
+				EmojiNode,
+				MentionNode,
+				ChatKeymap,
+			],
```
Saksham-Sirohi, line 224: Old Quill set `placeholder: this.$t('ChatInput:input:placeholder')`. The new CSS still looks for `data-placeholder` / `.is-editor-empty`, but there is no `@tiptap/extension-placeholder`, so the input is blank until the user types.

Add the Placeholder extension (and keep the existing i18n key) so chat does not regress.

app/eventyay/base/middleware.py
```diff
@@ -345,6 +345,7 @@ def process_response(self, request, resp):
                 '{static}',
                 'data:',
                 'https://fonts.gstatic.com',  # fix Google Fonts
+                'data:',
```
Saksham-Sirohi: `'data:'` is already on `font-src` two lines above. This duplicate does not change CSP. Revert this file unless there is a real missing source (if Tiptap emoji/fonts need something, it is more likely `img-src`, which already allows `data:`).

app/eventyay/webapp/video/package.json
```diff
@@ -17,6 +17,17 @@
...
+    "@tiptap/extension-document": "^2.27.2",
+    "@tiptap/extension-history": "^2.27.2",
+    "@tiptap/extension-image": "^2.27.2",
+    "@tiptap/extension-link": "^2.27.2",
+    "@tiptap/extension-paragraph": "^2.27.2",
+    "@tiptap/extension-text": "^2.27.2",
+    "@tiptap/extension-text-align": "^2.27.2",
+    "@tiptap/extension-underline": "^2.27.2",
+    "@tiptap/starter-kit": "^2.27.2",
+    "@tiptap/vue-3": "^2.27.2",
```
Saksham-Sirohi: `@tiptap/vue-3` is unused — both editors construct `Editor` from `@tiptap/core`. Drop it (and the lockfile entry) unless you plan to switch to `EditorContent`.

Saksham-Sirohi (changes requested):
### Must fix
1. **Text alignment is saved then stripped on render.** Tiptap `TextAlign` emits `style="text-align: …"`. `RichTextContent` sanitizes with default DOMPurify, which drops `style`. Do **not** globally allow `style` (XSS). Emit alignment classes instead — this file already has `.text-center` / `.text-right` / `.text-left`.
2. **Feature drop vs old editor.** The Quill toolbar had video embed, full-width, and H4. Those are gone and old Delta video/image embeds become empty in the plain-text fallback. Please restore video embed **or** explicitly confirm organizers no longer need it on landing pages.
3. **CSP change is a no-op.** `font-src` already had `'data:'`; this PR adds a second copy. Drop the duplicate.

### Should fix
- Chat placeholder is gone (old Quill had `placeholder:`). Add `@tiptap/extension-placeholder`.
- Empty `catch` in `autocompleteCoordinates` (project JS rule: log or rethrow).
- `target=_blank` links need `rel="noopener noreferrer"`.
- Internal click handler drops `url.search` and can throw on a bad `href`.
- `@tiptap/vue-3` is unused (you construct `@tiptap/core` `Editor` directly).

### Already looking good
- Chat still serializes mentions as `@<id>` so `ChatContent` keeps working.
- Image upload disables base64 (`allowBase64: false`).
- Enter-to-send / mention autocomplete / emoji node are a reasonable port of the old Quill bindings.
- Video Vite CI passing means the earlier duplicate-Tiptap import issue is resolved.

MukundC25 (commented):
@Rachit7168 resolve conflicts

Saksham-Sirohi (changes requested):
### Must fix

1. **Toolbar active state does not track selection.** `editorState` is written in `onSelectionUpdate` / `onUpdate` but never read. `isActive()` only looks at a non-reactive `editorInstance`, so Bold/Italic/heading buttons will not update when the caret moves unless the HTML also changes. Read `editorState.value` at the start of `isActive()` (or drive the buttons off that ref).

2. **Chat line breaks can be dropped.** `serializeContent()` concatenates text/emoji/mention nodes with no `\n` between paragraphs. Paste or Shift+Enter becomes `helloworld`. Quill ops included newlines. Append `\n` when leaving a block node.

### Please confirm

Video embed, full-width, and H4 are still gone, and legacy Delta video/image embeds still render as empty in the plain-text fallback. That is fine if it is intentional — please say so on the thread so we do not treat it as a regression.

### Nits

- `CustomTextAlign` is declared between `import` statements in `RichTextEditor.vue`. Move it below all imports.
- `[style*="text-align"]` CSS in `RichTextContent.vue` is unused now that alignment is class-based.

After the two must-fixes (and the video-embed note), this looks like a solid Tiptap port.

ArnavBallinCode (approved):
LGTM

app/eventyay/webapp/video/src/components/ChatInput.vue
```diff
@@ -37,21 +37,81 @@ bunt-input-outline-container.c-chat-input
...
+	atom: true,
+	addAttributes () {
+		return {
+			id: { default: null },
+			name: { default: null },
+		}
+	},
+	parseHTML () {
+		return [{ tag: 'span[data-mention]' }]
+	},
```
ViRUS-0-0, line 98: `MentionNode` defines `parseHTML` as `[{ tag: 'span[data-mention]' }]` without `getAttrs`. If HTML containing a mention is pasted or parsed by Tiptap, `id` and `name` will default to `null` instead of being restored from `data-id` and `data-name`.

Add `getAttrs` to extract `id` and `name` attributes from the matched DOM element.

app/eventyay/webapp/video/src/components/ChatInput.vue
```diff
@@ -203,126 +373,100 @@ export default {
...
-					body: text
+					body: text,
 				})
 			}
-			this.quill.setContents([{insert: '\n'}])
+			this.editor?.commands.clearContent()
 		},
-		async attachFiles(event) {
+
+		async attachFiles (event) {
```
ViRUS-0-0, line 435: `attachFiles` does not reset `event.target.value`. If an attendee uploads a file, removes it using the remove button, and then attempts to re-attach the exact same file, the file input's `change` event will not fire.

**Resolution**:
Reset `event.target.value = ''` at the end of the method (matching the pattern used in [`RichTextEditor.vue:177`](app/eventyay/webapp/video/src/components/RichTextEditor.vue:177)).

app/eventyay/webapp/video/src/components/RichTextEditor.vue
```diff
@@ -37,134 +41,208 @@ bunt-input-outline-container.c-rich-text-editor(ref="outline", :label="label", :
...
+	if (!editorInstance) return
+	const existing = editorInstance.getAttributes('link').href || ''
+	const url = prompt(i18n.t('RichTextEditor:link:prompt') || 'Enter URL', existing)
+	if (url === null) return
+	if (url === '') {
+		editorInstance.chain().focus().unsetLink().run()
+	} else {
+		editorInstance.chain().focus().setLink({ href: url, target: '_blank', rel: 'noopener noreferrer' }).run()
+	}
+}
```
ViRUS-0-0, line 157: When editing an existing link while the cursor is inside the link text (without an active range selection), `setLink` / `unsetLink` without `extendMarkRange('link')` can update only part of the link text or fail to unset the full link mark.

**Resolution**: Add `.extendMarkRange('link')` before `.unsetLink()` and `.setLink(...)`.

ViRUS-0-0 (changes requested):
Small minor issues. Please take a look.
The PR looks good to me overall

ViRUS-0-0 (approved):

Saksham-Sirohi (approved):

PR #4902, opened 2026-08-13
refactor: replace pytz with zoneinfo

### LINKED ISSUE #4786: Replace pytz with zoneinfo
For timezone, we should not need a third-party lib like pytz. Just use the `zoneinfo` module from standard lib.

Rachit7168:
## Summary
- Remove the direct `pytz` dependency and migrate timezone handling to Python's standard library (`zoneinfo` and `datetime.timezone.utc`).
- Add `eventyay.timezones` with the curated `common_timezones` list previously provided by `pytz`, plus a `localize_datetime` helper for naive datetime conversion.
- Fix remaining incomplete migration spots that caused CI failures (`timezone is not defined`, `.localize()` on `ZoneInfo`).
- Update admin documentation to refer to IANA time zone names instead of `pytz` names.

Closes #4786

## Test plan
- [ ] CI integration tests pass (previously failing `tests/stable/test_publication.py` settings views)
- [ ] Event settings save correctly with timezone changes
- [ ] Timezone dropdowns in event/user forms still populate as expected
- [ ] PDF exports and iCal generation show correct event-local times
- [ ] Relative date / DST edge-case tests pass (`tests/tickets/base/test_reldate.py`)

## Summary by Sourcery

Switch application timezone handling from pytz to Python’s standard library zoneinfo/UTC and introduce a shared timezones helper module.

Enhancements:
- Introduce eventyay.timezones module providing a curated common_timezones list and a localize_datetime helper for consistent timezone-aware datetime handling.
- Refactor core models, forms, views, services, exporters, tasks, and utilities to use zoneinfo and datetime.timezone.utc instead of pytz throughout the codebase.

Build:
- Remove the pytz dependency from the project configuration and lockfile.

Documentation:
- Adjust admin and configuration documentation references to IANA/zoneinfo time zone usage instead of pytz-specific naming.

Tests:
- Update ticketing, presale, API, and plugin test suites to work with zoneinfo-based datetimes and verify DST/relative date behavior under the new timezone implementation.


hongquan (approved):

app/eventyay/timezones.py
```diff
@@ -0,0 +1,444 @@
+common_timezones = [
```
Aqil-Ahmad, line 1: not sure if its better to keep this as a manual copied list

Aqil-Ahmad (approved):

app/eventyay/timezones.py
Rachit7168: It's a valid concern, but zoneinfo doesn't provide an equivalent curated list—zoneinfo.available_timezones() returns over 600 timezones, including many historical and obscure ones which would severely clutter our UI dropdowns. The list in pytz.common_timezones was specifically curated to only show relevant timezones to users. By hardcoding this list, we keep the UI exactly the same and prevent dropdown choices from randomly changing based on the host server's OS timezone database.

Saksham-Sirohi (approved):

PR #60, opened 2024-03-20
Changed settings.py and changed the config and system name to eventyay 

KanzaAkram:
I have gone through settings.py file located under src => pretix =>test-utils =>settings.py and changed the config and system name to eventyay.

PR #4916, opened 2026-08-13
feat(video): plugin language streams in admin SPA

### LINKED ISSUE #4915: Add interpretation plugin language streams in Video area
## Problem
 
I am adding the video area's translation logic to sit in the [interpretation plugin](https://github.com/fossasia/eventyay-interpretation) instead of core eventyay. All logic for per-room language streams are stored and served by the `eventyay-interpretation` plugin: the model, REST API, event flag, and room-config injection comes from plugin. need a PR for just the frontend to render those changes in eventyay videos. 
 
Right now the video app only knows about the core "Languages and Audio Source" list. Even when the plugin has streams configured, video admin and the attendee language dropdown won't use them.
 
We need a small change in the Eventyay video app so organizers can edit plugin streams in video admin and attendees will see them in the room, but only when the event has turned on "Use plugin language streams" in the Interpretation plugin.
 
---
 
## What we want
 
When the interpretation plugin is enabled and the event flag is on:
 
1. **Video admin (room editor)**
   - Keep the existing core language section as-is.
   - Show a second section: "Interpretation Source".
   - Load and save that list via the plugin API (not core room save).
2. **Attendee room**
   - A second Language dropdown will use the plugin stream list (from room config), not the core list.
When the flag is off (or plugin not installed):
- Everything will work exactly as today. No extra section, no behavior change.
---
 
## Why this is an Eventyay change
 
The plugin cannot modify the compiled video frontend on install. The UI for video admin and the attendee room lives in the Eventyay video SPA, so this wiring has to live here.
 
The plugin PR handles storage, API, and adding stream data to room JSON. This issue is the frontend half.
 
---
 
## Acceptance criteria
 
- [ ] With flag on + plugin installed: video admin shows both language sections; plugin entries save and reload correctly.
- [ ] With flag on: second language dropdown shows plugin streams.
- [ ] With flag off: no plugin section; dropdown uses core `languageUrls` only.
- [ ] With plugin not installed: no crash; stock video behavior unchanged.
- [ ] Video frontend rebuilt and tested (`npm run build` + `collectstatic`).
---
 
## Notes
 
- Depends on the `eventyay-interpretation` plugin for full functionality; safe to merge in Eventyay alone (graceful when plugin absent).

weshaan:
resolves #4915 
## Summary
 
Made minimal video SPA changes so [interpretation plugin](https://github.com/fossasia/eventyay-interpretation) language streams work in the video admin and the attendee room when the event flag `interpretation_use_plugin_streams` is on.
 
This PR is just the frontend: load/save via the plugin API, show a second editor in admin, and second interpretation dropdown when the flag is on. 

image

 
---
 
## Why this PR exists
 
**Companion plugin (full feature):** I am adding the video area's translation logic to sit in the plugin instead of core eventyay. This PR is the Eventyay video SPA half. All logic for per-room language streams are stored and served by the `eventyay-interpretation` plugin: the model, REST API, event flag, and room-config injection comes from plugin. This PR is just the frontend to render those changes in eventyay videos. 

**Safe to merge on its own:** Without the plugin installed, or with the event flag off, behavior is unchanged : core `languageUrls` only, no plugin section in admin, attendee dropdown unchanged. If the plugin is missing, plugin API calls fail gracefully and the UI stays hidden.

**Full feature when both are deployed:** Install/upgrade the interpretation plugin, enable "Use plugin language streams" in the video room on the Interpretation overview, rebuild video frontend (`npm run build` + `collectstatic`). Then video admin shows the second language section and the attendee dropdown uses plugin streams.

**Plugin PR:** [Duplicate Video Area translation logic In plugins - #60](https://github.com/fossasia/eventyay-interpretation/pull/60)

---
 
## How it works
 
**Event flag ON** (set on Interpretation overview in plugin):
 
1. **Video admin** : room editor shows two sections:
   - *Languages and Audio Source* : unchanged core path (`languageUrls`, normal room save)
   - *Interpretation Source* : reads/writes plugin `.../interpretation/streams/` API
2. **Attendee room** : two language dropdown, interpretation uses `interpretation_language_streams` from room config (plugin injects this via `video_integration.py`).
**Event flag OFF** : stock behavior; no plugin section, one dropdown uses core `languageUrls`.
 
image

 
---
 
## What changed (8 files)
 
| File | Purpose |
|---|---|
| `lib/interpretation-api.js` | Build plugin API URLs + auth headers |
| `lib/interpretation-language-streams.js` | GET/PATCH plugin streams for a room |
| `interpretation-streams.js` | Attendee helper: use plugin list when flag on |
| `components/LanguageAudioSourceList.vue` | Shared language/audio row UI (extracted from inline markup) |
| `views/admin/rooms/EditForm.vue` | Load plugin streams on open; save them after room save |
| `views/admin/rooms/types-edit/stage.vue` | Core + plugin language sections for stage rooms |
| `views/admin/rooms/StreamSchedule.vue` | Same for schedule-driven rooms |
| `views/rooms/item.vue` | Attendee dropdown branches on plugin flag |
 
**Note:** Core language editor refactored to `LanguageAudioSourceList` — same behavior, less duplication. Plugin section reuses that component.
 
---
 
## Test plan
 
- [ ] Install/enable interpretation plugin; turn on "Use plugin language streams" on Interpretation overview
- [ ] Rebuild video SPA + `collectstatic` before testing in Django-served environment
- [ ] Video admin → edit stage room → see both language sections when flag on
- [ ] Add plugin languages → Save → reload → entries persist (via plugin API)
- [ ] Core Languages and Audio Source still saves independently
- [ ] Attendee room → dropdown shows plugin list when flag on
- [ ] Turn flag off → plugin section hidden; dropdown uses core `languageUrls`

 ## Feature video
 
 

https://github.com/user-attachments/assets/3cd1961a-ce01-4201-82f5-662357c3b91f

## Summary by Sourcery

Integrate interpretation plugin language streams into the video admin and attendee room while reusing a shared language/audio configuration UI.

New Features:
- Expose plugin-provided interpretation language streams in stage and schedule-driven room editors when the event uses plugin streams.
- Display separate core and plugin language dropdowns in the attendee room, allowing selection of either original/core audio or plugin-based interpretation streams.

Enhancements:
- Extract the language/audio source editor into a reusable LanguageAudioSourceList component shared by core and plugin language configurations.
- Improve the audio translation dropdown layout and semantics to support multiple labeled selectors in the stage tools bar.


Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/e94f1567-d05f-42b7-83fc-e30e84876a45

ArnavBallinCode (approved):

Saksham-Sirohi (changes requested):
## Review ([PR #4916](https://github.com/fossasia/eventyay/pull/4916))

The split (core `languageUrls` vs plugin streams, hide UI when the flag is off, skip PATCH after a failed GET) is the right shape. A few bugs and product-JS issues to fix before merge.

### High — Plugin streams never save on **create**

`loadInterpretationLanguageStreams()` returns immediately when `this.creating || !this.config?.id`, and **does not** set `interpretationAdmin.loaded = true`.

`save()` only PATCHes plugin streams when `loaded && !streamsLoadFailed`. On first create that is always false, so the Interpretation source list is dropped.

After create, either:
- set `loaded = true` when there is nothing to fetch, then PATCH with the new `roomId`, or
- call `loadInterpretationLanguageStreams()` once `config.id` exists (watch `config.id`).

### High — Possible crash in `stage.vue`

```js
if (!this.modules['livestream.youtube'].config.languageUrls) {
```

The previous guard was `config && !config.languageUrls`. If `config` is missing this throws and the stage editor fails to open. Restore the null check.

### Medium — Two dropdowns, one store value

Core and Interpretation both write `youtubeTranslationsByRoom[roomId]`. Picking a language in one resets the other to “Original” (unless the YouTube IDs happen to match). That will confuse attendees.

Keep two selections, or treat “Original” on Interpretation as “use core audio” without clearing a distinct core choice. At least don’t show both as selected for the same URL.

### Medium — `LanguageAudioSourceList` mutates the `entries` prop

`addEntry` / `removeEntry` / `normalizeEntry` mutate the parent array in place, and `:key="index"` is unstable when rows are deleted. Emit `update:entries` (or mutate only nested fields) and key rows by a stable id.

### Medium — New Vue files use Options API

Project JS rules ask for Vue 3 Composition API on **new** components (`LanguageAudioSourceList.vue`). Existing Options API files can stay as they are.

### Low

- `interpretation-streams.js` lives at `src/` and is imported as `../../interpretation-streams`; put it under `lib/` like the other helpers.
- `interpretationRouting()` copies StreamSchedule’s pathname fallback (`pathParts[0]/[1]`). That is wrong on hosts where the video app is not `/{organizer}/{event}/...`. Prefer `world.organizer_slug` / `world.slug` only, or the same helper StreamSchedule already has (don’t fork it).
- Delete button in `LanguageAudioSourceList` has no accessible name (`aria-label`).
- No unit tests for `pluginLanguageStreams` / skip-save-on-load-failure. Those are cheap and would lock the wipe-protection.

**Looks good:** skip PATCH when the initial streams GET failed; flag-off path unchanged; shared list component for admin duplication.

Saksham-Sirohi (approved):
Please address the pointed issues in a follow-up issue as per the discussion with @ArnavBallinCode

PR #4942, opened 2026-08-15
feat: add user specific default organizer preselection

### LINKED ISSUE #4937: Add default organiser selection for event creation
The event creation form has a dropdown to choose the organiser of an event. Currently, this dropdown does not reliably preselect the most relevant organiser.

The form should automatically select the user’s default organiser. If a user only belongs to one organiser, that organiser should be preselected.

Image
Image

### Event creation form

* Preselect the user’s default organiser in the `Organizer` dropdown.
* If the user belongs to only one organiser, preselect that organiser automatically.
* If the user belongs to multiple organisers and no default is set yet, use the first organiser the user was added to as the default.
* The user should still be able to choose a different organiser manually when creating the event.

### Default organiser logic

* Each user should have their own default organiser setting.
* The default organiser must be user-specific.
* Setting a default organiser for one user must not affect other users of the same organiser.
* By default, the first organiser a user is added to should become that user’s default organiser.
* If the user is removed from their default organiser, the system should choose the next available organiser as default or leave the default empty if none exists.

### New organiser flow

When a user creates a new organiser:

* Provide an option to make the new organiser the user’s default organiser.
* If the user has no existing default organiser, the newly created organiser should automatically become the default.
* If the user already has a default organiser, do not override it unless the user explicitly chooses to make the new organiser the default.

### Organiser overview page

On the organiser overview page:

```text
/common/organizers/
```

Show an indication behind the organiser name if it is the user’s default organiser, for example:

```text
FOSSASIA (default)
```

### Edit organiser page

When editing an organiser:

* Add an option to make this organiser the current user’s default organiser.
* This action should only apply to the currently logged-in user.
* It must not change the default organiser for other members or admins of the organiser.

## Edge cases to cover

* User has no organiser.
* User has exactly one organiser.
* User has multiple organisers and one default.
* User has multiple organisers and no default yet.
* User creates a new organiser.
* User leaves or is removed from their default organiser.
* User’s default organiser is deleted.
* User edits an organiser and sets it as default.
* Admin mode should not accidentally change another user’s default organiser.
* API/backend permissions must prevent setting an organiser as default if the user is not a member of that organiser.

## Acceptance criteria

* Event creation form preselects the user’s default organiser.
* If the user has only one organiser, it is preselected.
* If no default exists, the first organiser the user was added to becomes the default.
* Users can set a newly created organiser as their default.
* Organiser overview shows `(default)` behind the default organiser name.
* Organiser edit page allows the current user to make that organiser their default.
* Default organiser setting is per user and does not affect other users.
* Removing or deleting a default organiser is handled safely.
* All relevant edge cases are tested.

prayag78:
## Summary

Adds user-specific default organizer settings and automatically preselects the relevant organizer in the event creation form.

## Approach

- Added `default_organizer` field and `get_default_organizer()` resolver to the `User` model, with automatic fallback to the first organizer the user joined.
- Updated `EventCreateView` and `EventWizardFoundationForm` to automatically preselect the user's default organizer (or single organizer) on the event creation page.
- Added a `set_as_default` option to the new organizer creation and organizer edit forms, persisting default organizer preferences scoped strictly to the current user.
- Updated the organizer overview pages (`/common/organizers/` and `/control/organizers/`) to display a `(default)` badge next to the user's default organizer name.
- Added `m2m_changed` signal handling for team membership additions and removals to safely auto-assign or transition default organizers when team access changes.
- Added `is_default` field to `OrganizerSerializer` and a `set-default` action on `OrganizerViewSet` with membership validation.
- Added comprehensive unit and integration test suite in `test_default_organizer.py` covering all edge cases.

https://github.com/user-attachments/assets/48ef7816-c4de-4ffa-9bcd-c09bcfaad354

Fixes #4937

## Summary by Sourcery

Add user-specific default organizer selection across organizer management, event creation, and API workflows.

New Features:
- Add user-specific default organizer preferences with automatic fallback to the first organizer the user joined.
- Preselect the applicable organizer during event creation and provide controls for setting defaults when creating or editing organizers.
- Expose default-organizer status and a membership-validated API action for managing the preference.

Enhancements:
- Keep default organizer preferences scoped to individual users and automatically clear invalid defaults when organizer membership changes.
- Display the user's default organizer in organizer overviews.

Tests:
- Add unit and integration coverage for default resolution, event preselection, organizer management, membership changes, deletion, and API behavior.


app/eventyay/base/models/auth.py
```diff
@@ -638,6 +646,44 @@ def get_organizers_with_permission(self, permission, request=None):
 
         return Organizer.objects.filter(id__in=self.teams.filter(q).values_list('organizer', flat=True))
 
+    def get_default_organizer(self, can_create_events=False):
```
Rachit7168, line 650: A "getter" method should not have side-effects. Remove all the self.save() and self.default_organizer = ... assignment logic from this method. If a user doesn't have a default organizer, simply return the fallback first_team.organizer dynamically without persisting it to the database.

 Also, instead of pulling all teams into memory with list(self.teams.values_list('organizer_id', flat=True)), you can just run a much faster query: if self.teams.filter(organizer_id=self.default_organizer_id).exists():.

app/eventyay/base/models/organizer.py
```diff
@@ -899,3 +901,40 @@ def delete(self, *args, **kwargs):
     def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
         self.organizer.cache.clear()
+
+
+@receiver(m2m_changed, sender=Team.members.through)
```
Rachit7168, line 906: If you remove the "auto-save" side-effect from get_default_organizer (as suggested in my previous comment), you don't need to manually re-assign everyone's fallback organizer in the database one by one! You can replace this entire loop with a single bulk update query:

Don't introduce an N+1 query; try to make it as efficient as possible

Rachit7168 (changes requested):
Thanks for working on this. Please address

app/eventyay/base/models/auth.py
prayag78: fixed

app/eventyay/base/models/organizer.py
prayag78: fixed

Rachit7168 (approved):
LGTM , Tested

https://github.com/user-attachments/assets/86b26d19-4c88-4669-bf61-d723b1ddc081

Saksham-Sirohi (approved):

Issue #4606, opened 2026-07-28
No application-level overload protection or graceful degradation


ArnavBallinCode:
**Parent issue:** Eventyay performance and resilience under event traffic  
**Component:** Django middleware, `app/eventyay/config/settings.py`, application architecture  
**Severity:** High — without overload protection, any capacity boundary causes a hard failure rather than a controlled degradation

---

## Problem

When Eventyay approaches capacity, it has no mechanism to:

1. Shed non-essential load to protect critical endpoints.
2. Return informative responses to overloaded clients instead of timing out silently.
3. Serve stale cached data when the database is slow.
4. Circuit-break failing dependencies (database, Redis) so slow dependencies do not cascade.

The result: a traffic spike causes broad application unavailability rather than a partial degradation that preserves check-in, registration, and other critical workflows.

---

### 1. Add a concurrency-limiting middleware

Install and configure `django-ratelimit` or implement a custom middleware that rejects requests when the worker queue depth exceeds a threshold:

```python
# app/eventyay/middleware/overload.py

import threading
from django.http import HttpResponse

_active_requests = 0
_lock = threading.Lock()
MAX_CONCURRENT = int(os.environ.get('MAX_CONCURRENT_REQUESTS', 50))

SHED_EXEMPT_PREFIXES = (
    '/api/v1/checkin',
    '/api/v1/organizers/',   # Narrow to only checkin paths in production
    '/health',
)

class LoadSheddingMiddleware:
    def __init__(self, get_response):
        self.get_response = get_response

    def __call__(self, request):
        global _active_requests
        
        # Never shed check-in or health requests
        if any(request.path.startswith(p) for p in SHED_EXEMPT_PREFIXES):
            return self.get_response(request)
        
        with _lock:
            if _active_requests >= MAX_CONCURRENT:
                response = HttpResponse(
                    '{"detail": "Server is temporarily overloaded. Please try again shortly."}',
                    status=503,
                    content_type='application/json',
                )
                response['Retry-After'] = '10'
                return response
            _active_requests += 1
        
        try:
            return self.get_response(request)
        finally:
            with _lock:
                _active_requests -= 1
```

Add to `MIDDLEWARE` in settings after `SecurityMiddleware` but before `SessionMiddleware`:

```python
MIDDLEWARE = [
    'django.middleware.security.SecurityMiddleware',
    'eventyay.middleware.overload.LoadSheddingMiddleware',
    # ... rest of middleware
]
```

[...]

Issue #4604, opened 2026-07-28
ORM N+1 queries and missing caching on high-traffic public API endpoints


ArnavBallinCode:
**Parent issue:** Eventyay performance and resilience under event traffic  
**Component:** `app/eventyay/api/`, `app/eventyay/schedule/`, `app/eventyay/agenda/`  
**Severity:** Medium-High — N+1 queries on endpoints hit by every schedule page load multiply DB load with attendee count

---

## Problem

During a large event, public-facing API endpoints for schedules, sessions, speakers, rooms, and tracks are loaded by every attendee. If these endpoints have N+1 query patterns, adding 100 more attendees adds 100× the DB queries, not 100 additional flat queries.

Known concerns from the codebase:

### 1. `streams/current` — already covered in sub-issue 03

Each room requires a separate query to `StreamSchedule`. A schedule page with 10 rooms generates 10 queries minimum.

### 2. Session/submission list endpoints

Session list APIs that return speaker relationships, track assignments, room assignments, and tag data are common sources of N+1 queries. Without `select_related`/`prefetch_related`, each session object triggers additional queries for:
- Its speakers (`Person` objects)
- Its room assignment
- Its track
- Its submission type

A schedule with 200 sessions could generate 600–800 DB queries per page load.

### 3. Speaker/person endpoints

If speaker lists fetch submission counts, room assignments, or avatar URLs in a loop per speaker, the pattern compounds.

### 4. Event configuration API

If the event configuration endpoint queries plugin settings, feature flags, or organizer data in a loop, it runs on every page load.

---

### Audit step (must be done first)

Enable Django's query count logging for the test environment:

```python
# In settings (DEBUG only):
LOGGING = {
    'loggers': {
        'django.db.backends': {
            'level': 'DEBUG',
            'handlers': ['console'],
        },
    }
}
```

Or use `django-debug-toolbar` or `silk` to profile queries per endpoint during load testing.

For each of the following endpoints, record:
- Total DB queries per request (cold and warm)
- Queries that repeat with the same table but different IDs (N+1 signature)

Priority endpoints to audit:
- `GET /api/v1/organizers/{org}/events/{event}/sessions/`
- `GET /api/v1/organizers/{org}/events/{event}/speakers/`
- `GET /api/v1/organizers/{org}/events/{event}/schedule/`
- `GET /api/v1/organizers/{org}/events/{event}/rooms/`
- `GET /api/v1/organizers/{org}/events/{event}/tracks/`
[...]

Issue #4603, opened 2026-07-28
Invalid and scanning URLs trigger full Django template rendering


ArnavBallinCode:
**Parent issue:** Eventyay performance and resilience under event traffic  
**Component:** `app/eventyay/base/views/errors.py`, Nginx config, URL routing  
**Severity:** Medium — amplifies impact of automated scanner traffic that coincided with the Wikimania incident

---

### 404 handler renders a full Django template

`app/eventyay/base/views/errors.py` lines 48–78:

```python
@requires_csrf_token
def page_not_found(request, exception):
    context = {
        'request_path': request.path,
        'exception': exception_repr,
    }
    template = get_template('404.html')
    body = template.render(context, request)
    return HttpResponseNotFound(body)
```

For every request to a nonexistent URL — including `/wp-admin/`, `/.env`, `/phpmyadmin/`, and invalid room IDs — Django:

1. Runs the full middleware stack (session, CSRF, authentication, organizer resolution, etc.)
2. Resolves the URL and raises `Http404`
3. Runs `get_template('404.html')` with a full template render
4. May execute organizer/event lookups in context processors

This is expensive. During the incident, the logs show `Not Found` responses mixed with `Forbidden` — scanning traffic was hitting the application alongside the polling storm.

### No edge-level rejection of clearly invalid paths

There is no Nginx `location` block that rejects common scanning targets before they reach Gunicorn. Every request to `/wp-admin`, `/.git`, `/xmlrpc.php`, etc. reaches Python.

### 403/401 responses also run the full middleware stack

When a request to `streams/current` returns 403 (Forbidden), the full DRF permission check runs: authentication, permission class evaluation, event lookup. This is not free.

---

### 1. Make the 404 handler return a static response

```python
# app/eventyay/base/views/errors.py

from django.http import HttpResponseNotFound

@requires_csrf_token
def page_not_found(request, exception):
    # Return a static, minimal 404 without database access or template rendering.
    # Template rendering for 404 is unnecessary and expensive under scanning traffic.
    return HttpResponseNotFound(
        b'<html><body><h1>404 Not Found</h1></body></html>',
        content_type='text/html'
    )
```

If a styled 404 page is required, pre-render it to a static file at deploy time and serve it from Nginx directly.

### 2. Reject common scanner paths at Nginx before reaching Django

Add to `deployment/nginx/enext-direct` or equivalent:

```nginx
# Reject common automated scanner targets before reaching the application.
[...]

PR #4945, opened 2026-08-15
fix: reduce video client polling load on streams and notifications

### LINKED ISSUE #4595: Incident Log Analysis:  Traffic Spike
## What the Logs Show

The logs captured during the event show the following pattern over roughly 7 seconds (13:21:22–13:21:29):

```
13:21:22,135  Forbidden:      /api/v1/organizers/wm/events/wm/rooms/8/streams/current
13:21:22,157  Forbidden:      /api/v1/organizers/wm/events/wm/rooms/163/streams/current
13:21:22,192  Forbidden:      /api/v1/organizers/wm/events/wm/rooms/8/streams/current
13:21:22,328  Forbidden:      /api/v1/organizers/wm/events/wm/rooms/165/streams/current
13:21:22,397  Unauthorized:   /api/v1/organizers/wm/events/wm/submissions/RMDAZ7
13:21:22,596  Unauthorized:   /api/v1/organizers/wm/events/wm/rooms/167/streams/current
13:21:23,018  Forbidden:      /api/v1/organizers/wm/events/wm/rooms/166/streams/current
...
```

**Endpoint hit:** `/api/v1/organizers/wm/events/wm/rooms/{room_id}/streams/current`  
**Room IDs hit:** 8, 163, 165, 166, 167, 162, 171, 172, 173, 174, 177  
**Status codes:** 403 Forbidden, 401 Unauthorized, 404 Not Found  
**Rate:** Dozens of requests per second, across multiple rooms simultaneously  
**Duration:** Several minutes (visible log span is 7 seconds)

---

### Primary cause: Synchronized frontend polling with no error backoff

**File:** `app/eventyay/webapp/video/src/store/index.js` — Lines 219–273

The video webapp polls `/rooms/{id}/streams/current` every **10 seconds** for every room the user can see. When an attendee opens a schedule page with multiple rooms listed, polling starts for each room simultaneously. With N attendees all opening the page at the same event start time, the polling cycles synchronize into a burst.

Critically: the polling **does not back off on error**. When a request returns 403, 401, or 404, the interval timer continues unchanged. Every attendee failing authentication still generates a request every 10 seconds per room, indefinitely.

**Secondary:** `app/eventyay/webapp/video/src/main.js` Line 141 polls `/notifications/externals` every **1 second** — an extreme rate that amplifies any load spike.

### Secondary cause: No rate limiting on the streams API

**File:** `app/eventyay/config/settings.py` — Lines 1335–1349

`REST_FRAMEWORK` has no `DEFAULT_THROTTLE_CLASSES`. Unauthenticated and unauthorized clients can call stream endpoints without limit. Rate limiting exists only for the contact form (`presale/views/contact.py`).

### Third cause: No caching on the stream endpoint

**File:** `app/eventyay/api/views/room.py` — Lines 68–77  
**File:** `app/eventyay/base/models/room.py` — Lines 392–407

`current_stream()` runs a `StreamSchedule.objects.filter(...)` database query **on every request**. There is no `cache_page`, no `Cache-Control` header, no Redis lookup. With dozens of requests per second, each generates a DB query.

### Fourth cause: Only 2 synchronous Gunicorn workers

**File:** `deployment/docker-compose.yml` — Line 7

```bash
gunicorn ... --workers 2 --timeout 120
```

Two sync workers means only 2 requests can be processed simultaneously. All others queue. A sustained 403/401 storm from polling quickly saturates both workers and stalls the entire application.

### Fifth cause: 404 handler renders a full Django template

**File:** `app/eventyay/base/views/errors.py` — Lines 48–78

The 404 handler runs `get_template('404.html')` and `template.render(context, request)`. Invalid paths from automated scanners (e.g. `/wp-admin`, `/admin`, unknown rooms) each trigger a full template render. Combined with no rate limiting, scanner traffic consumes significant worker capacity.

---

## Why Public Schedule Traffic Caused a Broader Outage

1. Attendees open schedule pages → video webapp starts 10-second polling per room.
2. Unauthenticated/unauthorized users generate sustained 403/401 responses but polling continues.
3. With no throttling, hundreds of concurrent failing polls reach Django.
4. Each request hits the DB (no cache on `streams/current`).
5. Only 2 sync Gunicorn workers → request queue grows.
6. Worker timeout of 120 seconds means stuck requests hold workers for 2 minutes.
7. Checkin, registration, and badge endpoints share the same 2 workers → go down with the rest.

### LINKED ISSUE #4599: Frontend stream polling has no error backoff and no visibility guard
**Component:** `app/eventyay/webapp/video/src/`  
**Severity:** High — directly caused the synchronized 403/401 burst seen in logs

---

## Problem

`app/eventyay/webapp/video/src/store/index.js` polls `/api/v1/organizers/{org}/events/{event}/rooms/{roomId}/streams/current` every **10 seconds**, unconditionally.

Key problems in the current implementation (lines 219–273):

1. **No backoff on error.** When the server returns 403, 401, or 404, the `catch` block stores the error state but `setInterval` continues at the same 10-second rate forever. Every authenticated failure still generates a database query every 10 seconds.

2. **No visibility guard.** Polling runs even when the browser tab is hidden (`document.visibilityState === 'hidden'`). If an attendee leaves the schedule open in a background tab for the duration of a multi-day event, requests continue the entire time.

3. **No deduplication.** If `startStreamPolling()` is called multiple times (e.g. route change, component remount), multiple `setInterval` timers accumulate. Each fires independently every 10 seconds, multiplying request volume per client.

4. **Per-room polling for all visible rooms.** When a schedule page shows multiple rooms, polling starts for each room ID independently. An event with 10 rooms running in 5 attendee browser tabs generates 50 requests every 10 seconds — 300 per minute — all unauthenticated, all hitting the database.

5. **No max-retry cap.** There is no upper bound on how many consecutive errors are tolerated before polling stops. A permanently unauthorized client polls indefinitely.

---

## Evidence from logs

The logs show room IDs 8, 162, 163, 165, 166, 167, 171, 172, 173, 174, 177 all being polled simultaneously within milliseconds of each other, returning 403/401/404. This is the signature of synchronized polling starting at the same time (event opening) from multiple clients, with no backoff after receiving error responses.

---

## Required changes

**File:** `app/eventyay/webapp/video/src/store/index.js`

### 1. Stop polling on permanent errors (403, 401, 404)

```js
// In the fetchCurrentStream catch block:
if (error.response && [400, 401, 403, 404].includes(error.response.status)) {
  // Permanent error — this client is not authorized or the room does not exist.
  // Stop polling immediately.
  commit('stopStreamPolling')
  return
}
```

### 2. Add exponential backoff with jitter on transient errors (5xx, network failure)

```js
// Replace fixed setInterval with a backoff-aware scheduler:
let retryDelay = 10000  // Start at 10s
const MAX_DELAY = 300000  // Cap at 5 minutes
const JITTER = () => Math.random() * 2000  // ±2s jitter

function scheduleNextPoll() {
  const delay = retryDelay + JITTER()
  pollingTimer = setTimeout(async () => {
    await dispatch('fetchCurrentStream', roomId)
    retryDelay = 10000  // Reset on success
    scheduleNextPoll()
  }, delay)
}

// On transient error, double the delay:
retryDelay = Math.min(retryDelay * 2, MAX_DELAY)
```

### 3. Pause polling when the tab is hidden

```js
// In startStreamPolling:
document.addEventListener('visibilitychange', () => {
  if (document.hidden) {
    clearTimeout(pollingTimer)
  } else {
    retryDelay = 10000  // Reset delay on resume
    scheduleNextPoll()
  }
})
```

### 4. Guard against duplicate polling timers

```js
// Before starting polling:
if (state.streamPollingActive) return
commit('setStreamPollingActive', true)
```

### 5. Add a max consecutive error count

```js
// Stop polling after 5 consecutive transient errors:
if (state.streamPollingErrorCount >= 5) {
  commit('stopStreamPolling')
  return
}
```

---

## Acceptance criteria

- [ ] A client receiving 403, 401, or 404 from `streams/current` stops polling immediately.
- [ ] A client receiving 5xx or network error backs off with exponential delay + jitter, capped at 5 minutes.
- [ ] Polling pauses when `document.visibilityState === 'hidden'` and resumes on return to visibility.
- [ ] Calling `startStreamPolling()` more than once does not create duplicate timers.
- [ ] After 5 consecutive transient errors, polling stops entirely.
- [ ] Unit tests cover the backoff, visibility pause, and permanent-error-stop paths.

---

## Related

- Sub-issue 02: Notifications endpoint polled every 1 second
- Sub-issue 03: `streams/current` endpoint has no server-side cache
- Sub-issue 04: No rate limiting on stream API endpoints

### LINKED ISSUE #4600: External notifications polled every 1 second
**Component:** `app/eventyay/webapp/video/src/main.js`  
**Severity:** High — 1-second polling from every connected client is unsustainable at any meaningful attendee count

---

## Problem

`app/eventyay/webapp/video/src/main.js` line 141:

```js
setInterval(() => store.dispatch('notifications/pollExternals'), 1000)
```

This fires a request to the notifications endpoint **every 1 second** for every client with the video webapp open. This is unconditional — no visibility guard, no backoff, no error handling.

At WM  scale (hundreds of concurrent attendees), this generates hundreds of requests per second to a single endpoint continuously, independent of whether any notifications have changed.

### Additional problem: `updateNow` polling

Line 138:
```js
setInterval(() => store.commit('updateNow'), 60000)
```

This is client-side only (no network request) so it is not a server concern, but it illustrates the pattern of unconstrained polling throughout the codebase.

### Why 1 second is wrong

- Notifications for a talk/session event do not change on a 1-second cadence.
- If 500 attendees are viewing a schedule page, this generates **500 requests/second** to `pollExternals` alone, before any other traffic.
- This endpoint is not cached (no Cache-Control or cache_page observed).
- Each request runs through Django middleware, authentication, and likely a database read.

---

## Required changes

**File:** `app/eventyay/webapp/video/src/main.js`

### 1. Increase the base polling interval to a minimum of 30 seconds

```js
setInterval(() => store.dispatch('notifications/pollExternals'), 30000)
```

Notification latency of up to 30 seconds is acceptable for schedule announcements and room messages.

### 2. Add visibility guard

```js
function pollExternalsIfVisible() {
  if (!document.hidden) {
    store.dispatch('notifications/pollExternals')
  }
}
setInterval(pollExternalsIfVisible, 30000)
```

### 3. Add jitter to desynchronize clients

Multiple clients starting at the same time (event opening) will synchronize their polling cycles. Add startup jitter to spread requests across the interval:

```js
const jitter = Math.random() * 15000  // 0–15s startup offset
setTimeout(() => {
  setInterval(pollExternalsIfVisible, 30000)
}, jitter)
```

### 4. Stop polling on permanent errors

In `notifications/pollExternals` action, stop the polling interval on 401/403:

```js
// In the Vuex action:
try {
  await api.fetchExternalNotifications()
} catch (error) {
  if (error.response && [401, 403].includes(error.response.status)) {
    // Not authorized — no point retrying
    clearInterval(notificationPollInterval)
  }
}
```

### 5. Consider server-sent events or WebSocket for live notifications

For true real-time notifications, replace polling with a WebSocket or SSE subscription (Eventyay already uses Django Channels). This eliminates polling overhead entirely and is more accurate.

---

## Impact estimate

Changing from 1-second to 30-second polling with visibility guard reduces notification traffic by approximately **97%** (30× reduction) for active clients and eliminates all traffic from background tabs.

---

## Acceptance criteria

- [ ] `pollExternals` is called no more frequently than every 30 seconds.
- [ ] Polling does not occur while the browser tab is hidden.
- [ ] Client startup jitter prevents synchronized polling bursts.
- [ ] 401/403 responses stop polling for that session.
- [ ] Load test shows >95% reduction in notifications endpoint traffic compared to baseline.

---

## Related

- Sub-issue 01: Stream endpoint polled every 10 seconds with no error backoff
- Sub-issue 04: No server-side rate limiting on API endpoints

Saksham-Sirohi:
## Summary

Part **1/3** of the load-hardening work.

- Live rooms include `currentStream` on websocket join so the player can skip an immediate HTTP fetch.
- `streams/current` HTTP polling runs **only when the websocket is down**: 60s interval with jitter, tab-visibility pause, exponential backoff, and stop on 401/403. Idle stream (404) keeps fallback polling, and transient 400s now retry instead of stopping polling for the rest of the session.
- Notification checks run every 30s when the tab is visible, with startup jitter.

Closes #4599
Closes #4600
Closes #4595

## Related PRs

| PR | Scope | Issues |
|---|---|---|
| **#4945 (this)** | Video client polling | #4599, #4600, #4595 |
| [#4992](https://github.com/fossasia/eventyay/pull/4992) | Stream/schedule Redis cache, N+1 prefetch | #4601, #4604 |
| [#4993](https://github.com/fossasia/eventyay/pull/4993) | Load shedding, throttling, 404, Celery/nginx | #4602, #4603, #4606 |

Each PR is independently testable and can merge in any order.

## Test plan

Automated:
- [x] `node --test app/eventyay/webapp/video/src/store/streamPolling.test.js`

Manual:
- [ ] Livestream room with websocket connected: no HTTP `streams/current` polling while connected.
- [ ] Websocket down: HTTP fallback polling resumes; reconnect stops it.
- [ ] A transient 400 during fallback polling retries on the next interval instead of disabling polling for the room.
- [ ] Notification poll pauses when tab is hidden.

## Risk and rollback

Low risk — frontend-only plus `currentStream` on websocket join. Revert and rebuild the video bundle.


Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/62d39b53-7bf5-466a-8bdd-7c2a354115fd

ArnavBallinCode (changes requested):
I think this PR is too broad and should be split into several focused PRs. It currently combines changes across streaming, caching, frontend polling, notifications, HTTP error handling, session behavior, Sentry configuration, Nginx hardening, Gunicorn configuration, Celery routing/concurrency, and load shedding. The summary itself references multiple separate issues, which is a strong indication that these concerns do not necessarily need to land as one atomic change.

I would suggest splitting this into approximately the following PRs:

1. **Live stream synchronization**

   * Include `currentStream` in room configuration.
   * WebSocket `room.stream.change` synchronization.
   * Stream schedule/cache integration required specifically for this feature.
   * Related streaming tests.

2. **HTTP stream fallback polling**

   * `streamPolling.js`.
   * WebSocket disconnect → HTTP polling.
   * Backoff/jitter.
   * Visibility handling.
   * 400/401/403/404 behavior.
   * Frontend polling tests.

3. **Current-stream HTTP caching**

   * `streams/current` cache.
   * ETag / Last-Modified / 304.
   * Cache-Control behavior.
   * Schedule-change invalidation.
   * Cache-specific tests.

   This should also address transaction-safe invalidation, e.g. invalidating on transaction commit rather than directly from `post_save`/`post_delete`. The current implementation adds the signal-based invalidation separately from the HTTP caching logic.

4. **Notification polling changes**

   * 30-second notification polling.
   * Startup jitter.
   * Visibility handling.
   * Authentication failure handling.

   This is logically independent from livestream fallback and should be reviewable independently.

5. **HTTP 404 / session / Sentry cleanup**

   * JSON `/api/` 404 responses.
   * Avoiding session saves for unmodified 404s.
   * Sentry `Http404` / `PermissionDenied` filtering.

   These are general platform/error-handling changes and don't need to be coupled to livestream functionality.

6. **Worker/runtime configuration**

   * Gunicorn gthread configuration.
   * Daphne proxy headers / ulimits.
   * Celery concurrency/prefetch/max-tasks-per-child.
   * Celery queue routing.

   This should be treated as an infrastructure/runtime PR with its own load and regression testing. In particular, the current worker still consumes all queues while running with `-c 1`, so the claimed queue isolation needs to be validated independently.

7. **Load shedding**

   * `MAX_CONCURRENT_REQUESTS`.
   * 503 + `Retry-After`.
   * Exemptions.
   * Load-shedding tests.

   This is an independent availability/scalability mechanism and should have its own rollout and rollback strategy.

8. **Nginx scanner protection**

   * Scanner-path regex.
   * `limit_req_zone`.
   * WebSocket proxy timeout changes.
   * Nginx tests.

   This is security/infrastructure work and should be independently deployable. Also, the current `limit_req` + `return 444` arrangement needs to be verified because the test currently checks configuration text rather than actual Nginx behavior.

### Why I think this matters

Having all of these changes in one PR makes it difficult to answer basic review questions such as:

* Which change caused a regression?
* Which part needs to be rolled back?
* Can the streaming feature be deployed without the infrastructure changes?
* Can the Nginx change be tested independently?
* Are the Celery changes required for the streaming feature?
* Which tests provide confidence for each production behavior?

It also makes the PR substantially harder to review because a reviewer has to reason about frontend state management, Django caching, middleware behavior, Celery scheduling, Gunicorn, Daphne, and Nginx in the same change.

I would strongly prefer smaller, cohesive PRs where each PR has one primary objective and can be independently tested, reviewed, deployed, and reverted.

If some of these changes are prerequisites for the streaming work, they can still be landed first as small dependency PRs and then referenced from the main streaming PR.

**In short: this looks more like a small stack of related PRs than one PR. Splitting it would materially reduce review and deployment risk.**

deployment/nginx/enext-direct
```diff
@@ -42,6 +44,13 @@ server {
 		set $cors_reject_preflight 0;
 	}
 
+	# Drop common scanner probes. Paths are anchored so /.github is not blocked.
+	location ~* "^/(wp-admin(/|$)|wp-login(\.php)?$|phpmyadmin(/|$)|xmlrpc\.php$|\.env$|\.git(/|$))" {
+		limit_req zone=scanner_limit burst=20 nodelay;
+		limit_req_status 429;
+		return 444;
+	}
```
ArnavBallinCode: Restructure this so the rate limiting actually executes before the connection is dropped, or use a mechanism/location structure that lets the request reach the `limit_req` phase. 

Then add an actual nginx-level test/manual verification that repeated requests from one IP hit the rate limit.

app/eventyay/base/models/stream_schedule.py
```diff
@@ -98,3 +100,11 @@ def is_active(self, at_time=None):
...
         at_time = at_time or now()
         return self.start_time <= at_time < self.end_time
+
+
+@receiver(post_save, sender=StreamSchedule)
+@receiver(post_delete, sender=StreamSchedule)
+def invalidate_current_stream_cache_on_change(sender, instance, **kwargs):
+    from eventyay.base.services.room import invalidate_current_stream_cache
+
+    invalidate_current_stream_cache(instance.room_id)
```
ArnavBallinCode: The cache invalidation should happen after the transaction commits.

`post_save` / `post_delete` can run while the surrounding transaction is still uncommitted. This creates a race where the cache is invalidated, another request reads the still-committed old value and repopulates the cache, and then the original transaction commits without another invalidation. The old stream can therefore remain cached until the TTL expires.

Could we schedule the invalidation with `transaction.on_commit(...)` instead, and add a regression test covering the transaction boundary/concurrent read case?

This is particularly important here because the cache is being introduced specifically to keep `streams/current` efficient and consistent. The current invalidation is implemented directly in the model signals.

deployment/docker-compose.yml
```diff
@@ -33,14 +37,15 @@ services:
...
+      - MALLOC_ARENA_MAX=2
     depends_on:
       - db
       - redis
 
   worker:
     image: eventyay/eventyay-next:${TAG}
     container_name: eventyay-next-worker
-    entrypoint: celery -A eventyay worker -l info
+    entrypoint: celery -A eventyay worker -l info -c 1 --prefetch-multiplier=1 --max-tasks-per-child=200
```
ArnavBallinCode: I think the Celery worker configuration needs another look.

The worker is now running with `-c 1 --prefetch-multiplier=1`, but it still consumes all queues, including `longrunning`. The PR routes potentially long-running work such as badges, exports, order imports, and ticket generation to that queue.
With concurrency set to 1, a long-running task can occupy the only worker slot while tasks from the default/notifications queues are waiting behind it. `--prefetch-multiplier=1` limits how much the worker reserves, but it does not provide queue/process isolation.

This is particularly relevant given that the goal of this PR is to reduce worker stress and avoid unnecessary 504s.

ArnavBallinCode (changes requested):

deployment/nginx/enext-direct
Saksham-Sirohi: Addressed in 23d0e762f4:

- Moved scanner rate limiting to **server scope** with a `map $request_uri $scanner_limit_key` so `limit_req` runs in the ACCESS phase before the scanner location’s `return 444` (CONTENT phase).
- Scanner location now only drops the connection; excess probes get **429** from the server-level limiter.
- Added a structural regression test asserting server-level `limit_req` appears before the scanner `return 444` and is not duplicated inside the location block.

Manual verification on nginx: `for i in {1..30}; do curl -s -o /dev/null -w \"%{http_code}\n\" https://<host>/wp-admin; done` — expect 444 then 429.

deployment/docker-compose.yml
Saksham-Sirohi: Addressed in 23d0e762f4:

Reverted the worker entrypoint to the default `celery -A eventyay worker -l info` (removed `-c 1`, `--prefetch-multiplier=1`, and `--max-tasks-per-child=200`).

The `longrunning` task routes remain as forward-looking queue assignment for ops that add a dedicated heavy worker later; they do not require a second worker in the default single-node compose file and no longer risk a single slot blocking default/notifications work.

The primary 504 mitigation on the HTTP side is gthread (2×4) plus stream-cache/websocket-first polling; Celery concurrency is left at framework defaults to avoid regressing badge/import/export throughput on the same host.

app/eventyay/base/models/stream_schedule.py
Saksham-Sirohi: Addressed in 23d0e762f4:

- Invalidation is now scheduled with `transaction.on_commit(lambda: invalidate_current_stream_cache(room_id))` so uncommitted writes cannot leave a stale repopulated cache.
- Added `test_stream_schedule_cache_invalidation_runs_on_commit` — cache remains populated inside `transaction.atomic()` and is cleared only after commit.

Saksham-Sirohi:
@ArnavBallinCode Thanks for the detailed review — addressed the three inline threads in 23d0e762f4 and updated the PR description with split rationale for the Wikimedia 504 urgency.

**Bundled vs split:** The core path (websocket-first stream sync + cached `streams/current` + gthread + nginx scanner limiting + load shedding) targets the same production failure mode on a constrained host. Notification polling, JSON 404, and Sentry filtering could ship separately but were kept together to unblock Wikimedia without multiple deploy cycles mid-incident. The PR body now lists what *could* be split later.

Also fixed: session-cookie / authenticated `no-store` on `streams/current` (not only `Authorization` header).

Rachit7168 (approved):
LGTM , Tested 

 In notifications.js, you use setTimeout for the initial jitter and then fall back to a strict setInterval. Because the initial jitter spreads all the connected clients out randomly across a 15-second window, their subsequent 30-second intervals will naturally stay staggered. That is a very smart and efficient way to handle it!

ArnavBallinCode (approved):
LGTM

PR #4959, opened 2026-08-15
fix: use ngettext for max_products cart size error pluralization

### LINKED ISSUE #4951: Fix pluralization (i18n) for max products error message
### Description
In `app/eventyay/base/services/cart.py`, we raise a `CartError` when a user tries to add more products to their cart than allowed. The error message is currently translated using `_()`, but it does not account for plurals correctly (e.g. "1 products" vs "2 products").

There is a `TODO` comment indicating this around line 329:

```python
            if max_products > 0 and cartsize > max_products:
                # TODO: i18n plurals
                raise CartError(_(error_messages['max_products']) % (max_products,))
```

### Expected Behavior
The error message should be translated using `ngettext` instead of just `_()` so that languages with different pluralization rules can render the sentence correctly.

1. Import `ngettext` from `django.utils.translation`.
2. Update the `error_messages['max_products']` usage (you may need to find where this error message is defined and adjust it to support singular/plural forms).
3. Replace the current `_()` call with `ngettext()` passing in the singular string, plural string, and the `max_products` count.
4. Remove the `# TODO: i18n plurals` comment.

### Steps to Reproduce / Fix
1. Clone the repository and set up your local development environment according to the `README.md`.
2. Open `app/eventyay/base/services/cart.py` and find line 329.
3. Check where `error_messages['max_products']` is defined (likely near the top of the file or in another file imported in `cart.py`).
4. Update the translation logic to use `ngettext`.
5. Run the test suite to ensure no tests are broken and write a small test if possible.
6. Commit your changes and open a Pull Request!

### Notes for First-Time Contributors
- For more info on how Django handles pluralization, see the [Django ngettext documentation](https://docs.djangoproject.com/en/stable/topics/i18n/translation/#pluralization).
- If you have any questions, feel free to ask in the comments!

shivanshumangal007-dev:
Closes #4951

## Problem
`CartError` for the max-products-per-order limit used `_()` on a
single string containing `%s`, so the message always rendered in
plural form ("...more than 1 products...") regardless of the count.
Since Django's `gettext` has no concept of count, this also meant
translators had no way to supply correct plural rules for languages
where pluralization doesn't map 1:1 to English.

## Fix
- Split `error_messages['max_products']` into singular and plural
  raw strings (no longer pre-wrapped in `_()`), since `ngettext`
  needs the untranslated forms to do its own lookup.
- Resolved the message with `ngettext(singular, plural, count)` at
  the raise site in `_check_max_cart_size`, where `count` is known.
- Switched formatting from positional `%s` to named `%(max)s` to
  match the convention already used elsewhere in this file
  (`max_products_per_product`, `addon_max_count`).

## Testing
- Ran `pytest tests/ -k cart` — all passing.
- Manually verified both singular (`max_products=1`) and plural
  (`max_products>1`) cart-limit errors render correctly.

## Summary by Sourcery

Improve cart error messaging for max products per order and clean up related cart service formatting.

Bug Fixes:
- Fix max-products-per-order cart error message to correctly handle singular and plural forms using count-aware translation.

Enhancements:
- Add singular and plural raw message templates for max-products-per-order errors and switch to named formatting for consistency.
- Reformat long lines and conditional blocks in cart services for readability without changing behavior.

### DIFF
--- a/app/eventyay/base/services/cart.py
+++ b/app/eventyay/base/services/cart.py
@@ -9,7 +9,7 @@ from django.db import DatabaseError, transaction
 from django.db.models import Count, Exists, IntegerField, OuterRef, Q, Value
 from django.dispatch import receiver
 from django.utils.timezone import make_aware, now
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext as _, ngettext
 from django.utils.translation import pgettext_lazy
 from django_scopes import scopes_disabled
 
@@ -73,7 +73,8 @@ error_messages = {
         'Some of the products you selected are no longer available in '
         'the quantity you selected. Please see below for details.'
     ),
-    'max_products': _('You cannot select more than %s products per order.'),
+    'max_products': 'You cannot select more than %(max)s product per order.',
+    'max_products_plural': 'You cannot select more than %(max)s products per order.',
     'max_products_per_product': _('You cannot select more than %(max)s products of the product %(product)s.'),
     'min_products_per_product': _('You need to select at least %(min)s products of the product %(product)s.'),
     'min_products_per_product_removed': _(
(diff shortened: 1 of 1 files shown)


Rachit7168 (changes requested):
Thanks for your first contribution !

Could you also add a short demo showing the feature working? Please include a video demonstrating it by trying out multiple languages.

PR #5033, opened 2026-08-18
fix(settings): introduce CACHE_TICKETS_MAX_AGE derived from CACHE_TICKETS_HOURS

Rachit7168:
**Problem**
Part of  #1546 

**Before**
The `CACHE_TICKETS_HOURS` setting was missing from the configuration. The ticket cleanup task (`clean_cached_tickets`) would attempt to build a `timedelta(hours=settings.CACHE_TICKETS_HOURS)` and throw an `AttributeError`, causing the background cleanup process to fail completely.

**After**
The setting has been properly introduced via `BaseSettings` (`cache_tickets_hours: int = 24`). In `settings.py`, it is explicitly evaluated as a timedelta named `CACHE_TICKETS_MAX_AGE`. The `clean_cached_tickets` task in `cleanup.py` was then refactored to use this new timedelta constant directly instead of generating one dynamically inline, resolving the error efficiently.

## Summary by Sourcery

Restore reliable cleanup of expired cached tickets through a configurable cache age.

Bug Fixes:
- Fix cached ticket cleanup failures by introducing a configured maximum cache age and using it when removing expired tickets.

Enhancements:
- Centralize ticket cache expiry as a reusable timedelta derived from the configured retention period.

Tests:
- Add coverage verifying expired cached tickets are removed while recent tickets are retained.


Saksham-Sirohi:
The change adds `cache_tickets_max_age` defaulting to `timedelta(days=7)`, but does not define or map `CACHE_TICKETS_HOURS` / `cache_tickets_hours`. Cleanup retention becomes 7 days unless operators set the new env var, and existing hours-based config has no equivalent.

Saksham-Sirohi (changes requested):
## Request changes

The cleanup bugfix is the right idea (`CACHE_TICKETS_HOURS` is undefined today, so `clean_cached_tickets` AttributeErrors and generated ticket files never age out). Please align the implementation with the PR description before merge.

### 1. Title/body do not match the diff

The PR says you add `cache_tickets_hours: int = 24` and derive `CACHE_TICKETS_MAX_AGE` from it. The diff instead adds:

```python
cache_tickets_max_age: datetime.timedelta = datetime.timedelta(days=7)
```

`CACHE_TICKETS_HOURS` is never introduced. Please either:

- **Prefer:** `cache_tickets_hours: int = 24` (or another documented hour count) and set `CACHE_TICKETS_MAX_AGE = timedelta(hours=conf.cache_tickets_hours)`, or
- Keep a timedelta setting with a **7-day** default and rewrite the title/description so operators know the TTL is 7 days, configured as `EVY_CACHE_TICKETS_MAX_AGE`.

Do not ship a 7-day default while the write-up still claims 24 hours.

### 2. Operator config

`EVY_CACHE_TICKETS_MAX_AGE` as a pydantic `timedelta` is harder to set in TOML/env than an integer hour count. If you keep hours as the public setting, operators can set `EVY_CACHE_TICKETS_HOURS=24` without duration syntax.

### 3. Tests

Please add a small regression so this cannot silently break again: assert `CACHE_TICKETS_MAX_AGE` (or hours) exists on Django settings, and that `clean_cached_tickets` uses it without `AttributeError`.

No security concern with the approach itself; ticket PDFs just need an intentional, documented retention period.

app/tests/base/test_cleanup.py
```diff
@@ -0,0 +1,36 @@
...
+from datetime import timedelta
+from django.conf import settings
+from django.utils.timezone import now
+
+from eventyay.base.models import CachedCombinedTicket, CachedTicket
+from eventyay.base.services.cleanup import clean_cached_tickets
+
+
+@pytest.mark.django_db
+def test_clean_cached_tickets(event, order, order_position):
```
Saksham-Sirohi: These fixtures are not available here. `tests/base/` has no `conftest.py`; `event` / `order` / `order_position` are defined under `tests/tickets/api/` (and similar) and do not apply to this path.

This PR’s pytest job is `pytest tests/stable`, so this file never runs. A full `pytest tests/base/` would fail with `fixture 'event' not found`.

Move this to `app/tests/tickets/base/` and create the organizer/event/order/position in-module (see `test_shredders.py`), or put a self-contained test in `tests/stable/` so CI actually executes it.

Also set `type` and `extension` on `CachedTicket`, and use `file=None` instead of `file='test.pdf'` (the `post_delete` handler calls `instance.file.delete()`). Drop the unused `CachedCombinedTicket` import.

Saksham-Sirohi (changes requested):
## Re-review

The previous mismatch is fixed: `cache_tickets_hours: int = 24` and `CACHE_TICKETS_MAX_AGE = datetime.timedelta(hours=conf.cache_tickets_hours)`. Operators can set `EVY_CACHE_TICKETS_HOURS`. Cleanup no longer AttributeErrors. That part is good.

The regression test still does not work.

### Must fix
`app/tests/base/test_cleanup.py` requests `event`, `order`, and `order_position`, but those fixtures are not defined for `tests/base/` (they live under `tests/tickets/api/` and similar, directory-scoped). PR CI only runs `pytest tests/stable`, so this file is never executed and would fail with `fixture 'event' not found` if it were.

Please move it to `app/tests/tickets/base/` (same pattern as `test_shredders.py`: create organizer/event/order/position in the test module) **or** put a fixture-self-contained test under `tests/stable/` so the existing CI job actually runs it.

Also:
- `CachedTicket` needs `type` and `extension` (plain CharFields, no default).
- `file='test.pdf'` is not a real file; `CachedTicket` `post_delete` will try to delete it from storage. Use `file=None` (the 30-minute empty-file cleanup will not remove a ticket created just now).
- `CachedCombinedTicket` is imported and unused.

A minimal passing test is enough: assert `CACHE_TICKETS_MAX_AGE` is a timedelta, call `clean_cached_tickets(sender=None)`, and check one expired vs one fresh `CachedTicket`.

### Optional
- `cache_tickets_hours: int = Field(default=24, ge=1)` so `0` / negative hours cannot wipe or skip cleanup.
- Prefer `from datetime import timedelta` in `settings.py` instead of `import datetime`.

Production change is small and correct once the test actually runs.

Saksham-Sirohi (changes requested):
please look into test failures

ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):

PR #4768, opened 2026-08-06
Add order ticket data anonymization

SxxAq:
## Summary
Adds the ability to anonymize ticket sales and personal attendee data on an order without disabling or modifying the associated user account.

Closes #4627

## Changes
- `anonymize_order()` service that scrubs email, phone, attendee name/address, invoice address, question answers, and cached tickets, with an audit log entry written on completion
- `OrderAnonymize` view (GET confirmation page + POST action) protected by `can_change_orders` permission
- "Anonymize ticket data" button on the order detail toolbar
- REST API action: `POST /api/v1/organizers/{org}/events/{event}/orders/{code}/anonymize/`
- Unit tests covering service logic, data preservation, user account safety, and view endpoints
- Fix: guard `debug_toolbar` URL include against it being absent from `INSTALLED_APPS`

https://github.com/user-attachments/assets/acd1e3a6-801d-4ce0-b6bd-09ff390352b4

## Summary by Sourcery

Provide a safe, auditable way to permanently anonymize completed-order ticketing and personal data without altering customer accounts or financial reporting information.

New Features:
- Add order-level anonymization for ticket sales, attendee, billing, payment, questionnaire, invoice, cached-ticket, and historical log data while preserving financial records and associated user accounts.
- Expose anonymization through the order management interface and a protected REST API action, with confirmation, event-completion checks, permissions, and audit logging.

Bug Fixes:
- Prevent debug toolbar URL registration unless the toolbar is both installed and enabled.

Tests:
- Add coverage for anonymization service behavior, user-account and financial-data preservation, event and subevent eligibility, control views, API responses, and permissions.

Chores:
- Update order API test fixtures to use the current product relationships.

PR #5040, opened 2026-08-18
fix(presale): restrict order access to the authenticated owner

### LINKED ISSUE #4534: Restrict checkout and video access to the authenticated order owner
### Description

Currently, an Order Details page can be accessed using a direct order/token URL. However, access to actions such as checkout and joining the online event/video stream is not consistently restricted based on authentication and order ownership.

The expected behavior should be:

* An **unauthenticated user** should not be able to proceed to checkout or access protected event/video functionality.
* A **logged-in user who does not own the order** should also not be able to access or continue checkout using another user's order URL/token.
* Only the **authenticated user associated with the order** should be allowed to access protected order actions, including checkout and joining the online event/video stream.

Currently, direct access through an order/token URL can allow inconsistent behavior between different CTAs and potentially allow access to functionality without verifying that the current authenticated user is the owner of the order.

### Steps to Reproduce

1. Create or use an existing order belonging to **User A**.
2. Copy the direct Order Details URL containing the order/token information.
3. Log out and open the URL as a guest.
4. Attempt to access checkout or click the available event/video CTAs.
5. Log in as **User B**, who does not own the order.
6. Open the same Order Details URL.
7. Attempt to proceed to checkout or access the online event/video functionality.

### Current Behavior

Access control is inconsistent:

* Some actions redirect unauthenticated users to the login page.
* Other actions may continue using the order token context.
* A different authenticated user may potentially access functionality associated with an order that does not belong to them.

### Expected Behavior

Access should be consistently validated before allowing protected actions:

1. **User is not logged in**

   * Redirect to the login page.
   * Do not allow access to checkout or protected video/event functionality.

2. **User is logged in but does not own the order**

   * Deny access to the order's protected actions.
   * Do not allow the user to proceed to checkout or join the event/video using another user's order/token.

3. **User is logged in and owns the order**

   * Allow access to checkout and other authorized order functionality.
   * Allow access to the online event/video stream when the order and ticket permissions permit it.

### Acceptance Criteria

* [ ] Unauthenticated users cannot access checkout through a direct order/token URL.
* [ ] Unauthenticated users cannot access protected online event/video functionality.
* [ ] Users cannot access checkout for an order belonging to another account.
* [ ] Users cannot access protected event/video functionality using another user's order URL/token.
* [ ] Authorization is validated server-side and is not based solely on the presence of an order token in the URL.
* [ ] All checkout and video/event CTAs follow the same authentication and authorization rules.
* [ ] The actual order owner retains access to the appropriate functionality.

### Screenshots

https://github.com/user-attachments/assets/c0a266cb-cb69-4879-ba3f-b62210c06782

jagritixjha:
## Summary

This PR restricts access to protected order actions and sensitive order data so that only the authenticated order owner can view or act on their order.

## Why

The order token alone was being treated as sufficient context for protected actions. That allowed unauthenticated users and other signed-in users to access order details, downloads, and checkout-related flows for orders they did not own.

## Changes

* Added a shared authorization check for sensitive presale order views.
* Redirect unauthenticated users to the login flow before allowing access to protected order actions.
* Reject access for authenticated users whose email does not match the order owner email.
* Applied the protection consistently across order detail, download, payment, cancellation, and change flows.
* Added regression tests covering owner, unauthorized, and unauthenticated access.

## Testing

* Verified the relevant Django authorization regression tests in the project environment:
  * `tests/tickets/presale/test_order_authorization.py`

## Issue

Fixes #4534

## Summary by Sourcery

Enforce authenticated order-owner authorization across protected presale order views and actions.

Bug Fixes:
- Restrict presale order details, downloads, payment, invoice, cancellation, modification, and change flows to authenticated users who own the order.
- Redirect unauthenticated users to login and deny access to authenticated users whose email is not associated with the order.

Enhancements:
- Centralize authorization for protected presale order actions in a shared view mixin.

Tests:
- Add regression coverage for owner, unauthorized-user, and unauthenticated access to protected order views and actions.


app/eventyay/presale/views/order.py
```diff
@@ -144,8 +144,29 @@ def position(self):
...
+            from eventyay.common.views.helpers import build_login_url_with_next
+            return redirect(build_login_url_with_next(request.path))
+
+        if order and order.email:
+            if order.email.lower() != request.user.email.lower():
+                from django.core.exceptions import PermissionDenied
+                from django.utils.translation import gettext_lazy as _
+                raise PermissionDenied(_('You are not authorized to access this order.'))
+
+        return super().dispatch(request, *args, **kwargs)
```
Rachit7168, line 172: This is a great security addition! However, there are a few critical flaws with the current implementation that will break ticket sharing and file downloads, and cause server crashes.

1. **Locks out individual attendees (Ticket Sharing):** If a user buys 5 tickets for their team, the `Order` object uses the buyer's email (`order.email`). The individual tickets (`OrderPosition` objects) use the team members' emails (`position.attendee_email`). If a team member tries to view their individual ticket, they will get a `403 PermissionDenied` because their account email doesn't match the *Buyer's* `order.email`. You must allow access if their email matches `order.email` **OR** `position.attendee_email`.
2. **Query parameters are lost on login:** Using `request.path` strips all query parameters (like `?pdf=1` or `?download=1`). Unauthenticated users clicking a download link will lose the download parameter after logging in. Use `request.get_full_path()` instead.
3. **`AttributeError` crash:** If `request.user.email` happens to be empty, calling `.lower()` on it will crash the server.
4. **Code Style:** Imports should be at the top of the file, not inside the `dispatch` function.

Here is the corrected implementation that safely handles all of these cases. Please move the imports to the top of the file and replace your `dispatch` method with this:

```python
        self.request = request
        order = getattr(self, 'order', None)
        position = getattr(self, 'position', None)

        if order is None and position:
            order = position.order

        if not request.user.is_authenticated:
            return redirect(build_login_url_with_next(request.get_full_path()))

        if order:
            user_email = (request.user.email or '').lower()
            allowed_emails = { (order.email or '').lower() }
            
            if position and position.attendee_email:
                allowed_emails.add(position.attendee_email.lower())

            allowed_emails.discard('')

            if user_email not in allowed_emails:
                raise PermissionDenied(_('You are not authorized to access this order.'))
                
        return super().dispatch(request, *args, **kwargs)

jagritixjha: done

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/c16580eb-2ac4-46c8-a1f6-35f2ab94a9c1

Aqil-Ahmad (approved):
looks good. @Rachit7168 I think this will affect anonymized orders when this PR https://github.com/fossasia/eventyay/pull/4768 lands, as its using email for allowing.

Saksham-Sirohi (approved):

ArnavBallinCode (approved):

PR #5061, opened 2026-08-20
fix: add compilemessages to dev entrypoint so translations work locally

### LINKED ISSUE #5041: Translations not working in local dev — compilemessages missing from entrypoint.sh
### **Description**
All translations (every language, not just Bangla) are broken in local Docker dev because .mo files are never compiled. The UI shows English for everything except fragments that happen to have hardcoded translations.

### **Root cause**

app/entrypoint.sh (used by the dev Dockerfile) does not run python manage.py compilemessages
app/entrypoint.prod.sh has the call commented out (line 17)
The production Dockerfile.prod compiles at build time (line 100), so prod is unaffected
.gitignore excludes *.mo files, so they're never in the repo

### Expected behavior
Switching language in the frontend should show translated strings in local dev, same as production.

### Proposed fix
Add python manage.py compilemessages to app/entrypoint.sh after the migrate step (line 38).

Rishika3D:
Translations for all languages are broken in local Docker dev because .mo files are never compiled.

- Production Dockerfile.prod runs compilemessages at build time (line 100) — works fine
- Dev entrypoint.sh never calls it, and .gitignore excludes .mo files from the repo
- Result: every language silently falls back to English in local dev

This adds `python manage.py compilemessages` to entrypoint.sh after the migrate step, matching what production already does.

Fixes #5041

## Summary by Sourcery

Bug Fixes:
- Compile and correctly assign ownership to local development translation files so localized content works in Docker dev environments.


Rachit7168 (changes requested):
This is a helpful addition for local development, but there are two potential side effects to keep in mind:
 
1. **Container Crash Loop:** If this script has `set -e` at the top, any syntax error in a translation (`.po`) file or missing `gettext` utilities will cause `compilemessages` to exit with an error. This would crash the container on startup and block developers from even running the web server. Consider allowing it to fail gracefully:
    ```bash
    python manage.py compilemessages || true
    ```
 
2. **File Permissions (Root Ownership):** Since dev environments typically mount the source code as a Docker volume, generating `.mo` files here means they might be created with `root` ownership on the developer's host machine (if the container runs as root). This often leads to annoying "Permission denied" errors for developers when they run `git checkout`, `git clean`, or try to delete the `.mo` files locally. 
 
If the container isn't running as root (or you're okay with the ownership changes) and you don't mind the slight startup delay, adding `|| true` is probably the only minor tweak needed!

app/entrypoint.sh
```diff
@@ -36,6 +36,10 @@ if [ -d /usr/src/plugins ]; then
 fi
 
 python manage.py migrate
+python manage.py compilemessages -i .venv
+# Match ownership of generated .mo files to their .po siblings so host bind-mounts
```
Rachit7168: Remove comments explaining steps

Rachit7168 (approved):
LGTM

Saksham-Sirohi (approved):

Saksham-Sirohi (approved):

PR #5072, opened 2026-08-20
fix(mail): prevent noname attachment on emails without images

### LINKED ISSUE #4963: Unnamed attachment (`noname`) generated in transactional order confirmation emails
**Description**
Transactional emails sent from the platform (e.g., order confirmations from `test-deployment@eventyay.com`) include an empty or unnamed file attachment labeled as `noname` in email clients alongside the valid PDF attachments.

This usually indicates an issue with the multipart MIME structure (e.g., an inline HTML/plaintext body part being treated as an attachment or a missing `Content-Disposition: inline` / filename header).

**Steps to Reproduce**

1. Place an order or register for an event (e.g., Open Source Connect).
2. Check the received order confirmation email in an email client (e.g., Gmail).
3. Inspect the attached files at the bottom of the email.

**Expected Behavior**
Only the intended PDF ticket/badge attachments should be attached, with no extraneous `noname` files.

**Actual Behavior**
Three attachments are delivered: the intended PDFs and an additional `noname` attachment (MIME/HTML snippet).

Image

DRJ-YDV:
Fixes #4963

## Summary by Sourcery

Ensure HTML emails use multipart related content only when inline images are successfully attached, avoiding noname attachments in image-free messages.

Bug Fixes:
- Prevent HTML emails without successfully embedded images from being sent as empty or unnamed multipart attachments.
- Fall back to a standard HTML alternative when inline image conversion fails or no inline images are present.

Enhancements:
- Preserve text MIME alternatives correctly and add inline image filenames based on their detected subtype.

### DIFF
--- a/app/eventyay/base/services/mail.py
+++ b/app/eventyay/base/services/mail.py
@@ -371,7 +371,7 @@ def mail_send_task(
             attach_cid_images(html_message, cid_images, verify_ssl=True)
             email.attach_alternative(html_message, 'multipart/related')
         else:
-            email.attach_alternative(html, 'text/html')
+            email.attach_alternative(html_with_cid, 'text/html')
 
     if user:
         user = User.objects.get(pk=user)
--- a/app/eventyay/base/services/mail.py
+++ b/app/eventyay/base/services/mail.py
@@ -364,11 +364,14 @@ def mail_send_task(
 ) -> bool:
     email = CustomEmail(subject, body, sender, to=to, bcc=bcc, headers=headers)
     if html is not None:
-        html_message = SafeMIMEMultipart(_subtype='related', encoding=settings.DEFAULT_CHARSET)
         html_with_cid, cid_images = replace_images_with_cid_paths(html)
-        html_message.attach(SafeMIMEText(html_with_cid, 'html', settings.DEFAULT_CHARSET))
-        attach_cid_images(html_message, cid_images, verify_ssl=True)
-        email.attach_alternative(html_message, 'multipart/related')
+        if cid_images:
+            html_message = SafeMIMEMultipart(_subtype='related', encoding=settings.DEFAULT_CHARSET)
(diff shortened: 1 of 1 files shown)


Rachit7168 (changes requested):
Please address : 

1. Clean Abstraction: Moving the gating logic into room_creation_gate.py and removing the ad-hoc Jitsi checks from event.py makes the codebase much cleaner and easier to maintain moving forward.
2. Safe Fallbacks: In admin_dashboard.py, the use of getattr(server, "event_exclusive", None) gracefully handles models (like StreamingServer) that don't actually have an event_exclusive attribute, correctly falling back to the "All events" UI in the template without crashing.
3. Smart Permissions: In api_auth.py's has_object_permission, the logic to only enforce the admin trait if the user is adding a new server-backed module (newly_added_server_modules) is perfect. It ensures users can still safely update other aspects of the room (like its name) without falsely triggering the admin gate.
4. Robust Validation: The list comprehensions that use if isinstance(module, dict) safely prevent 500 errors if a user accidentally (or maliciously) submits malformed JSON (like passing a string instead of a dictionary for the module_config).

Rachit7168 (changes requested):
https://github.com/user-attachments/assets/286c7f82-b050-45b3-a27e-57e0d4b5b119

Not working. I still see noname attachment in the email

Rachit7168 (changes requested):
Add the Google SMTP credentials and show a working demo of the implementation.

Also, your reasoning and comments currently seem overly AI-inspired and do not properly align with the issue. Please make sure you actually test the implementation and attempt to reproduce the issue before submitting your work. Your PR description and comments should clearly demonstrate what you tested, what you observed, and how your changes address the actual problem.

Further AI-inspired reasoning, unnecessary bloating, or comments that do not reflect actual testing may lead to the PR being closed. Please keep the work focused, relevant to the issue, and backed by proper reproduction and testing.

Issue #5097, opened 2026-08-21
Group Vouchers with same tag


Sak1012:
Group vouchers with the same tag in the Voucher list, this will help assign them to plugins like exhibitions and team shifts helping automate sending vouchers to users.

Sak1012:
Also remove the Voucher Tags Tabs clicking on it just opens all vouchers tab with a filter of the tag name.
<img width="1818" height="1180" alt="Image" src="https://github.com/user-attachments/assets/8e64257d-aef4-43fb-8384-97b5fccd4cfe" />
<img width="1818" height="1180" alt="Image" src="https://github.com/user-attachments/assets/681bb34e-d778-421c-9acc-04833a29f79a" />

Rachit7168:
Closing as completed via #5099

PR #5099, opened 2026-08-21
fix(vouchers): group vouchers with tag

Sak1012:
Solves #5097

Vouchers with the same Tag grouped together and removes tabs
image
image
image

## Summary by Sourcery

Group tagged vouchers in a unified expandable list and support generating prefixed voucher codes from tags.

New Features:
- Group vouchers sharing a tag into expandable list entries with aggregate usage totals and lazy-loaded members.
- Allow bulk voucher creation to use the tag as a voucher-code prefix with validation for required tags, uniqueness, and length limits.

Bug Fixes:
- Handle missing filter forms safely when calculating advanced-filter counts.
- Use URL-safe prefix encoding when generating voucher codes.

Enhancements:
- Replace the separate voucher and tag tabs with a unified grouped voucher list and support selecting or deleting complete voucher groups.
- Improve voucher list selection behavior for grouped and dynamically loaded rows.


app/eventyay/control/views/vouchers.py
```diff
@@ -47,64 +48,55 @@ class VoucherList(PaginationMixin, EventPermissionRequiredMixin, ListView):
...
-                    t['percentage'] = int((t['redeemed'] / t['total']) * 100)
-            ctx['tags'] = tags
-        else:
-            ctx['tags'] = []
+        ctx['advanced_filter_count'] = advanced_filter_count(self.filter_form)
+        vouchers = list(ctx['vouchers'])
+        group_tags = {voucher.tag for voucher in vouchers if voucher.tag}
+        if group_tags:
+            grouped_vouchers = defaultdict(list)
+            for voucher in self.request.event.vouchers.filter(waitinglistentries__isnull=True, tag__in=group_tags):
```
ArnavBallinCode: There are three significant issues with this query:

1. **Pagination Bypass (OOM Risk):** Fetching all vouchers for a tag ignores pagination. If a bulk voucher was created with 10,000 codes, this query fetches 10,000 objects into memory at once.
2. **Missing `select_related` (N+1 Query):** When rendering `_voucher_row.html` for each member, properties like `product`, `variation`, and `seat` are accessed. This will trigger a separate database query for every single member in the group.
3. **Missing Budget Annotation:** This query is missing `Voucher.annotate_budget_used_orders(qs)`. Because of this, the `budget_used_orders` property will evaluate to empty in the template, rendering broken text like `" / $50.00"` instead of `"$10.00 / $50.00"`.

We must at least apply `Voucher.annotate_budget_used_orders` and `select_related('product', 'variation', 'seat')`. However, we also need to consider if we want to fetch all members unconditionally.

app/eventyay/control/templates/pretixcontrol/vouchers/index.html
```diff
@@ -186,146 +171,48 @@
...
+                                        </td>
+                                        <td class="list-meta"><span class="category-pill">{{ v.group_redeemed }} / {{ v.group_max_usages }}</span></td>
+                                        <td></td>
+                                        <td></td>
+                                        <td></td>
+                                        {% if request.event.has_subevents %}<td></td>{% endif %}
+                                        <td></td>
+                                    </tr>
+                                    {% for member in v.group_vouchers %}
+                                        {% include 'pretixcontrol/vouchers/_voucher_row.html' with v=member row_class='voucher-group-member hidden' group_id=v.pk %}
```
ArnavBallinCode: If a tag has thousands of vouchers, this loop will render thousands of hidden `<tr>` elements into the DOM on every page load, causing massive HTML payloads and slowing down the browser significantly.

We should either lazy-load these rows via AJAX when the group is expanded, or at least place a hard cap (e.g. max 50) on the number of members rendered inline, perhaps with a "View all..." link.

ArnavBallinCode (changes requested):
Thank you for this PR! The grouping logic looks solid, but there are a few critical performance and UI issues related to fetching and rendering the grouped voucher members that need to be addressed before we can merge this.

ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):

PR #5129, opened 2026-08-23
Fix naive datetime RuntimeWarnings in tests

Rachit7168:
Replaced naive datetimes (`datetime.date.today()`) with aware datetimes (`django.utils.timezone.now()`) in test fixtures to resolve Django RuntimeWarnings.

## Summary by Sourcery

Use timezone-aware datetimes throughout event scheduling, filtering, billing updates, and test fixtures to eliminate runtime warnings and ensure consistent date handling.

Bug Fixes:
- Eliminate Django timezone warnings and prevent naive/aware datetime mismatches across billing, reminder scheduling, event filtering, and test fixtures.

Enhancements:
- Use timezone-aware current datetimes consistently when updating billing records, calculating reminders, and querying event date ranges.
- Handle invalid reminder days gracefully when constructing dates for the current month.

Tests:
- Update event-related test fixtures to use timezone-aware datetimes.


Saksham-Sirohi (changes requested):
Django emits that warning whenever a naive datetime (or a date, which Django turns into midnight) is assigned to or compared against Event.date_from / Event.date_to while USE_TZ is on. The value 2026-08-23 00:00:00 is today’s date at midnight — exactly what you get from timezone.localdate().

That happens on live pages. The start page and event-list views filter with a date object:


startpage.py
Ln 81–116
today = timezone.localdate()
...
        .filter(Q(date_to__gte=today) | Q(date_to__isnull=True, date_from__gte=today))
The same today = timezone.localdate() filter is used for upcoming, past, and followed events. Hitting / while logged in and following an organizer, or opening those list pages, is enough to trigger both warnings.



Same class of bug, different fields (would warn, but not as Event.date_from / date_to):

eventyay_common/tasks.py: paid_datetime=datetime.now() (naive) on BillingInvoice
Same file: get_next_reminder_datetime() builds a naive datetime(year, month, day) and writes it to next_reminder_datetime



this warning does not come from tests tests are not run there

Saksham-Sirohi (approved):

PR #5132, opened 2026-08-23
Fix: Properly handle SVG uploads when creating thumbnails

### LINKED ISSUE #5131: Properly handle SVG uploads when creating thumbnails
### Description

Several forms still accept SVG uploads (speaker avatars, event/organizer logos, session images, CFP file questions). Thumbnail generation then fails because Pillow cannot decode SVG.

Production log:

```
PIL.UnidentifiedImageError: cannot identify image file <ImageFieldFile: avatars/NHSDAA_QemGgPY.svg>
ERROR ... image: Thumbnail creation failed
  File ".../eventyay/common/image.py", line 125, in create_thumbnail
    img = Image.open(image, formats=('PNG', 'JPEG', 'GIF'))
```

`create_thumbnail()` in `app/eventyay/common/image.py` only opens PNG/JPEG/GIF. The same Pillow path exists in `app/eventyay/helpers/thumb.py` (event preview / OG images). `process_image()` also calls `Image.open()` with no SVG handling.

When a thumbnail is requested, `User.get_avatar_url(thumbnail=...)` ends up with no image, and the worker logs an exception on every request. One-off converting individual files is not a fix — more users can upload SVGs through the current forms.

`IMAGE_EXTENSIONS` in `app/eventyay/common/forms/fields.py` includes `.svg`. Event/organizer logo settings also allow SVG and say those files are left unmodified.

### Expected behavior

SVG uploads keep working, and thumbnail/preview paths handle them properly instead of raising `UnidentifiedImageError`.

Possible approaches (pick one and document it):

- Rasterize SVGs with a real SVG renderer (for example resvg / librsvg / cairosvg) when a raster thumbnail is required
- Or serve the original SVG at thumbnail sizes (CSS-scaled) and skip Pillow
- Either way, `process_image()` must not rewrite SVG bytes with JPEG/PNG

Do not special-case individual files on disk.

### Acceptance criteria

- [ ] Uploading an SVG avatar (or logo) does not log `UnidentifiedImageError`
- [ ] `get_avatar_url(..., thumbnail='tiny'|'default')` still returns a usable URL for SVG avatars
- [ ] Event/organizer logo and preview paths that allow SVG do not fail thumbnail creation
- [ ] Raster images (PNG/JPEG/GIF) keep the current thumbnail behavior
- [ ] Tests cover an SVG upload through the thumbnail path

### Additional context

Relevant code:

- `app/eventyay/common/image.py` (`process_image`, `create_thumbnail`, `get_thumbnail`)
- `app/eventyay/helpers/thumb.py`
- `app/eventyay/common/forms/fields.py` (`IMAGE_EXTENSIONS`)
- `app/eventyay/base/models/auth.py` (`User.get_avatar_url`)
- `app/eventyay/helpers/image_optimize.py` (logo optimization also uses Pillow)

Rachit7168:
Fixes #5131

## Problem

Speaker avatar uploads broke in several ways when using SVG or switching between image formats:

- Pillow tried to rasterize SVG files during validation, processing, and thumbnail generation, which failed or produced broken output.
- Switching between SVG and PNG/JPEG left **stale thumbnails** and **cached speaker schedule JSON** pointing at deleted files, so the public speaker page showed a broken image while session cards could still show an old thumbnail.
- SVG files were subject to the same 10 MB limit as raster images with no content validation.
- WebP was not supported for speaker avatars despite being used elsewhere in the app.

### SVG handling
- Skip Pillow validation, `process_image`, and thumbnail generation for `.svg` avatars; serve the original scalable file via `get_avatar_url()`.
- Validate SVG uploads separately: **1 MB max** (`IMAGE_SVG_MAX_SIZE`), safe XML parsing via `defusedxml`, and rejection of scripts, event handlers, and other disallowed content.

### Format switching reliability
- `clear_avatar_thumbnails()` deletes old raster thumbnail files when the avatar is replaced or removed.
- `invalidate_speaker_avatar_caches()` clears per-speaker `eagenda:speaker:*` cache keys on avatar save.
- `get_thumbnail()` rejects thumbnails that no longer match the current avatar file.
- `get_avatar_url()` falls back to raster thumbnails if the main file is missing.

### Broader format support
- Added **WebP** to allowed avatar extensions with Pillow validation and thumbnail generation.
- Animated GIF/WebP skips destructive re-encoding but still gets thumbnails where possible.

### UI & API
- `SpeakerDetail.vue` prefers `avatar_thumbnail_*` URLs (same as session cards) before the full avatar URL.
- `ImageField.clean()` calls `validate_image()` and accepts Django 6’s `initial` argument.
- API `SpeakerUpdateSerializer` validates avatar uploads with the same rules as the orga form.

## Files changed

| Area | Changes |
|------|---------|
| `common/image.py` | SVG validation, WebP support, thumbnail/cache helpers |
| `base/models/auth.py` | Cache invalidation on save, safer `get_avatar_url()` |
| `person/forms/profile.py` | Clear thumbnails on change, SVG size help text |
| `api/serializers/speaker.py` | Avatar validation + correct user save path |
| `SpeakerDetail.vue` | Thumbnail-first avatar display |
| `helpers/image_optimize.py`, `helpers/thumb.py` | SVG bypass in shared image helpers |
| `tests/common/test_image.py` | SVG, WebP, thumbnail, and validation coverage |

## Test plan

- [ ] Upload SVG avatar in orga → save → public speaker page shows image
- [ ] Replace SVG with PNG → public page updates immediately (no broken image)
- [ ] Replace PNG with SVG → same check
- [ ] Upload WebP, PNG, JPG, JPEG, GIF → all display on speaker page and session cards
- [ ] Upload SVG > 1 MB → rejected with clear error
- [ ] Upload SVG with `<script>` → rejected
- [ ] Switch formats multiple times → no stale/broken images on speaker detail or session list

## Summary by Sourcery

Handle speaker avatar uploads safely across SVG, WebP, and raster formats while keeping thumbnails, caches, and public displays in sync.

New Features:
- Support secure SVG and WebP speaker avatar uploads alongside existing raster formats.
- Provide a maintenance command for recompressing eligible stored images and regenerating thumbnails.

Bug Fixes:
- Prevent broken or stale speaker avatars when replacing, removing, or switching between image formats.
- Avoid rasterizing SVG files while preserving reliable avatar and thumbnail display.
- Invalidate cached speaker schedule data when avatar files change.

Enhancements:
- Unify image validation across organization forms and the speaker update API.
- Improve raster image processing for animated GIF/WebP files and thumbnail format handling.
- Prefer available speaker avatar thumbnails in the speaker detail interface.

Tests:
- Add coverage for SVG security and size validation, WebP support, format switching helpers, thumbnail behavior, and image optimization.


Saksham-Sirohi (changes requested):
Reviewed in depth against SVG handling, thumbnail size, and format switching. The SVG bypass and 1 MB validation are the right direction, but raster thumbnails were not actually compressed and a few production paths would still fail.

I applied follow-up fixes locally (commit `eb8c272d83`) because maintainers can modify this branch, but could not push to the fork from this environment. Cherry-pick that commit or I can push it.

## Must-fix (already patched locally)

1. **WebP processing can crash Celery.** `process_image()` guessed `.jpg` from pixel mode, then called `Image.save(path, quality='web_high')`. Pillow infers format from the filename, so RGB `.webp` files received a JPEG quality preset and failed. Large raster avatars then stayed uncompressed.

2. **Thumbnails were not size-optimized.** `create_thumbnail()` saved with `format=img.format` and no quality/optimize flags. A 460px PNG thumbnail can stay huge. Raster thumbs should be JPEG quality 80 (PNG only when alpha is present). Default thumbs should stay well under ~200 KB; tiny thumbs under ~50 KB.

3. **`invalidate_speaker_avatar_caches()` can 500.** `SpeakerProfile` is event-scoped. Saving an avatar outside the current event scope (Celery is fine because it uses `scopes_disabled()`, web/API paths are not always) raises `ScopeError`, and even in-scope it only clears the current event. Query profiles with `scopes_disabled()`.

4. **SVG `data:` URIs were allowed.** Namespaced `href` / `xlink:href` checks missed Clark notation, and `data:text/html` payloads have no `<script>` or `javascript:` in the raw bytes. That is a stored-XSS vector if the file is opened directly.

5. **Shared SVG thumbs aliased the original file.** `helpers/thumb.py` pointed `Thumbnail.thumb` at the source path. Deleting the “thumbnail” would delete the original logo. Copy the SVG into thumb storage instead.

6. **API avatar updates could drop the file.** `UploadedFileField(source='speaker.user')` puts the upload at `validated_data['speaker']['user']`, but `update()` only popped `'avatar'`. Extract the file from both locations.

## What already looks good

- Skipping Pillow for `.svg` in validation, `process_image`, and avatar URL resolution.
- 1 MB SVG cap plus `defusedxml` parsing.
- Clearing raster thumbs and schedule JSON when the avatar format changes.
- Restoring `IMAGE_DEFAULT_MAX_WIDTH` / `HEIGHT` so raster originals can be resized.
- Speaker detail preferring `avatar_thumbnail_*` (for SVG those URLs already resolve to the original vector file).

## Residual risk

- `process_image()` still writes via `image.path`, so remote storage (S3) will not compress in place. That is pre-existing.
- Logo SVG uploads that skip `ImageField.clean()` are not size-capped the way avatars are.

## Test evidence

`pytest tests/common/test_image.py tests/test_image_optimize.py` — 27 passed, including noisy JPEG compression, WebP save, SVG bypass, and `data:` URI rejection.

app/eventyay/common/image.py
```diff
@@ -78,12 +210,21 @@ def process_image(*, image, generate_thumbnail=False):
...
         logger.exception("Failed to process image")
         return
 
+    if getattr(img, 'is_animated', False):
+        if generate_thumbnail:
+            for size in THUMBNAIL_SIZES:
+                create_thumbnail(image, size)
+        return
+
     extension = '.jpg'
```
Saksham-Sirohi: This still guesses `.jpg` from pixel mode, then later saves with `quality='web_high'` while Pillow infers format from the filename. RGB `.webp` avatars therefore get a JPEG quality preset and Celery `process_image` can crash, leaving large files uncompressed.

Save using the real file extension: JPEG `quality='web_high'` for `.jpg`, integer WebP quality for `.webp`, `optimize=True` for PNG. Do not pass JPEG presets into WebP/PNG.

app/eventyay/common/image.py
```diff
@@ -121,8 +262,11 @@ def create_thumbnail(image, size):
     if not image.instance._meta.get_field(thumbnail_field_name):
         return
 
+    if is_svg_filename(image.name):
+        return None
+
     try:
-        img = Image.open(image, formats=('PNG', 'JPEG', 'GIF'))
+        img = Image.open(image, formats=RASTER_THUMBNAIL_FORMATS)
```
Saksham-Sirohi: `create_thumbnail()` still saves with `format=img.format` and no quality/optimize flags (below this hunk). That does not produce compressed thumbnail sizes.

Encode RGB thumbs as JPEG quality ~80 (progressive, optimize). Keep PNG only when the image has alpha. Please add a test that a noisy ~2000px JPEG yields a tiny thumb well under 50 KB and a default thumb well under 200 KB.

app/eventyay/common/image.py
```diff
@@ -1,23 +1,90 @@
...
+def invalidate_speaker_avatar_caches(user):
+    """Drop cached speaker schedule JSON after avatar URLs may have changed."""
+    if not user.code:
+        return
+
+    from eventyay.agenda.views.utils import clear_schedule_caches
+    from eventyay.base.models import Event
+    from eventyay.base.models.profile import SpeakerProfile
+
+    event_ids = SpeakerProfile.objects.filter(user_id=user.pk).values_list('event_id', flat=True)
```
Saksham-Sirohi: `SpeakerProfile` uses `ScopedManager(event='event')`. This query can raise `ScopeError` when the avatar is saved outside the current event scope, and even in-scope it only clears one event.

Load profile/event IDs inside `scopes_disabled()` so cache invalidation cannot 500 and so a speaker on multiple events is updated.

app/eventyay/common/image.py
```diff
@@ -27,10 +94,75 @@
...
+        )
+
+    for element in root.iter():
+        if _svg_local_tag(element.tag) in FORBIDDEN_SVG_TAGS:
+            raise ValidationError(_('This SVG file contains disallowed content and cannot be uploaded.'))
+        for attr_name, attr_value in element.attrib.items():
+            attr_name_lower = attr_name.lower()
+            if attr_name_lower.startswith('on'):
+                raise ValidationError(_('This SVG file contains disallowed content and cannot be uploaded.'))
+            if attr_name_lower in {'href', 'xlink:href'} and re.match(
```
Saksham-Sirohi: This misses namespaced attributes (`{http://www.w3.org/1999/xlink}href`) and `data:` URLs. A `data:text/html` href has no raw `<script>` or `javascript:` bytes, so it bypasses the earlier content scan.

Compare the local attribute name (after stripping the `{ns}` prefix) and reject both `javascript:` and `data:` on `href`/`src`/`xlink:href`. Add a test for a `data:` URI SVG.

app/eventyay/helpers/thumb.py
```diff
@@ -60,6 +60,12 @@ def get_sizes(size, imgsize):
 
 
 def create_thumbnail(sourcename, size):
+    if str(sourcename).lower().endswith('.svg'):
+        t = Thumbnail.objects.create(source=sourcename, size=size)
+        t.thumb.name = sourcename
```
Saksham-Sirohi: Aliasing `thumb` to the source path means deleting the thumbnail record deletes the original SVG logo.

Copy the SVG bytes into thumb storage (same pattern as raster thumbs) so the original file stays independent.

app/eventyay/api/serializers/speaker.py
```diff
@@ -220,6 +222,18 @@ class Meta(SpeakerSerializer.Meta):
...
+        if extension not in ALLOWED_IMAGE_EXTENSIONS:
+            allowed_formats = ', '.join(sorted(ALLOWED_IMAGE_EXTENSIONS))
+            raise exceptions.ValidationError(
+                f"The file type '{extension}' is not supported. Please upload one of: {allowed_formats}."
+            )
+        validate_image(avatar)
+        return avatar
+
     def update(self, instance, validated_data):
         avatar = validated_data.pop('avatar', None)
```
Saksham-Sirohi, line 243: `avatar` is declared with `source='speaker.user'`, so DRF puts the upload at `validated_data['speaker']['user']`, not `'avatar'`. This `pop('avatar')` can be `None` and the file is never saved.

Read the uploaded file from `avatar`, nested `speaker.user`, and `user.avatar` before saving onto `instance.user.avatar`.

app/tests/common/test_image.py
```diff
@@ -0,0 +1,134 @@
...
+    def save(self, *args, **kwargs):
+        pass
+
+class DummyImage:
+    def __init__(self, name):
+        self.name = name
+        self.instance = DummyInstance()
+        self.field = DummyField('avatar')
+
+def test_process_image_svg():
```
Saksham-Sirohi, line 102: Please cover the production gaps, not only Dummy SVG short-circuits:
- noisy JPEG is resized/compressed and thumbs stay small
- WebP `process_image` does not raise
- SVG with a `data:` href is rejected
- JPEG thumbnails may use `.jpg` even when the avatar is `.png` (`thumbnail_matches_avatar` currently requires the same suffix)

Saksham-Sirohi (changes requested):
Please address these on this PR. Do not rely on a maintainer-side local patch.

SVG bypass and the 1 MB cap are the right direction, but raster thumbnails are still not compressed, WebP processing can crash Celery, and a few save/cache paths are unsafe. Inline comments have the required changes.

prayag78 (approved):

app/eventyay/common/image.py
```diff
@@ -100,7 +260,15 @@ def process_image(*, image, generate_thumbnail=False):
...
-    img_without_exif.save(image.path, quality='web_high' if extension == '.jpg' else 95)
+    save_kwargs = {}
+    if extension in ('.jpg', '.jpeg'):
+        save_kwargs = {'quality': 'web_high'}
+    elif extension == '.webp':
+        save_kwargs = {'quality': 95}
+    elif extension == '.png':
+        save_kwargs = {'optimize': True}
+
+    img_without_exif.save(image.path, **save_kwargs)
```
Saksham-Sirohi: Pillow still infers format from `image.path`, not from `extension`. GIF is an allowed avatar type; a non-animated `.gif` is remapped to `.jpg` kwargs here, then saved to a `.gif` path with `quality='web_high'`. That is the same class of bug as the WebP crash.

Please either:
- skip rewriting the original when the suffix is not `.jpg`/`.jpeg`/`.png`/`.webp` (still generate thumbs), or
- pass an explicit `format=` that matches the file on disk.

Also pass `format='JPEG'|'PNG'|'WEBP'` for the matching suffixes so quality kwargs cannot leak into the wrong encoder.

app/eventyay/common/image.py
```diff
@@ -121,32 +289,97 @@ def create_thumbnail(image, size):
...
+    except Exception:
+        new_size = None
+
+    logger.info(
+        'Recompressed image %s (%s -> %s bytes)',
+        image.name,
+        size,
+        new_size if new_size is not None else 'unknown',
+    )
+    return True
```
Saksham-Sirohi, line 384: `process_image()` swallows open/save failures and returns `None`, so this `except` never runs and the backfill reports success even when nothing was compressed.

`process_image` should return a boolean (or raise), and this helper should return that result. Right now `backfill_images` can print `Compressed ...` for files that were skipped internally.

Saksham-Sirohi (changes requested):
Thanks for addressing the previous review. The original six items are fixed:

- WebP uses integer quality instead of JPEG presets
- Raster thumbs encode as JPEG quality 80 (PNG only with alpha), with size tests
- Cache invalidation queries SpeakerProfile under `scopes_disabled()`
- SVG `data:` / namespaced hrefs are rejected
- Shared SVG thumbs copy bytes instead of aliasing the original
- API update reads the nested `speaker.user` / `user.avatar` upload

Two leftovers still need a follow-up on this PR before merge.

app/eventyay/common/image.py
Saksham-Sirohi: Fixed in later commits: save kwargs now follow the file suffix, and WebP uses integer quality. Remaining related issue is the GIF/filename format mismatch: https://github.com/fossasia/eventyay/pull/5132#discussion_r3838328253

app/eventyay/common/image.py
Saksham-Sirohi: Fixed. Raster thumbs now encode as JPEG quality 80 (PNG only when there is alpha), and tests cover size bounds.

app/eventyay/common/image.py
Saksham-Sirohi: Fixed. SpeakerProfile/Event lookups now run inside `scopes_disabled()`.

app/eventyay/common/image.py
Saksham-Sirohi: Fixed. Local attribute names are used, and both `javascript:` and `data:` hrefs are rejected, with a `data:` URI test.

app/eventyay/helpers/thumb.py
Saksham-Sirohi: Fixed. SVG thumbs now copy bytes into thumb storage instead of aliasing the original path.

app/eventyay/api/serializers/speaker.py
Saksham-Sirohi: Fixed. `update()` now reads the upload from nested `speaker.user` and `user.avatar` as well as `avatar`.

app/tests/common/test_image.py
Saksham-Sirohi: Fixed. Tests now cover noisy JPEG compression/thumb sizes, WebP process_image, data: SVG rejection, and JPEG thumbs for PNG avatars.

app/eventyay/common/image.py
```diff
@@ -98,13 +231,66 @@ def process_image(*, image, generate_thumbnail=False):
...
+    """
+    This function receives an image that has been uploaded, and processes it
+    by reducing its file size and stripping its metadata.
+    Image must be an ImageFieldFile, e.g. user.avatar.
+
+    Returns True when processing completed (including intentional skips such as
+    animated GIF originals), False when the image could not be processed.
+    """
+    if is_svg_filename(image.name):
+        return False
```
Saksham-Sirohi: This is an intentional skip, same as animated GIF and other non-rewritable originals below, but it returns `False`.

`task_process_image` treats False as failure and logs `Could not process image %s`. Every successful SVG avatar upload will therefore error in Celery.

Please `return True` here (and assert that in `test_process_image_svg`).

Saksham-Sirohi (changes requested):
The two leftovers from the last review are fixed:

- GIF originals are no longer rewritten; JPEG/PNG/WebP saves pass an explicit `format=`
- `process_image()` returns a boolean, and `recompress_image_field()` / Celery use it

One leftover on this SVG-focused PR: SVG is an intentional skip but returns `False`, so Celery will log `Could not process image` on every SVG avatar upload. Please return `True` there, matching animated GIF / non-rewritable originals, and update `test_process_image_svg`.

app/eventyay/common/image.py
Saksham-Sirohi: Fixed in 61e03f2. Non-rewritable originals (including GIF) skip the rewrite and still generate thumbs; JPEG/PNG/WebP saves pass an explicit `format=`.

app/eventyay/common/image.py
Saksham-Sirohi: Fixed in 61e03f2. `process_image()` now returns a boolean, and `recompress_image_field()` returns False when processing fails.

Saksham-Sirohi (approved):
The last leftover is fixed: `process_image()` now returns `True` for SVG (intentional skip), matching GIF/other non-rewritable originals, so Celery will not log a failure after a successful SVG upload.

SVG bypass, compressed raster thumbs, format switching, and the backfill boolean path all check out.

app/eventyay/common/image.py
Saksham-Sirohi: Fixed in a8cbb44. SVG now returns `True` as an intentional skip, and `test_process_image_svg` asserts that.

ArnavBallinCode (commented):
### Potential data-loss window when replacing avatar

`clear_avatar_thumbnails()` is called before the new avatar is saved and processed:

```python
clear_avatar_thumbnails(instance.user)
instance.user.avatar.save(..., save=True)
instance.user.process_image(...)
```

If the avatar save or image processing fails after the thumbnails have been deleted, we lose the existing thumbnails before we've established that the replacement succeeded.

Could we reorder this so the new avatar is successfully saved/processed first, and only then remove/regenerate the old thumbnails? Ideally the replacement should leave the existing avatar state intact if any step of the new upload fails.

MukundC25 (approved):

ArnavBallinCode (approved):

PR #5133, opened 2026-08-23
fix: keep schedule API up when Redis blips and quiet host-header logs

Saksham-Sirohi:
## Summary
- Fall back to the event `locale_array` when Redis times out while reading settings, so orga schedule API requests no longer 500.
- Allow `data:` in CSP `font-src` to stop schedule-rooms font violations.
- Log unknown `Host` headers as a one-line WARNING (no traceback) so scanner probes stay visible without flooding error logs.

## Test plan
- [x] Unit tests added for locale Redis fallback and CSP `font-src`
- [x] Verified `OneLineWarningFilter` loads from Django logging config and emits one WARNING line with no traceback
- [ ] Confirm orga schedule talk API still works when Redis is healthy
- [ ] Confirm orga schedule talk API returns 200 (not 500) if Redis briefly times out
- [ ] Confirm a request with a forged Host header is still 400 and logs one WARNING line
- [ ] Confirm schedule rooms page no longer reports `font-src` / `data` CSP violations

## Summary by Sourcery

Keep schedule-related requests resilient during Redis outages and reduce noise from invalid Host-header probes.

Bug Fixes:
- Keep event locale resolution available when Redis settings reads fail by falling back to the event’s stored locale configuration.
- Allow video asset requests to bypass load shedding so schedule resources remain available during overload.
- Permit data-backed fonts through the Content Security Policy.
- Reduce unknown Host-header security logs to a single warning without traceback.

Tests:
- Add coverage for Redis locale fallback, data-font CSP support, and video asset load-shedding exemptions.

### DIFF
--- a/app/eventyay/base/middleware.py
+++ b/app/eventyay/base/middleware.py
@@ -482,7 +482,7 @@ def request_prefers_json_api(request):
 
 
 def is_load_shed_exempt(path):
-    if path.startswith('/healthcheck'):
+    if path.startswith('/healthcheck') or '/video/assets/' in path:
         return True
     return bool(CHECKIN_EXEMPT_RE.search(path))
 
--- a/app/eventyay/base/middleware.py
+++ b/app/eventyay/base/middleware.py
@@ -343,6 +343,7 @@ class SecurityMiddleware(MiddlewareMixin):
             + img_src,
             'font-src': [
                 '{static}',
+                'data:',
                 'https://fonts.gstatic.com',  # fix Google Fonts
                 *vite_http,
             ],
--- a/app/eventyay/helpers/security.py
+++ b/app/eventyay/helpers/security.py
@@ -1,4 +1,5 @@
 import hashlib
+import logging
 import time
 
 from django.conf import settings
@@ -49,3 +50,13 @@ def assert_session_valid(request: HttpRequest) -> bool:
 
     request.session[KEY_LAST_LOGIN_CHECK] = int(time.time())
     return True
+
+
+class OneLineWarningFilter(logging.Filter):
+    def filter(self, record: logging.LogRecord) -> bool:
+        record.exc_info = None
+        record.exc_text = None
+        record.levelno = logging.WARNING
+        record.levelname = 'WARNING'
+        return True
+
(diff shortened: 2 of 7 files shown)


Rachit7168 (approved):
LGTM , Tested 

Tested locally with Docker: schedule/orga pages OK with Redis up; simulated Redis pause — no 500, warning logged, locales fall back to DB. Schedule rooms: no CSP font-src violations. Forged Host → 400 + single WARNING, no traceback. Unit tests for locale fallback and CSP pass.

app/eventyay/helpers/security.py
```diff
@@ -49,3 +50,13 @@ def assert_session_valid(request: HttpRequest) -> bool:
...
     return True
+
+
+class OneLineWarningFilter(logging.Filter):
+    def filter(self, record: logging.LogRecord) -> bool:
+        record.exc_info = None
+        record.exc_text = None
+        record.levelno = logging.WARNING
+        record.levelname = 'WARNING'
+        return True
```
ViRUS-0-0, line 61: [`OneLineWarningFilter`]  strips `exc_info` and `exc_text`, but does not reset `stack_info`. If a log event is emitted with `stack_info=True`, Python's `logging.Formatter` will still output a multi-line stack trace.

Consider explicitly resetitng `record.stack_info = None` to guarantee strict single-line output across all formatting configurations.

ViRUS-0-0 (changes requested):

app/eventyay/base/middleware.py
```diff
@@ -481,7 +482,7 @@ def request_prefers_json_api(request):
 
 
 def is_load_shed_exempt(path):
-    if path.startswith('/healthcheck'):
+    if path.startswith('/healthcheck') or '/video/assets/' in path:
         return True
     return bool(CHECKIN_EXEMPT_RE.search(path))
```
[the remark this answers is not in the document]
Saksham-Sirohi: The exemption is only meant for video UI static files under /{organizer}/{event}/video/assets/…. Load shedding is a concurrency cap, not an auth check, so a crafted URL that happens to contain that substring would not grant extra privileges. This is pretty much useless to implement, as the majority is ui assets and other is api which is already auth gated

app/eventyay/helpers/security.py
Saksham-Sirohi: @ViRUS-0-0 Django logs DisallowedHost with exc_info, not stack_info=True, and nothing in this codebase uses stack_info. Clearing exc_info/exc_text is what stops the scanner traces; resetting stack_info would not change what we actually log. Leaving it as-is.

ViRUS-0-0 (approved):
LGTM

ArnavBallinCode (approved):

PR #5172, opened 2026-08-24
feat: hide exhibition permissions when plugin not enabled

### LINKED ISSUE #166: Hide the "Exhibition permissions" section when the exhibition plugin is disabled
### Problem

The organizer team edit form always renders an "Exhibition permissions" fieldset with three checkboxes — Reviewer Manager, Exhibitor Reviewer, and Hide emails of applicants. These appear for every team of every organizer, including organizers whose events do not have the exhibition plugin enabled and installations where the plugin is not installed at all.

The result is that organizers are asked to make decisions about permissions that grant access to nothing. There is no exhibition area for those events, so the checkboxes are unreachable configuration that only adds noise to an already long permissions form.

### Current behavior

eventyay_common/organizers/teams/_permissions_fieldsets.html renders the fieldset with no condition:

<fieldset>
    <legend>{% translate "Exhibition permissions" %}</legend>
    {% bootstrap_field form.can_change_exhibition_proposals layout="control" %}
    {% bootstrap_field form.is_exhibition_reviewer layout="control" %}
    {% bootstrap_field form.hide_exhibition_applicant_emails layout="control" %}
</fieldset>

Why the per-event checkbox list doesn't complicate this

A team is scoped to a single organizer and may apply to all of that organizer's events or a subset of them, so there is no single event to check the plugin against. That question is already settled elsewhere in this same form: the TeamShifts permissions resolve it at organizer level — the section is shown if the plugin is installed and enabled on at least one event belonging to the organizer. If no event under the organizer uses the plugin, the permissions cannot apply to anything and are hidden outright. The same reasoning holds for exhibition, so no per-event permission granularity is needed.

### Suggested approach

1. Determine availability once:
has_exhibition = apps.is_installed('exhibition') and organizer.events.filter(
    plugins__regex=r'(^|,)exhibition(,|$)'
).exists()
self.exhibition_plugin_enabled = has_exhibition
   (exhibition is the plugin's app label, per its AppConfig.name.)
2. When not available, delete can_change_exhibition_proposals, is_exhibition_reviewer and hide_exhibition_applicant_emails from self.fields, mirroring how the teamshifts_fields list is removed.
3. Guard the fieldset in the template so the legend disappears with its contents:
{% if form.exhibition_plugin_enabled %}
<fieldset>
    <legend>{% translate "Exhibition permissions" %}</legend>
    ...
</fieldset>
{% endif %}

Notes

- Removing a field from the form rather than merely hiding it means existing values are left untouched on save, so a team that already has a permission set does not silently lose it if the plugin is later disabled.
- This is a UI-visibility change only. The model fields and the API serializers (api/serializers/team.py, api/serializers/organizer.py) are intentionally left as they are, so stored data and API contracts are unaffected.
- The "Video permissions" fieldset immediately below has the same unguarded structure. Worth confirming whether it should get the same treatment, though that is out of scope here.

### Acceptance criteria

- An organizer with no exhibition-enabled event sees no "Exhibition permissions" section, and no empty legend, on the team create and edit forms.
- An organizer with at least one exhibition-enabled event sees the section unchanged.
- Disabling the plugin does not alter permission values already stored on existing teams.

Aqil-Ahmad:
fixes https://github.com/fossasia/eventyay-exhibition/issues/166

Permissions are not shown when the exhibition plugins is not enabled for any of the organizer events. Once any event having plugin enabled is checked on team creation/edit form then the permissions will appear progressively/dynamically.

https://github.com/user-attachments/assets/f145a56a-b163-4ba6-abe4-e97d6c7e9aeb

## Summary by Sourcery

Restrict exhibition team permissions to teams that cover events with the exhibition plugin enabled.

Bug Fixes:
- Hide exhibition-specific team permissions when no organizer event has the exhibition plugin enabled.
- Prevent exhibition-only team permissions from being accepted when the selected events do not support the exhibition plugin.

Enhancements:
- Dynamically show or clear exhibition permissions based on the team’s selected events while preserving previously stored permission values when the fields are unavailable.

Tests:
- Add coverage for exhibition permission visibility, event selection validation, preservation of stored permissions, and interaction with other team permissions.

## Summary by Sourcery

Restrict exhibition team permissions to events where the exhibition plugin is enabled.

Bug Fixes:
- Hide exhibition-specific team permissions when no selected organizer event has the exhibition plugin enabled.
- Reject exhibition-only permissions when a team covers events without exhibition support.

Enhancements:
- Dynamically show or clear exhibition permissions as team event selections change while preserving stored permissions when the fields are unavailable.

Tests:
- Add coverage for exhibition permission visibility, event selection validation, stored permission preservation, and interaction with other team permissions.


ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):

PR #5175, opened 2026-08-24
feat: smart LANCZOS-based image optimizations

ArnavBallinCode:
This PR introduces a comprehensive image optimization pipeline for user uploads. It ensures that large image uploads are downscaled and re-encoded to optimal formats (WebP, PNG, or JPEG) to reduce bandwidth and storage usage while maintaining visual quality.

### Changes
*   **Centralized Encoder**: Added `encode_optimized` in `common/image.py` to standardize processing.
*   **Inline Optimization**: `UploadedFileField` now natively optimizes images for standard models (`picture`, `logo_image`, etc.).
*   **Settings Optimization**: `SettingsForm` now applies this optimization automatically to event and organizer configuration images.
*   **Backfill Command**: Refactored `backfill_images` to act as the single source of truth for retroactively optimizing all historical images, supporting dry-runs, precise model targeting (`--model`), and robust `MEDIA_ROOT`-safe backups (`--backup-dir`).
*   **Atomic Storage Updates**: Improved the local and remote file replacement logic in `process_image` to be safe and atomic.

### Testing
*   Added `max_size` enforcement to `RoomSerializer.picture`.
*   Tested end-to-end local generation, storage integration, and backfill edge-cases.
*   All Copilot code review comments resolved.

## Summary by Sourcery

Implement a centralized, safe image optimization pipeline for new uploads and historical media while adding room image support and interpretation-service synchronization.

New Features:
- Optimize uploaded event, organizer, room, product, and related images automatically using dimension-aware encoding and format selection.
- Add room picture uploads with validation and configured size limits.
- Add a room-admin action for synchronizing interpretation services.
- Extend the image backfill command to cover additional models and settings-backed images with model selection and dry-run support.

Bug Fixes:
- Prevent unsafe image processing by rejecting corrupted, oversized, or decompression-bomb uploads.
- Make image replacement safer across local and remote storage backends and avoid unnecessary output growth.

Enhancements:
- Centralize image resizing, metadata removal, and optimized encoding in the common image-processing pipeline.
- Require backups outside MEDIA_ROOT for non-dry-run image backfills and update image-size guidance to the new limits.

Tests:
- Update room API expectations to include the picture field.

## Summary by CodeRabbit

- **New Features**
  - Room images can be uploaded in PNG, JPEG, GIF, and WebP formats.
  - Administrators can sync interpretation services from the room editor.
  - Room creation supports stream scheduling options.

- **Improvements**
  - Uploaded images are automatically optimized, with a maximum width of 1920 pixels where applicable.
  - Optimization now covers additional logos, headers, previews, profile images, and Open Graph images.
  - Image backfill tools now support rooms, products, and settings with required backups.

- **Bug Fixes**
  - Improved handling of invalid images, optimization failures, and storage updates.


app/eventyay/common/image.py
```diff
@@ -281,9 +281,30 @@ def process_image(*, image, generate_thumbnail=False):
...
+        dir_name = os.path.dirname(image.path)
+        temp_path = None
+        try:
+            fd, temp_path = tempfile.mkstemp(dir=dir_name, suffix=extension)
+            with os.fdopen(fd, 'wb') as temp_file:
+                prepared.save(temp_file, format=save_format, **save_kwargs)
+            
+            os.chmod(temp_path, os.stat(image.path).st_mode)
+            # Atomically replace the old file with the new one
+            os.replace(temp_path, image.path)
```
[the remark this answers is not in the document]
ArnavBallinCode: we dont use S3 , storage buckets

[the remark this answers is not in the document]
ArnavBallinCode: we use a local storage

Saksham-Sirohi (changes requested):
## Review

Backups are a good addition. Two things still block a production run.

### Must fix

1. **The bulk command does not touch the images on event cards / event details.** Those live in the settings store (`logo_image`, `event_logo_image`, organizer header/logo), not `Event.logo` / `Event.header_image`. As written, `optimize_all_images` will not shrink the public headers this PR is aiming at. Process those settings files (or drop the command and only ship the upload-time quality/size changes).

2. **Do not write backups into `MEDIA_ROOT/image_backups/`.**  nginx serves `MEDIA_ROOT`. Originals would be public at predictable URLs, and they stay there until someone remembers to delete them. Write backups outside the served media tree (or refuse to run unless `--backup-dir` is set to a non-public path).

3. **Skip files that are already small.** There is no `--min-size-kb` (see existing `backfill_images`, default 500). Without it this re-encodes every avatar/product image, including ones already optimized, and a second run stacks more JPEG/WebP loss. Only rewrite files over a threshold.

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/84e8849f-3b02-4a84-b615-1b8e6d35a596

Saksham-Sirohi (changes requested):
## Test report

I ran the current head (`00d4d06`) against local JPEG/PNG/WebP/GIF/SVG uploads, the remote-storage fallback, and `optimize_all_images` (backup → overwrite, dry-run, re-run). `--backup-dir` is a real improvement. This is still not safe to run as a production backfill, and it is still not one pipeline.

### What works

- Local raster rewrite: JPEG/JPEG/WebP are compressed in place with LANCZOS + quality 70/75, SVG and animated GIF are left untouched, PNG alpha is kept.
- Backup-then-overwrite on local disk: with `--backup-dir` set, the command copies the original first, then `process_image` atomically replaces the live file. If `os.replace` fails, the live file stays intact and the backup still has the original bytes.
- Dry-run does not write backups or live files.
- `--backup-dir` is required unless `--dry-run` (no longer dumps originals into public `MEDIA_ROOT`).

### Must fix (data loss / wrong target)

1. **Remote fallback can delete the only copy.** `process_image` still does `storage.delete(image.name)` then `storage.save(image.name, ...)`. If that save fails, the `finally` block also deletes the temp key. I reproduced this: original gone, temp gone, `process_image` returns `False`. If `Storage.save()` rewrites the name (NanoCDN / `get_available_name`), the field still points at the deleted key. Do not delete first. Upload to a temp key, copy over the original key only after success, and never delete the temp until the original key has the new bytes.

2. **Re-running the command destroys the original backup.** Backup names are `{Model}_{pk}_{basename}` with no timestamp. A second run copies the already-compressed file over the backup of the true original, then compresses again (generational JPEG/WebP loss). There is still no `--min-size-kb`. Use unique backup names and skip files below a size threshold (existing `backfill_images` default is 500 KB).

3. **The backfill still misses the images on event cards / event details.** Those live in the settings store (`logo_image`, `event_logo_image`, `event_preview_image`, `og_image`, organizer logo/header), not `Event.logo` / `Event.header_image`. `optimize_all_images` never walks those keys. A production run of this command will not shrink the public headers this PR is aiming at.

4. **Not all uploads go through compression.** This is still two compressors (`process_image` vs `optimize_uploaded_image`), and several upload paths skip both:
   - Product pictures via the API serializer
   - `Event.logo` / `Event.header_image` model fields (orga wizard; no `process_image` call)
   - `Room.picture`
   - `invoice_logo_image`, `startpage_header_image`
   - pretix `EventSettingsForm` (has the image fields, does not call `optimize_uploaded_image`; only `EventCommonSettingsForm` does)
   - The command itself **skips remote storage** even though `process_image` now has a (unsafe) remote fallback
   - Existing `recompress_image_field` / `backfill_images` still crash on `getattr(image, 'path')` for remote backends (`NotImplementedError`)

### Formats (not flawless)

| Format | `process_image` (avatars/submissions) | `optimize_uploaded_image` (event/organizer/product UI) |
|---|---|---|
| JPEG | rewrite q70, shrinks | rewrite q70, shrinks |
| WebP | stays WebP q75 | converted to JPEG |
| PNG | stays PNG; can **grow** (observed 9.7KB → 23KB after `putdata` + re-encode) | stays PNG if alpha, else JPEG |
| GIF static | **not rewritten** | converted to JPEG |
| GIF animated | left as-is | left as-is |
| SVG | left as-is | left as-is |

Pillow 14 also deprecates `Image.getdata()` used in `_prepare_original_image`.

### Also

- `transaction.atomic()` does not roll back filesystem/storage writes. It only opens a DB transaction around a file rewrite (and makes the command fail in any context without a DB connection).
- Unused `as e` in the `process_image` except clause; `handle()` still reads `options` instead of declared parameters.
- Help text in `default_setting.py` still says “max 3000 px wide” while `MAX_WIDTH['logo_image']` is now 1920.

Please keep the quality/dimension tweaks if you want, but do not ship this bulk rewriter until backups cannot be clobbered, remote replace cannot drop the file, settings-store images are in scope (or the command is dropped in favor of extending `backfill_images`), and there is one upload-time path that actually covers the formats we accept.

Saksham-Sirohi:
Requesting changes on the remaining safety and coverage gaps. Please address these before this is runnable as a production backfill.

### 1. No data loss

**Remote storage:** `process_image` still deletes the original, then saves, then deletes the temp key in `finally`. If the final `storage.save()` fails, both copies are gone. If `Storage.save()` returns a rewritten name (`get_available_name` / NanoCDN), the field still points at a deleted key.

Do not delete first. Upload to a temp key, write the compressed bytes onto the original key only after that succeeds, and only then delete the temp.

**Backups:** `{Model}_{pk}_{basename}` is overwritten on every run. A second run backs up the already-compressed file over the true original, then compresses again. Use a unique backup name (timestamp or content hash), refuse to overwrite an existing original backup, and add `--min-size-kb` (see existing `backfill_images`, default 500) so already-small files are not re-encoded.

`transaction.atomic()` does not roll back filesystem/storage writes. It is not a safety net here.

### 2. Backup then overwrite, as a backfill

Local JPEG backup-then-overwrite works when `--backup-dir` is set. That is not enough:

- The command only walks `Event.logo`, `Event.header_image`, `User.avatar`, `Product.picture`, `Submission.image`, `Room.picture`.
- Public event cards / event details use the **settings store** (`logo_image`, `event_logo_image`, `event_preview_image`, `og_image`, organizer logo/header). Those are never touched.
- Remote backends are skipped entirely.

Either process the settings-store files (the actual public headers) or drop this command and extend `backfill_images`.

### 3. Every upload must go through one compression route

There are still two compressors (`process_image` vs `optimize_uploaded_image`), and several upload paths hit neither:

- Product pictures via the API
- `Event.logo` / `Event.header_image` from the orga wizard
- `Room.picture`
- `invoice_logo_image`, `startpage_header_image`
- pretix `EventSettingsForm` (the common event settings form compresses; this one does not)

New avatars / submission images go through `process_image`. Event / organizer / product UI uploads go through `optimize_uploaded_image`. Please make this one pipeline so every accepted upload is compressed the same way (or explicitly skipped for the same documented reasons).

### 4. Formats, including animated GIFs

We **do allow GIFs** (`.gif` is in `IMAGE_EXTENSIONS` and in the event/organizer/product/avatar whitelists). Animated GIFs must **not** be recompressed — flattening them would destroy the animation. Both pipelines currently bypass animated originals, which is the correct behavior; please keep that, add a test that the original animated bytes are unchanged after upload and after backfill, and still generate a still thumbnail where we already do for avatars.

What is not consistent today:

| Format | `process_image` | `optimize_uploaded_image` |
|---|---|---|
| JPEG | rewrite q70 | rewrite q70 |
| WebP | stays WebP q75 | converted to JPEG |
| PNG | stays PNG (can grow) | PNG if alpha, else JPEG |
| GIF static | not rewritten | converted to JPEG |
| GIF animated | left as-is (correct) | left as-is (correct) |
| SVG | left as-is (correct) | left as-is (correct) |

Please pick one policy per format, apply it in a single route, and cover JPEG, PNG, WebP, SVG, static GIF, and animated GIF in tests — including the backup-then-overwrite path and a failed-save path that must leave the original intact.

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/84ebee9d-6127-4f24-b634-f2cb3f876fa3

https://github.com/user-attachments/assets/4e70082e-c1fb-456c-966e-2b7aaceb4d77

app/eventyay/base/forms/__init__.py
```diff
@@ -143,6 +146,25 @@ def get_initial_setting_value(self, settings_proxy, key, declared_type, default=
                 return value
         return settings_proxy.get(key, as_type=declared_type, default=default)
 
+    def clean(self):
```
ViRUS-0-0, line 148: If an organizer uploads an invalid image or an oversized decompression bomb, the form silently ignores the error and saves the broken file.

Catch the exception and register a field validation error using `self.add_error()`.

app/eventyay/common/management/commands/backfill_images.py
```diff
@@ -5,27 +5,60 @@
...
+    def open(self, mode='rb'):
+        return self.storage.open(self.name, mode)
+
+    @property
+    def size(self):
+        return self.storage.size(self.name)
+
+    @property
+    def path(self):
+        return self.storage.path(self.name)
```
ViRUS-0-0, line 72: When running `backfill_images` on remote storage, if storage renames the file on upload, the settings database table is not updated and still points to the old filename.

Attach the settings store row to `MockImageFieldFile` so any renamed file path is saved back to `store.value`.

app/eventyay/webapp/video/src/views/admin/rooms/EditForm.vue
```diff
@@ -14,6 +14,7 @@
 			sidebar-addons(v-if="inferredType && inferredType.id === 'stage'", :config="config", :modules="modules", :creating="creating")
 	.ui-form-actions
 		bunt-button.btn-save(@click="save", :loading="saving", :error="!!error") {{ creating ? $t('Create') : $t('Save') }}
+		bunt-button.btn-sync(v-if="!creating && interpretationAdmin.usePluginStreams", @click="syncServices", :loading="syncing", :error="!!syncError", style="margin-left: 10px") {{ $t('Sync Services') }}
```
ViRUS-0-0, line 55: The sync button in video admin enters an error state when it fails, but the actual error message never gets rendered to the user.

Include `syncError` in the `.errors` container.

ViRUS-0-0 (changes requested):

prayag78 (commented):
Uploaded images might not be getting saved:

https://github.com/user-attachments/assets/ac22c7f0-259d-4790-9889-5689844fdc40

ViRUS-0-0 (approved):

prayag78 (approved):

app/eventyay/common/image.py
```diff
@@ -281,10 +325,65 @@ def process_image(*, image, generate_thumbnail=False):
...
+                # Fallback for remote storage backends (e.g., S3)
+                original_name = image.name
+                temp_key = f'{original_name}.tmp{extension}'
+                
+                # 1. Upload to temp key first
+                saved_temp_key = image.storage.save(temp_key, ContentFile(buf.getvalue()))
+                
+                try:
+                    # 2. Upload succeeded, replace file contents
+                    image.storage.delete(original_name)
```
Saksham-Sirohi, line 376: The temp-key upload is an improvement, but this still deletes `original_name` before the replacement `save` succeeds, and `finally` deletes the temp key too. If that second `save` fails (or returns a different name), both copies are gone. Do not delete the live key until the new bytes are stored under it (copy/overwrite from the temp key, then delete temp).

Saksham-Sirohi (changes requested):
Backups, `--min-size-kb`, and settings-store images look addressed. One leftover from the last review still drops the live file on remote storage.

app/eventyay/common/image.py
ArnavBallinCode: we dont use remote storage so this is okay for now and no one self hosting on remote storage is going to use this backup command. So I think this is okay

Saksham-Sirohi: Local-only today does not make this safe. `delete(original_name)` still runs before the replacement `save`, and `finally` still drops the temp key. If that second `save` fails (disk full, permissions, backend returns a different name), both copies are gone.

Keep the live key until the new bytes are stored under it, then delete temp. That is independent of whether production uses S3 right now.

Saksham-Sirohi (approved):
The remote path no longer deletes the live key before the replacement `save`. Local uses `os.replace`; remote writes first, then updates the name, then deletes the old key. Settings-store images still go through `MockImageFieldFile` so the new path is persisted.

PR #5177, opened 2026-08-24
fix(video): stop force-muting YouTube room streams on autoplay

### LINKED ISSUE #4558: YouTube room stream starts muted unexpectedly
## Observed behavior
 
While reviewing stream initialization for YouTube-embedded room streams, I found that the stream is force-muted on autoplay whenever the room's controls are visible which is the default, most common room setup. The current mute condition in `MediaSource.vue` only allows the stream to start unmuted when `hideControls` is enabled; as soon as controls are visible, autoplay always mutes the stream regardless of any other setting.
 
In practice this means viewers opening a typical room (controls visible, autoplay on) get a silent YouTube stream and have to manually unmute every time, even when there's no reason the stream shouldn't just play with sound.
 
## Why it matters
 
Forcing mute in the common configuration defeats the purpose of autoplay for streams that are meant to be heard organizers and viewers expect the stream to play with audio in the default setup, not to require a manual unmute click on every room visit.
 
## Expected behavior
 
- YouTube streams should not be force-muted just because controls are visible.
- Mute should only be applied where it's actually needed (e.g., to satisfy browser autoplay-with-sound restrictions, or when a translation/language track is active), not unconditionally whenever controls are shown.
- Users should still be able to mute/unmute manually at any time.
- Scheduled and always-on YouTube streams should follow the same behavior.

Saksham-Sirohi:
## Summary
- YouTube room streams no longer start muted just because player controls are visible (the default room setup).
- Autoplay is independent of mute. Mute is applied only for **Start muted**, an audio-only translation track, or an explicit caller mute.
- New YouTube rooms default to unmuted. Existing rooms with **Start muted** saved keep that setting until organisers turn it off.

Fixes #4558

## Test plan
- [x] ESLint on the three changed Vue files
- [x] Playback matrix check: default unmuted autoplay, hidden controls skip autoplay, Start muted / translation / caller mute still mute, autoplay-off stays paused
- [ ] Open a typical YouTube room (controls visible, autoplay on, Start muted off) and confirm the stream starts with sound
- [ ] Enable **Start muted** and confirm the stream still autoplays muted
- [ ] Enable an audio-only translation track and confirm the main player mutes; disable it and confirm sound returns unless Start muted is on
- [ ] Hide controls and confirm autoplay is off so a click can start playback with sound
- [ ] Repeat for always-on and schedule-driven YouTube rooms

## Risk and Rollback
Browsers may still block unmuted autoplay on a first visit; in that case the video stays paused until the viewer clicks play, then it plays with sound. Revert this commit to restore the previous muted-autoplay default.

## Summary by Sourcery

Allow YouTube room streams to autoplay with sound by default while retaining intentional muting behavior.

Bug Fixes:
- Stop YouTube room streams from being force-muted when autoplay is enabled and player controls are visible.
- Preserve muting for Start muted, audio-only translation tracks, and explicit caller mute requests while restoring sound when those conditions no longer apply.

Enhancements:
- Make new YouTube rooms default to unmuted and clarify that hiding controls disables autoplay rather than enabling muted playback.

### DIFF
--- a/app/eventyay/webapp/video/src/components/CreateStagePrompt.vue
+++ b/app/eventyay/webapp/video/src/components/CreateStagePrompt.vue
@@ -55,7 +55,7 @@ export default {
 			streamSource: 'hls',
 			url: '',
 			youtubeId: '',
-			startMuted: true,
+			startMuted: false,
 			description: '',
 			loading: false,
 			error: null,
--- a/app/eventyay/webapp/video/src/components/MediaSource.vue
+++ b/app/eventyay/webapp/video/src/components/MediaSource.vue
@@ -319,9 +319,10 @@ async function applyYoutubeTranslation(transConfig) {
 	} else {
 		// Remove translation audio iframe
 		languageIframeUrl.value = null;
-		// Unmute the main player using postMessage after a short delay
+		// Restore the main player unless the organiser asked to start muted
 		setTimeout(() => {
 			if (updateToken !== translationUpdateToken) return;
+			if (module.value?.config?.startMuted) return;
 			unmuteYouTubePlayer();
 		}, 100);
 	}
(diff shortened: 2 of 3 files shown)


Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/e07e19c9-4800-483c-b8fc-01efdcebdef9

ArnavBallinCode (approved):

PR #5196, opened 2026-08-25
fix(video): apply settings to YouTube translation streams

Saksham-Sirohi:
## Summary
- Apply the room's YouTube playback settings to hidden translation streams, including Start muted, autoplay, looping, related-video limits, video-info, keyboard controls, branding, and privacy-enhanced mode.
- Avoid delayed auto-mute postMessage races by recreating the main player with the correct mute URL when switching to audio-only translation.
- Follow-up to [PR #5177](https://github.com/fossasia/eventyay/pull/5177), addressing the remaining translation-stream issue reported after that fix.

## Test plan
- [x] `npm --prefix app/eventyay/webapp/video run lint`
- [x] `git diff --check`
- [ ] Verify switching between original, audio-only, and video translation streams in a YouTube room.
- [ ] Verify Start muted and the remaining YouTube settings on the translation stream.

## Risk and rollback
The change is isolated to YouTube translation iframe initialization and switching. Revert this PR to restore the previous translation behavior.

## Summary by Sourcery

Apply room YouTube settings consistently to translation streams and make audio-only stream switching reliably mute the main player.

Bug Fixes:
- Apply the effective YouTube playback settings to hidden translation streams, including autoplay, looping, related-video behavior, controls, branding, video information, keyboard controls, and privacy-enhanced playback.
- Ensure switching to audio-only translations applies the correct mute state without delayed postMessage races.

Enhancements:
- Use the active stream and room configuration when initializing and switching YouTube players.


ArnavBallinCode (changes requested):
The PR correctly removes the `setTimeout(() => muteYouTubePlayer(), 1000)` from `initializeIframe` since `hasAudioOnlyYoutubeTranslation()` is now included in the `shouldStartMuted` check, meaning the iframe starts muted via the `mute=1` URL parameter natively. This is a great fix for the initial load race condition!

However, replacing the `muteYouTubePlayer()` call with a complete iframe recreation inside `applyYoutubeTranslation` creates a disruptive user experience. Every time a user switches translation languages mid-stream, the entire main YouTube player will be destroyed and recreated, causing VODs to restart from the beginning and livestreams to re-buffer.

Instead of recreating the iframe, we should keep the `muteYouTubePlayer()` function and simply call it synchronously without the `setTimeout`:

```javascript
		// Create hidden translation audio iframe first
		languageIframeUrl.value = getLanguageIframeUrl(audioSource);

		// Mute the main player immediately using postMessage
		muteYouTubePlayer();
```

Since the main player is already fully loaded when the user dynamically changes the translation language, calling `muteYouTubePlayer()` immediately will reliably mute the player without causing a reload. The only time it might race is during initial mount, but as you've correctly handled, the initial mount already uses `mute=1` natively via the URL, so we don't need the `postMessage` command at that stage.

Please restore `muteYouTubePlayer()` and avoid destroying the `iframeEl` when applying a translation. Other than that, the centralization of `getYoutubeConfig()` and the translation URL params look great!

ArnavBallinCode (approved):

PR #5121, opened 2026-08-22
fix(ui): improve file upload controls

siddharthg2309:
## Summary

- restore the primary-color fallback for native file upload buttons
- make the Choose file button visible on attendee-data upload questions
- hide the default “No file chosen” text for empty file inputs across the platform
- display the actual filename after a file is selected
- preserve the existing upload, replacement, cropping, and validation behavior

Fixes #5113
Fixes #5122

## Testing

- `docker compose exec web make production`
- manually verified attendee-data file selection and upload
- manually verified empty and selected states in Event Settings
- manually verified desktop and mobile layouts
- verified existing file-size validation remains unchanged

### Before

image

### After

image

### Empty file inputs
image

### Selected file

image

## Risk

This is a frontend-only change. It does not introduce database migrations or modify the existing upload and validation logic.

## Summary by Sourcery

Restore consistent, theme-aware file upload button styling and visibility without changing upload behavior.

Bug Fixes:
- Restore visible native file upload buttons when no highlight color is configured and on attendee-data upload questions.
- Preserve selected-file visibility and reset behavior while removing redundant no-file placeholder text.

Enhancements:
- Apply consistent file input styling and selection state handling across public, organizer, control-panel, and shared pages.

## Summary by Sourcery

Improve file upload controls across the platform while preserving existing upload behavior.

Bug Fixes:
- Restore visible, theme-aware file upload controls across attendee, public, organizer, control-panel, and shared forms.
- Show selected filenames while hiding redundant empty-input placeholder text without changing upload, replacement, cropping, or validation behavior.

Enhancements:
- Centralize file-input styling and initialization for consistent behavior across platform interfaces.

PR #5198, opened 2026-08-25
fix: consistent file-selection UI

### LINKED ISSUE #5194: UI Overhaul for File Selection Form Fields
## Description

The file-selection fields across the website should be redesigned and standardized so that they all follow the same visual style, interaction pattern, and button design.

Currently, different parts of the application may use different styles for choosing/uploading files, which makes the UI feel inconsistent. The goal is to introduce a **single, reusable file-selection UI pattern** and apply it consistently wherever users are asked to choose a file.

## Requirements

* Create a consistent UI component/style for file-selection fields.
* Use the same button style, sizing, spacing, typography, and states across the application.
* Ensure the **Choose File / Browse** interaction looks consistent everywhere.
* Clearly display the selected filename.
* Provide appropriate visual states for:

  * No file selected
  * File selected
  * Hover/focus
  * Disabled
  * Validation/error
* Ensure the component works well with the existing form design system.
* Keep the UI responsive and accessible.
* Replace existing inconsistent file-selection implementations with the standardized design wherever applicable.

## Expected Outcome

After the overhaul, users should encounter the **same file-selection experience throughout the website**, regardless of which form or page they are using.

The implementation should ideally use a reusable component rather than duplicating styling across individual forms.

lxyaverse:
# fix: consistent file-selection UI

Closes #5194

## What changed

This builds on #5121 rather than changing anything it decided. The empty file label stays hidden, and the i18n meta keeps only `data-choose-file`. What this PR fixes is what happens once a file is selected, and the states the control could not express.

- **Long names stay on one line.** The name span gets `min-width: 0`, `overflow: hidden`, `text-overflow: ellipsis` and `flex: 0 1 auto`, and the wrapper gets `max-width: 100%`. A long filename no longer wraps and stretches the control.
- **Validation errors are visible on the control.** `updateFileInput()` reads `aria-invalid="true"` (which Django sets on any bound field with errors) or `.is-invalid`, and toggles `file-input-has-error` on the wrapper for a `--color-danger` border. An `invalid` listener keeps it in sync with native constraint validation.
- **Disabled inputs look disabled.** `file-input-is-disabled` gives the wrapper a muted background and a `not-allowed` cursor.
- **Hard-coded grays replaced with tokens.** `#fff`, `#e9ecef`, `#dde0e3` and `#6c757d` became `--color-bg`, `--color-grey-lighter`, `--color-grey-light` and `--color-text-muted`, so the control follows the theme.
- **Multiple selection is handled.** When an input carries `multiple`, all selected names are joined with ", " instead of only the first one being shown.
- **The label is always wired to the input.** The input id is assigned before the `<label for="...">` is built, so clicking the label always opens the picker.
- **`common/base.html` gets the i18n meta.** It loads `fileInput.js` but had no meta, so "Choose file" was untranslated on every page extending it. This adds the same tag #5121 put in `base_public.html`, with `data-choose-file` only.
- **Responsive padding under 576px** keeps the label and name row comfortable on narrow screens.

## How this was tested

Ran the stack locally on this branch (PostgreSQL, Redis, `manage.py runserver`) and used **Organiser -> Talks -> Import/Export** (`/orga/event/<organizer>/<event>/settings/import-export/`), which renders a wrapped `input[type="file"]`. Each pair below is the same page in the same browser, with only the changed files swapped between `dev` and this branch.

### 1. Empty state, unchanged

The hidden empty label from #5121 is preserved. These two are the same.

01-empty-state

### 2. File selected

The chosen filename appears next to the "Choose file" label.

02-file-selected

### 3. Long filename

`eventyay-conference-2026-speaker-roster-final-approved-with-bios-and-photos-v3.csv` used to wrap onto several lines and stretch the control. It now truncates with an ellipsis and the control keeps its 34px height.

03-long-filename

### 4. Validation error

Uploading a non-CSV returns "Please upload a CSV file." Django marks the field `aria-invalid="true"`, so the wrapper picks up `file-input-has-error` and the danger border. Before, only the message under the field turned red.

04-error-state

### 5. Responsive at 390px

With a long filename the control was 172px tall. It is now 38px, and the page has no horizontal overflow.

05-mobile-responsive

Also checked: `form.reset()` clears the name and re-hides the span, and the error class survives a re-render.

## Not covered by screenshots

- **Multiple files:** no Django-rendered file input in the app sets the `multiple` attribute today, so the comma-joined list has no page to appear on. The code path is there for when one is added.
- **Disabled state:** no current page renders a disabled wrapped file input. The read-only speaker avatar is the only disabled file field, and avatar inputs are deliberately skipped by `wrapFileInput()`.

## Summary by Sourcery

Improve file-selection controls so selected filenames, validation feedback, disabled states, localization, and responsive behavior remain consistent across the application.

New Features:
- Add consistent file-input states for selected, multiple, invalid, and disabled controls.
- Expose translated file-input labeling on the common base template.

Bug Fixes:
- Prevent selected filenames from wrapping or stretching file controls, including on narrow screens.
- Ensure file labels are always associated with their inputs and keep displayed filenames synchronized with validation and reset state.

Enhancements:
- Replace hard-coded file-input colors with theme tokens and improve responsive styling and accessibility of displayed filenames.

## Summary by CodeRabbit

* **New Features**
  * Added an improved profile avatar upload control with image preview, replacement, deletion, and file selection actions.
  * Added translated file-picker labeling and accessibility updates for upload controls.

* **Improvements**
  * File inputs now display multiple selected filenames, handle validation and disabled states more clearly, and truncate long names.
  * Updated upload controls for improved theme consistency and responsive behavior on smaller screens.


ArnavBallinCode (changes requested):

Rachit7168 (dismissed):
LGTM , Tested 

https://github.com/user-attachments/assets/debee8cb-ba3e-4ad9-8ea7-1ca3817b112c

prayag78 (approved):

Saksham-Sirohi (changes requested):
The current image name delete button and replace button should be shown in the same dialog; we do not want this
Image

app/eventyay/common/templates/common/widgets/avatar_input.html
```diff
@@ -0,0 +1,16 @@
...
+            <button type="button" class="btn btn-sm btn-outline-secondary eventyay-image-replace"{% if widget.attrs.disabled %} disabled{% endif %}>{% translate "Replace" %}</button>
+            <button type="button" class="btn btn-sm btn-outline-danger eventyay-image-delete"{% if widget.attrs.disabled %} disabled{% endif %}>{% translate "Delete" %}</button>
+        </div>
+        <input type="checkbox" class="eventyay-image-clear" name="{{ widget.checkbox_name }}" id="{{ widget.checkbox_id }}"{% if widget.attrs.disabled %} disabled{% endif %}>
+    {% endif %}
+    <div class="eventyay-file-pick-wrapper eventyay-image-picker"{% if widget.is_initial %} hidden{% endif %}>
+        <label for="{{ widget.attrs.id }}">{% translate "Choose file" %}</label>
+        <span class="eventyay-file-name text-muted small" aria-live="polite"></span>
+    </div>
+    <input type="{{ widget.type }}" name="{{ widget.name }}" class="eventyay-image-file"{% include "django/forms/widgets/attrs.html" %}>
```
Rachit7168: The initFileInput() function in fileInput.js automatically queries all input[type="file"] elements and wraps them in a newly generated .eventyay-file-pick-wrapper.

Since this template already manually provides its own .eventyay-file-pick-wrapper directly above this input, the input will end up getting processed by initFileInput() and wrapped a second time, resulting in two "Choose file" buttons displaying on the screen.

You should add data-eventyay-file-input-initialized="true" to this input so that fileInput.js ignores it and lets profile.js handle it exclusively

app/eventyay/static/cfp/js/profile.js
```diff
@@ -117,4 +117,46 @@ const initFileInput = function () {
...
+
+        if (picker) {
+            picker.addEventListener('click', (event) => {
+                if (event.target.tagName !== 'LABEL') fileInput.click()
+            })
+        }
+
+        fileInput.addEventListener('change', () => {
+            const files = fileInput.files
+            if (nameSpan) nameSpan.textContent = files && files.length ? files[0].name : ''
```
Rachit7168: There are two logic bugs here when a user selects a new file:

1. Stale UI on "Replace": If the user clicks the "Replace" button and selects a new file, this change listener fires and updates nameSpan. However, picker remains hidden and current remains visible with the old file name. The user will not see the name of the new file they just selected.
2. Django Validation Error: If the user clicks "Delete" (which checks clearCheckbox) and then decides to pick a new file via the picker, the clearCheckbox remains checked. When submitted, Django will throw a FILE_INPUT_CONTRADICTION validation error because the form simultaneously contains a new file upload and a true clear signal.

You should update the UI state and uncheck the clear checkbox whenever a new file is successfully selected

Rachit7168 (changes requested):
Thanks for working on this. Please address

app/eventyay/static/cfp/js/profile.js
lxyaverse: Thanks, both JS bugs were real — fixed in the latest commit.

**Stale UI on Replace** — correct. The name span lives inside the picker, which
is hidden while an image is set, so the row kept showing the old filename. The
change handler now writes the chosen name into the visible row too.

**FILE_INPUT_CONTRADICTION** — correct in principle, and now explicit. It happened
not to fire today because `initFileInput()` in profile.js already unchecks the
first checkbox in `.avatar-form` on any file change, which is the clear checkbox
by DOM order. That's an implicit dependency on markup order rather than a
guarantee, so the handler now unchecks it directly:

    if (clearCheckbox) clearCheckbox.checked = false

**Double wrapping** — `wrapFileInput()` already returns early on
`input.closest('.avatar-upload')`, which is the first of its six guards, and the
avatar input is inside that container, so it was never wrapped twice. I confirmed
in the browser: one `.eventyay-file-pick-wrapper`, one "Choose file" button, and
`data-eventyay-file-wrapped` never set on the input.

That guard is positional though, so I've added an explicit opt-out as you
suggested. One note: `data-eventyay-file-input-initialized` isn't a thing
fileInput.js checks — the attribute it reads is
`data-eventyay-file-wrapper="disabled"`, the same one used in
`import_start.html` and the organizer settings forms. I used that one so it stays
consistent with the rest of the codebase.

Verified all four paths in the browser: load, Replace then choose, Delete, and
Delete then choose. No console errors, and the clear flag ends up correct in
every case.

Rachit7168: Which commit?

lxyaverse: Sorry — the commit was sitting unpushed on my machine. It's up now: 01e3f5e
"fix: keep avatar name and clear flag in sync when a file is picked".

Both JS fixes are in it: the change handler writes the chosen filename into the
visible row, so Replace no longer shows the stale name, and it unchecks the clear
checkbox directly instead of relying on the existing handler's side effect. The
template also carries `data-eventyay-file-wrapper="disabled"` now.

Rachit7168 (approved):
LGTM , tested 

https://github.com/user-attachments/assets/328c3410-a01f-443f-9bd0-c29ca65290c7

Rachit7168 (commented):
Please resolve conflicts

Aqil-Ahmad (approved):

ArnavBallinCode (dismissed):
Please resolve merge conflicts so that we can merge this.

Rachit7168 (commented):
Please resolve conflicts

Saksham-Sirohi (approved):

PR #5206, opened 2026-08-26
fix(permissions): map Jitsi moderation to video moderator role

ArnavBallinCode:
Supersedes #4925.

Resolves migration collisions by updating the migration sequence number to `0063` and fixing its dependencies. The Jitsi moderator permissions are now correctly mapped to the video moderator role.

## Summary by Sourcery

Correct Jitsi moderation permission mapping for video moderators and migrate existing role assignments.

Bug Fixes:
- Map Jitsi join and moderation permissions to the video moderator role.
- Add the Jitsi permissions to the supported permission set and default role configuration.

Chores:
- Add migration 0063 to update existing event and world role assignments without migration collisions.


Rachit7168 (approved):
LGTM , Already tested the old PR

PR #5241, opened 2026-08-27
feat(eventyay-common): expand My Orders filters with code/status/date…

### LINKED ISSUE #5234: Improve "My Orders" filter UX: missing tooltip on Apply button + limited filter parameters
Two potential improvements for the "My Orders" page (`/common/orders/`) that I'd like to propose.

## 1. Missing hover tooltip on the "Apply filters" button

The button currently uses only `aria-label`, which serves screen readers but doesn't provide a hover tooltip for sighted mouse users. The neighbouring "Clear filters" button already has both `aria-label` and `title`, so this appears to be an inconsistency worth addressing.

**File:** `app/eventyay/eventyay_common/templates/eventyay_common/orders/orders.html:23-25`

Proposed change:

    <button type="submit" class="btn btn-primary"
            aria-label="{% trans 'Apply filters' %}"
            title="{% trans 'Apply filters' %}">
        <span class="fa fa-filter" aria-hidden="true"></span>
    </button>

## 2. Filter form only exposes an event dropdown

`UserOrderFilterForm` currently supports filtering only by `event`. For users with orders across multiple events, this becomes limiting — there's no way to narrow results by date range, order status, or order code.

Proposed additional filter parameters:

- **Order code** — partial-match text search
- **Status** — paid / pending / canceled / expired
- **Date range** — from / to

**Files this would touch:**

- `app/eventyay/eventyay_common/forms/filters.py` — extend `UserOrderFilterForm`
- `app/eventyay/eventyay_common/views/orders.py` — apply new filters in `get_queryset`
- `app/eventyay/eventyay_common/templates/eventyay_common/orders/orders.html` — render new inputs alongside the existing event dropdown

Both additions would follow filter-form patterns already used elsewhere in the codebase (e.g., `orga/cfp/forms.html`).

If this direction sounds reasonable, I'd be happy to open a PR — either combined or split into two, whichever the team prefers.

Rishika3D:
Fixes #5234

## Summary

Extends the "My Orders" page (`/common/orders/`) so users with orders across multiple events can narrow their results by order code, status, and date range — not just event — and adds the missing hover tooltip on the "Apply filters" button so it matches the neighbouring "Clear filters" button.

## Changes

**`app/eventyay/eventyay_common/forms/filters.py`**
- `UserOrderFilterForm` gains four new optional fields:
  - `code` — partial-match text search (CharField)
  - `status` — All / pending / paid / expired / canceled (uses `Order.STATUS_CHOICE`)
  - `date_from` and `date_to` — date pickers (HTML5 `type="date"`)
- Added a `clean()` method that surfaces a validation error on `date_to` when the range is inverted (`from > to`).

**`app/eventyay/eventyay_common/views/orders.py`**
- Refactored `get_queryset` to apply the new filters:
  - `code` → `code__icontains`
  - `status` → `status=<choice>`
  - `date_from` / `date_to` → `datetime__date__gte` / `datetime__date__lte`
- Extracted `_build_filter_form()` so the view builds the form the same way for both `get()` and `get_queryset()`.
- The invalid-form redirect now strips only the invalid fields from the query string instead of hard-coding `event`.

**`app/eventyay/eventyay_common/templates/eventyay_common/orders/orders.html`**
- Renders the four new filter inputs alongside the existing event dropdown using the same `bootstrap_field` pattern already used on the page.
- Added `title="{% trans 'Apply filters' %}"` to the Apply button so it now shows a hover tooltip, matching the Clear button.

## Verification

Tested locally on `dev`:
- Each filter narrows results independently and in combination with others.
- Invalid date range (`from > to`) surfaces the validation error on the `date_to` field without discarding the other filters.
- The Apply button now shows the "Apply filters" tooltip on hover.
- No changes for users with only a handful of orders — the extra inputs are all optional and default to empty.

## Summary by Sourcery

Expand My Orders filtering to help users find orders across events more precisely.

New Features:
- Expand the My Orders page with optional filtering by order code, status, and date range.

Bug Fixes:
- Validate inverted date ranges and preserve valid filter parameters when redirecting invalid submissions.

Enhancements:
- Improve filter form consistency and add client-side date-range constraints and filter clearing behavior.
- Add a hover tooltip to the Apply filters button.


Saksham-Sirohi (changes requested):
please add video showcasing the changes

Saksham-Sirohi (changes requested):
Change string from all statuses to all status also when you apply the date filter the dates are not actually shown please fix

Rachit7168 (changes requested):
https://github.com/user-attachments/assets/b2f6ef14-36ae-49eb-b44b-de79d59752d6

Clearing the filter does not clear the selected date, and there is also no indication whether it is the start date or the end date

app/eventyay/eventyay_common/views/orders.py
```diff
@@ -16,30 +16,41 @@ class MyOrdersView(PaginationMixin, ListView):
     template_name = 'eventyay_common/orders/orders.html'
     paginate_by = 25
 
+    def _build_filter_form(self):
```
Rachit7168: Because self._build_filter_form() is explicitly called in get(), get_queryset(), and get_context_data(), the filter form is being instantiated and validated three distinct times for a single page load. This causes unnecessary processing and redundant database queries (since the form's is_valid() checks and __init__ do DB lookups for the events dropdown).

You should use Django's @cached_property to ensure the form instance is built and validated only once per request

app/eventyay/eventyay_common/templates/eventyay_common/orders/orders.html
```diff
@@ -72,4 +84,50 @@ <h1>{% trans "My Orders" %}</h1>
...
+                    } else {
+                        fromInput.removeAttribute('max');
+                    }
+                }
+                fromInput.addEventListener('change', syncRangeLimits);
+                toInput.addEventListener('change', syncRangeLimits);
+                syncRangeLimits();
+            }
+
+            var clearBtn = document.querySelector('[data-clear-filters]');
```
Rachit7168: The JavaScript event listener block that manually loops through and resets inputs for clearBtn is redundant and can be safely removed.

Since the "Clear filters" button is built as an anchor tag (<a href="{{ request.path }}">), clicking it natively instructs the browser to navigate to the current URL without any query parameters. The browser will navigate away immediately to fetch the clean URL from the server, making the JS DOM-clearing loop completely unnecessary. (If you were intercepting the click to clear the form via JS, you'd need e.preventDefault() to stop the navigation—but letting the native href handle the filter clearing is already the perfect approach here).

Rachit7168 (changes requested):
Thanks for working. Please address

Rachit7168 (approved):
Image

LGTM , Tested : Add UI indication for start date and end date

Saksham-Sirohi (approved):

PR #5274, opened 2026-08-28
Fix TypeError when creating and editing subevents in bulk

### LINKED ISSUE #5272: Bulk sub-event creation ("Create multiple dates") crashes with TypeError: tzinfo argument must be None or of a tzinfo subclass, not type 'str'
Submitting the "Create multiple dates" form (bulk sub-event creation via recurrence rule) crashes with an unhandled HTTP 500 error, regardless of the recurrence rule values entered. This happens on every submission, blocking the entire bulk sub-event creation feature.

## Steps to Reproduce
1. Create or use an event configured as an "Event series."
2. Go to Tickets Dashboard → Dates → "Create many new dates."
3. Fill in the form with any valid values (e.g. Repeat every: 1, week(s), Repeat for: 10 times, valid Event start time).
4. Click Save.
5. Observed: Unhandled HTTP 500 error:

## Expected Behavior
The form should save successfully and create the sub-events as configured, without any server error.

## Actual Behavior
Crashes with an unhandled TypeError every time, even with fully valid input.

## Root Cause
In app/eventyay/control/views/subevents.py, line 866, in form_valid():
```python
se.date_from = make_aware(datetime.combine(rdate, t['time_from']), tz)
```
`tz` is being passed as a string (e.g. "UTC") instead of an actual tzinfo object. Django's `make_aware()` requires a proper tzinfo instance (e.g. `zoneinfo.ZoneInfo` or `pytz.timezone(...)`), not a raw string. This suggests `tz` was likely derived from something like `event.settings.timezone` (a string setting) without being converted into an actual timezone object before being passed to `make_aware()`.

## Impact
- Severity: Critical — completely blocks the bulk sub-event creation feature; no valid input can work around it.
- Affected area: Control Panel → Dates → "Create many new dates" (bulk_add view for series events).

## Affected Files
- app/eventyay/control/views/subevents.py (form_valid(), around line 866, and any other make_aware() calls in the same view using the same `tz` variable)

## Proposed Fix
- Locate where `tz` is defined in form_valid() (likely something like `tz = self.request.event.settings.timezone` or similar).
- Convert it to a proper timezone object before use, e.g.:
```python
  from django.utils import timezone as tz_utils
  tz = tz_utils.get_current_timezone() if isinstance(tz, str) is False else ZoneInfo(tz)
```
  or more directly:
```python
  from zoneinfo import ZoneInfo
  tz = ZoneInfo(event.settings.timezone) if isinstance(event.settings.timezone, str) else event.settings.timezone
```
- Check for any other `make_aware()` calls in the same file/view using the same pattern, and fix consistently.
- Add a regression test that submits the bulk "Create multiple dates" form with valid data and confirms it succeeds (HTTP 302 redirect / success), not a 500.

## Test Plan
- Submit "Create multiple dates" with fully valid data → succeeds, sub-events created, no crash.
- Confirm date_from/date_to on created sub-events reflect the correct timezone-aware datetime.

## Demo Video
Will attach a recording showing the crash on valid submission before the fix, and successful creation after the fix.

https://github.com/user-attachments/assets/4b70a0cf-0f1b-4daf-8b6b-cfb61181f111

anuragverma025:
### Summary of Changes
Fixes #5272.

Submitting the "Create multiple dates" (bulk sub-event creation) or "Bulk edit" forms crashed with `TypeError: tzinfo argument must be None or of a tzinfo subclass, not type 'str'`.

#### Root Cause
`self.request.event.timezone` and `self.event.timezone` return a `CharField` string (e.g. `'Europe/Berlin'`), but `django.utils.timezone.make_aware()` and `datetime.astimezone()` require an actual `tzinfo` instance.

#### Changes
- In `app/eventyay/control/views/subevents.py` (`SubEventBulkCreate`), converted timezone string to `ZoneInfo(self.request.event.settings.timezone)`.
- In `app/eventyay/control/forms/subevents.py` (`SubEventBulkEditForm`), converted timezone string to `ZoneInfo(self.event.settings.timezone)`.
- Updated bulk creation and bulk edit test cases in `app/tests/tickets/control/test_subevents.py`.

### Verification
Ran test suite inside Docker:
```bash
docker compose exec web pytest tests/tickets/control/test_subevents.py
 
### Demo Video
will attach a recording showing on valid submission after the fix.

https://github.com/user-attachments/assets/568d253c-7e59-4ea0-88f9-f2731229070c

## Summary by Sourcery

Fix timezone handling in bulk sub-event creation and editing so submitted forms process successfully with configured event timezones.

Bug Fixes:
- Fix bulk sub-event creation and editing crashes caused by event timezones being passed as strings instead of timezone objects.

Enhancements:
- Align sub-event bulk date and time handling with configured event timezones, including daylight-saving transitions.

Tests:
- Update sub-event control tests to cover timezone-aware bulk operations and current product-based ticket associations.


Rachit7168 (changes requested):
When creating a PR that affects the UI or any existing workflow, please include a **screencast demonstrating the changes and the affected areas**.

Also, please **do not include the issue number in the PR title**. The issue number can be referenced in the PR description instead.

Rachit7168 (approved):
LGTM , tested 

https://github.com/user-attachments/assets/21f82b74-e569-4d39-8707-e7cc9edea5c5

Saksham-Sirohi (approved):
Bulk create/edit now stamps dates with `ZoneInfo(event.settings.timezone)` instead of a tzinfo that `datetime.replace()` cannot use. Tests match the DST-aware offsets.

PR #5285, opened 2026-08-29
fix(api): remove unauthenticated billing task trigger endpoint

adity1raut:
`/api/v1/billing-testing/<task>` was still registered in the production URLconf. It's served by `BillingInvoicePreview`, a plain `django.views.View` with no login check, no permission class and no throttle. It isn't a DRF view either, so the API's authentication never applies to it, and the middleware that guards the control panel skips anything outside /control, /common, /admin, /teamshifts, /exhibitors and /social.

The billing tasks are also called directly rather than with `.delay()`, so they run synchronously inside the web request instead of on a worker.

The view's own docstring says it is for testing and would be removed before production, and nothing else in the tree references it, so remove the view and its route. The tasks stay reachable from the Celery beat schedule and from a management shell for local debugging.

Adds regression tests: the path no longer resolves, the route name is no longer registered, an anonymous GET returns 404, and the view module is gone.

## Summary by Sourcery

Remove the unauthenticated billing task trigger endpoint and verify that it is no longer reachable or registered.

Bug Fixes:
- Remove the unauthenticated production billing task trigger endpoint to prevent anonymous callers from executing billing operations synchronously.

Tests:
- Add regression coverage confirming the billing route, route name, endpoint, and view module are no longer available.


Rachit7168 (approved):
LGTM , Approved! Great catch on finding this leftover testing endpoint !

ArnavBallinCode (approved):

PR #5294, opened 2026-08-29
fix(base): import cache in room service periodic receiver

adity1raut:
`check_stream_schedule_changes` in `base/services/room.py` calls `cache.get()` and `cache.set()`:

```python
last_broadcast_id = cache.get(cache_key)
...
    cache.set(cache_key, current_stream_id, cache_timeout)
```

but the module never imports `cache`. It pulls `NONE_SENTINEL`, `cache_delete`, `deserialize_none_sentinel` and `get_stale_cached` from `base.services.stale_cache`, and no cache object. Both lines raise `NameError`.

## Why it matters beyond this one function

It's a `@receiver(signal=periodic_task)`, and `send_periodic_signal` fires that signal every minute from Celery beat:

```python
# app/eventyay/common/tasks.py:138
periodic_task.send(sender=None)
```

`Signal.send()` doesn't swallow receiver exceptions, unlike `send_robust()`. There are 21 receivers on this signal — order cleanup, invoice generation, auth cleanup, telemetry, update checks. So once the exception fires, everything dispatched after this receiver is skipped too, and it repeats every minute.

The trigger is the loop being entered at all:

```python
rooms = Room.objects.filter(deleted=False, stream_schedules__isnull=False)...
for room in rooms:
    ...
    last_broadcast_id = cache.get(cache_key)   # NameError
```

So an instance with no stream schedules is unaffected, and one with a single scheduled room has a broken periodic pipeline.

## The fix

```python
```

That's the pattern the neighbouring services already use — `base/services/geo.py`, `user.py`, `orders.py` and `stale_cache.py` all import it exactly this way.

## Tests

Added `app/tests/stable/test_room_stream_schedule_periodic.py`, in `tests/stable` because that's the suite `video-backend-ci.yml` runs.

- `cache` is bound in the module
- the loop body actually executes for a room with a stream schedule

Neither needs a database. The `Room` queryset is mocked, and the test cache backend is `DummyCache`, so `cache.get()` returns `None`, the current stream is `None`, and the `current_stream_id != last_broadcast_id` branch isn't taken — meaning no channel layer is required either.

Bug Fixes:
- Import Django's cache backend so the periodic room stream-schedule receiver no longer raises a NameError and aborts subsequent periodic signal handlers.

Tests:
- Add database-free regression tests covering cache binding and execution for rooms with stream schedules.


ArnavBallinCode (approved):
Thank you for your interest in contributing to the project.

For new contributors, please start by working on existing confirmed issues before opening new issues or submitting related pull requests or atleast create an issue and link the PR to the issue to give reviewers relevant context. This helps ensure that your contribution matches an actual project need and can be reviewed effectively.

Saksham-Sirohi (approved):

PR #5301, opened 2026-08-29
Improve performance and reliability of the admin users page

### LINKED ISSUE #4773: Improve performance and reliability of the admin users page
The admin users page at `/admin/users/` has multiple issues affecting its usability and reliability.

Image

### 1. The page can result in a gateway timeout

Opening the page can result in:

> 504 Gateway Time-out

The page currently requests the complete user list through the unpaginated `user.list` endpoint.

For administrators, the backend also resolves additional information for every user, including:

* email address;
* external token ID;
* Wikimedia username;
* moderation state;
* order code; and
* ticket code.

On events with many users and orders, retrieving and enriching the complete user list can exceed the configured proxy timeout.

Although the frontend uses a virtual scroller, this only limits browser rendering. The backend still loads, enriches, serializes, and returns every user before the page becomes available.

### 2. Administrative user changes are not always persistent

Administrative actions and toggles on the user detail and user list pages must also be reviewed.

For example, marking a user as verified appears to succeed in the interface, but the change is not persistent. After reloading the page, the user is shown as unverified again.

This may indicate that:

* the frontend updates only its local state;
* the update request is not sent;
* the request uses the wrong field or endpoint;
* the backend accepts the request but does not save the field;
* the field is excluded from `update_fields`;
* the updated value is overwritten during synchronization;
* the API response does not reflect the saved value; or
* the page reload fetches the verification state from a different user record.

Other administrative toggles and editable fields may be affected by the same problem and must be tested.

## Root cause of the performance problem

The admin page currently uses the unpaginated `user.list` endpoint even though eventyay already provides the paginated `user.list.search` endpoint.

The existing paginated endpoint supports:

* page-based loading;
* configurable page size;
* server-side search;
* inclusion of administrative user fields;
* inclusion or exclusion of banned users; and
* an `isLastPage` response value.

The page also filters search results locally, which requires the complete user list to be loaded before search becomes usable.

### Use server-side pagination

Replace the complete `user.list` request with paginated calls to:

```javascript
api.call('user.list.search', {
  search_term: search,
  page,
  include_banned: true
})
```

On initial page load:

* request only the first page;
* display the results immediately;
* do not wait for the complete user list.

Load subsequent pages when the administrator reaches the end of the current list or selects a **Load more** action.

### Use server-side search

Move user search from client-side filtering to the paginated backend endpoint.

When the search value changes:

1. Debounce the request.
2. Reset the current page to `1`.
3. Clear the existing results.
4. Request the first matching page.
5. Ignore stale responses from previous search terms.

Search should support relevant administrative fields, including:

* user ID;
* external ID;
* display name;
* email address;
* ticket code;
* order code; and
* Wikimedia username.

### Verify persistence of all administrative changes

Review every editable field, toggle, and user action available from the admin users interface.

At minimum, test:

* verified or unverified state;
* active or disabled state;
* banned or unbanned state;
* silenced or unsilenced state;
* user roles or permissions;
* profile visibility;
* email-related status fields;
* external account fields;
* Wikimedia username;
* display name and other editable profile fields; and
* any event-specific user settings exposed in the interface.

For every control:

1. Change the value in the interface.
2. Confirm that the correct API request is sent.
3. Confirm that the backend saves the value.
4. Confirm that the API response contains the saved value.
5. Reload the page.
6. Confirm that the saved value remains unchanged.
7. Reopen the user from another browser session and verify the same state.

The frontend must not display a successful state before the backend confirms that the update has been persisted.

If saving fails, the control should return to its previous state and display an error.

## Verification-state investigation

Specifically investigate the verified-user toggle.

Determine:

* which model field represents the verification state;
* whether the admin interface edits the platform account or an event-scoped video user;
* which endpoint receives the update;
* whether the endpoint allows this field to be changed;
* whether the field is included when the model is saved;
* whether another login, import, or synchronization process overwrites it;
* whether duplicate account and event-user records exist for the same person; and
* whether the page reads the value from the same record that was updated.

The verified state should have a single authoritative source.

If platform account verification and event-specific user verification are separate concepts, the interface must label them clearly and update the correct state.

## Loading and error states

The interface should distinguish between:

* initial loading;
* loading another page;
* no users found;
* no matching search results;
* failure loading the first page;
* failure loading an additional page; and
* failure saving an administrative change.

A failed additional-page request must not remove users that are already displayed.

A failed update must not leave the interface showing a value that was not saved.

## Backend considerations

Administrative-field enrichment should run only for users on the requested page.

The backend must not:

* retrieve all users before applying pagination;
* resolve ticket information for users outside the requested page;
* scan all platform users separately for every displayed user; or
* scan all event orders separately for each user.

Also review whether the user update endpoint:

* explicitly allows all fields exposed in the admin interface;
* validates submitted values;
* writes every changed field;
* uses correct `update_fields` values;
* returns the saved database state; and
* records relevant administrative changes in the audit log.

### Performance

* [ ] `/admin/users/` loads only the first page of users initially.
* [ ] The initial page no longer calls the unpaginated `user.list` endpoint.
* [ ] The page becomes usable without waiting for all users to load.
* [ ] Opening the page no longer results in a 504 gateway timeout.
* [ ] Additional users can be loaded without reloading earlier pages.
* [ ] Search is performed on the server.
* [ ] Changing the search term resets pagination.
* [ ] Stale search responses do not replace newer results.
* [ ] Administrative fields are enriched only for users on the requested page.
* [ ] A failed subsequent-page request does not clear existing results.

### Persistence

* [ ] Marking a user as verified remains persistent after reloading the page.
* [ ] Marking a verified user as unverified remains persistent after reloading.
* [ ] The verification toggle updates the correct user or account record.
* [ ] The UI changes state only after the backend confirms the update.
* [ ] A failed update restores the previous UI state and displays an error.
* [ ] All other administrative toggles and editable fields are tested for persistence.
* [ ] Ban, unban, silence, unsilence, activation, and deactivation changes remain persistent.
* [ ] Saved values remain consistent in another browser session.
* [ ] API responses return the actual saved state.
* [ ] Relevant administrative changes are recorded in the audit log.

### Testing

* [ ] Frontend tests cover pagination, search, stale responses, and loading failures.
* [ ] Frontend tests cover successful and failed toggle updates.
* [ ] Backend tests verify that only one page is serialized and enriched.
* [ ] Backend tests verify persistence of every field exposed by the admin interface.
* [ ] Tests reload each updated user to confirm the value was committed.
* [ ] Tests cover platform accounts and event-scoped users where both are supported.
* [ ] Performance is tested with an event containing many users and orders.

## Performance verification

Record the following before and after the change:

* time to first user-list response;
* total database query count;
* slowest database queries;
* number of users serialized;
* response payload size;
* server request duration; and
* browser time until the first page is usable.

The initial request duration should depend on the configured page size rather than the total number of users in the event.

ArnavBallinCode:
## Description
This PR addresses issue #4773 by improving the performance and reliability of the admin users page.

### 1. The page can result in a gateway timeout
Opening the admin users page previously caused a 504 Gateway Time-out on large events due to the unpaginated `user.list` API endpoint serializing and rendering every user at once. 
- **Fix:** Switched the frontend `users.vue` to use the paginated `user.list.search` endpoint instead.
- Implemented state properties to handle pagination via a "Load more" pattern that tightly integrates with the `RecycleScroller`.
- Replaced client-side search filtering with debounced server-side search to keep loading fast without fetching thousands of records.

### 2. Administrative user changes are not always persistent
When toggling the verification state of a user on the Django admin interface, the frontend updated the visual state, but upon reload, it reverted to unverified. The `EmailAddress.save(update_fields=['verified'])` logic was suffering from database consistency or `allauth` hooks issues where updates were overridden.
- **Fix:** Refactored `_handle_toggle_verified` in `users.py` to use a direct ORM SQL update (`EmailAddress.objects.filter(pk=primary_email.pk).update(verified=new_verified_status)`). This bypasses the model's `save` hooks ensuring the persistence of verification edits directly in the DB.

Fixes #4773

## Summary by Sourcery

Improve the admin users experience by making large user lists responsive and ensuring verification updates remain persistent.

New Features:
- Add paginated, debounced server-side searching with incremental loading to the admin users page.

Bug Fixes:
- Prevent admin user listings from timing out on large events.
- Ensure user email verification changes persist reliably in the database.

Enhancements:
- Expand administrative user search to cover identifiers, contact details, Wikimedia usernames, and ticket-related references.


app/eventyay/base/services/user.py
```diff
@@ -1171,6 +1171,28 @@ def list_users(
...
             )
+
+        if include_admin_info:
+            conditions.append(Q(token_id__icontains=search_term))
+            conditions.append(Q(email__icontains=search_term))
+            conditions.append(Q(wikimedia_username__icontains=search_term))
+
+            try:
+                import uuid
+                conditions.append(Q(id=uuid.UUID(search_term)))
```
Saksham-Sirohi: This only matches a full UUID. The old admin filter used `user.id.startsWith(q)`, so a partial id (the usual case while typing or pasting a prefix) now returns nothing. Cast `id` to text and `icontains`/`istartswith` it, same as `token_id`.

Saksham-Sirohi (changes requested):
Sourcery’s race / debounce / toggle notes look addressed. One remaining search gap vs the old client filter.

app/eventyay/base/services/user.py
```diff
@@ -1171,6 +1171,27 @@ def list_users(
             conditions.append(
                 Q(**{"profile__fields__" + field + "__icontains": search_term})
             )
+
+        if include_admin_info:
+            conditions.append(Q(token_id__icontains=search_term))
+            conditions.append(Q(email__icontains=search_term))
```
Saksham-Sirohi: Also match `profile__contact_email`. The admin list shows that as `user.email` for JWT/ticket personas (`User.email` is often empty), so searching an address that used to match the client filter now returns nothing. Move the `Cast`/`CharField` imports to module scope too.

Saksham-Sirohi (changes requested):
Partial user-id search looks fixed. One leftover from the old client filter: ticket-user emails live on the profile, not `User.email`.

Saksham-Sirohi (approved):

PR #5305, opened 2026-08-30
fix(orga): block room deletion while sessions are scheduled in it

### LINKED ISSUE #4450: Room deletion lacks warning or restriction when linked to scheduled sessions
### **Description**

When an administrator attempts to delete a room that currently contains active, scheduled sessions, the system allows the action to proceed without throwing a warning or a blocking constraint. It should notify the admin that the room is linked to an existing schedule.

### **Steps to Reproduce**

1. Navigate to the **Schedule** or **Rooms** management view under the Event administration panel.
2. Identify a room (e.g., `"Introduction"`) that explicitly has active event sessions scheduled on the calendar grid.
3. Click the delete option next to the room to bring up the **Confirm deletion** prompt.
4. Observe that the confirmation page only displays a generic message: *"Please make sure that this is the item you want to delete. This action cannot be undone!"* without checking or highlighting existing session conflicts.

### **Expected Behavior**

The system should detect if the room has scheduled sessions attached to it. It should either:

* Block deletion until the sessions are moved or deleted.
* Display a prominent warning stating: `"This room has linked schedules/sessions. Deleting it will affect the event schedule."`

### **Actual Behavior**

The confirmation page presents a generic deletion block with no context regarding data dependencies or linked schedules, allowing accidental disruption of the event timeline.

https://github.com/user-attachments/assets/07bb988f-ab4e-490e-b2f2-327fa10510d1

adity1raut:
Closes #4450

## Why it happens

Room deletion in the schedule view is a soft delete  `RoomView.delete_handler` sets a flag instead of calling `delete()`:

```python
def delete_handler(self, request, *args, **kwargs):
    # Use soft delete to sync with video component
    obj = self.get_object()
    obj.deleted = True
    obj.save(update_fields=['deleted'])
    request.event.wip_schedule.talks.filter(room=obj, submission__isnull=True).delete()
```

Because nothing is ever deleted at the database level, the `on_delete=models.PROTECT` on `TalkSlot.room` never fires. There's no other check, so a room with sessions in it goes through.

The cleanup on the next line only removes slots **without** a submission, i.e. breaks and placeholders. Real sessions keep their `room` pointing at the room that was just flagged. Since `RoomView.get_queryset()` filters on `deleted=False`, the room then disappears from the organiser UI while the schedule still references it.

## What I changed

Refuse the deletion when the room still has linked submissions, and warn on the confirmation page rather than only rejecting the POST the issue asks for either behaviour, and this gives both, so the organiser finds out before they confirm rather than after.

There's already a helper for exactly this condition, `room_has_linked_submissions()` in `base/models/room.py`, used by `validate_is_unscheduled_change()` to block marking a room unscheduled while sessions are in it. Blocking deletion is the same rule applied to a second action, so I reused the helper and put the new message next to `UNSCHEDULED_LINKED_SUBMISSIONS_MESSAGE`.

Rooms with no scheduled sessions delete exactly as before.

## Tests

`test_delete_used_room` looked like it covered this, but it only issued a GET and asserted the room count was unchanged  it never POSTed, so the deletion itself was never exercised. That's why this went unnoticed. Extended it to POST and assert the room survives with its sessions intact.

I asserted on `room.deleted` rather than `event.rooms.count()`, because `Room.objects` has no default filter on `deleted`  `event.rooms.count()` still counts soft-deleted rooms, so a count assertion wouldn't actually prove anything here.

## Summary by Sourcery

Block soft deletion of rooms with scheduled sessions and provide clear feedback before and during deletion attempts.

Bug Fixes:
- Prevent deletion of rooms that still have scheduled sessions, preserving the room and its session associations.
- Warn organisers on the room deletion confirmation page and reject attempted deletion when linked sessions remain.

Tests:
- Extend the used-room deletion test to submit the deletion request and verify the room and scheduled session remain intact.


app/tests/talk/orga/views/test_orga_views_schedule.py
```diff
@@ -466,6 +466,17 @@ def test_delete_used_room(orga_client, event, room, slot):
     with scope(event=event):
         assert event.rooms.count() == 1
 
+    # Deleting a room is a soft delete, which bypasses the PROTECT on TalkSlot.room, so the
+    # POST has to be refused explicitly. Otherwise the room vanishes from the organiser UI
+    # while the sessions scheduled in it keep pointing at it.
+    response = orga_client.post(room.urls.delete, follow=True)
+    assert response.status_code == 200
```
ArnavBallinCode, line 473: Please move the test in the correct directory

ArnavBallinCode (changes requested):

ArnavBallinCode (changes requested):

PR #5386, opened 2026-09-03
fix(control): remove redundant eventyay Check-in sidebar entry

### LINKED ISSUE #5384: Remove redundant "eventyay Check-in" sub-tab under Check-in navigation
**Description:**
Under the **Check-in** sidebar menu, there is a sub-tab named **eventyay Check-in**. However, on the primary **Check-in lists** page, an **Open eventyay Check-in** action button is already present at the top of the interface. Having both the sidebar sub-tab and the button creates duplicate ways to access the exact same feature.

To streamline the sidebar navigation and reduce redundancy, the sub-tab under **Check-in** should be removed, leaving the **Open eventyay Check-in** action button on the main page.

**Steps to Reproduce:**

1. Navigate to **Check-in** in the sidebar menu.
2. Observe the expanded sub-menu items: **Check-in lists** and **eventyay Check-in**.
3. View the **Check-in lists** main page and note the **Open eventyay Check-in** button positioned alongside **Create a new check-in list** and **Connected devices**.

**Expected Result:**
The duplicate **eventyay Check-in** sub-menu entry in the sidebar should be removed to keep navigation clean and non-redundant.

**Actual Result:**
The feature is accessible both through the top-level button on the **Check-in lists** view and as a separate sidebar sub-tab.

Image

siddharthg2309:
## Summary

- remove the `eventyay Check-in` child entry from the Check-in sidebar menu
- keep the `Open eventyay Check-in` button on the check-in lists page
- drop the now-unused `checkin_app` import from `navigation.py`

The sidebar entry and the button linked to the same URL, so the feature was
reachable two ways from the same screen. The button is unaffected by this
change: its context comes from the check-in list view
(`ctx['can_open_checkin_app']` / `ctx['checkin_app_url']`), not from navigation.

Fixes #5384

## Testing

- `pytest tests/tickets/control/test_navigation.py` — identical results with and
  without the change (4 failures, 3 passes, same tests both times). The failures
  reproduce on unmodified `dev` in my local environment, so they are unrelated to
  this change.
- `ruff check` on the touched file reports the same findings before and after
- manually verified on a local dev server: the sidebar entry is gone, and the
  **Open eventyay Check-in** button still renders and still points at the check-in app

### After

Screenshot_20260903-132220

Screenshot_20260903-131745

## Risk

Removes one navigation entry. No model, view, URL or permission change, and no
migration. The `user_can_open_checkin_app` / `get_eventyay_checkin_app_url`
helpers remain in use by the check-in list view. Reverting the commit restores
the entry.

## Summary by Sourcery

Remove the duplicate eventyay Check-in navigation entry while keeping the check-in application available from the check-in lists page.

Bug Fixes:
- Remove the redundant eventyay Check-in entry from the Check-in sidebar while preserving access through the existing check-in lists button.

Enhancements:
- Remove the unused check-in application imports from control navigation.

## Summary by CodeRabbit

- **Changes**
  - Removed the external “eventyay Check-in” option from the event navigation menu.
  - The check-in submenu now displays only “Check-in lists.”


Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/82dfd1bf-7c66-4265-9e3a-2f5c22ab787c

ArnavBallinCode (approved):

Issue #5414, opened 2026-09-04
Implement built-in cookie consent system using Klaro and admin-managed privacy configuration


mariobehling:
Eventyay should implement a built-in privacy and cookie consent system using **Klaro** as the frontend consent manager, while keeping all configuration, service metadata, policy pages, and consent records managed by Eventyay through Django/admin settings.

This avoids hardcoding a commercial consent provider and keeps Eventyay suitable for open-source, self-hosted and multi-tenant deployments.

The implementation should support consent banners, script blocking, embedded-content placeholders, static/legal policy pages, consent storage, and an optional external CMP integration for deployments that want to use services such as CookieYes, OneTrust or similar providers.

## Goal

Create a configurable privacy and consent system that:

* works out of the box for Eventyay deployments
* is open-source and self-hostable
* blocks non-essential scripts before consent
* supports privacy/cookie policy pages
* supports third-party service classification
* allows admins to configure consent behavior
* supports optional external CMP scripts
* prepares the platform for better GDPR, California and international privacy compliance

## Current behavior

* Privacy and cookie policy pages exist but are mostly empty.
* There is no central admin-managed privacy/compliance configuration.
* There is no first-visit cookie consent banner.
* Optional third-party scripts and embeds are not consistently blocked before consent.
* There is no service registry defining which third-party services belong to which consent category.
* There is no consent preference center.
* There is no reusable mechanism for contextual consent on embedded content such as YouTube videos.

## Expected behavior

Eventyay should provide a built-in consent system with this architecture:

```text
Django / Eventyay backend:
- privacy and cookie policy configuration
- consent settings
- consent records
- third-party service registry
- admin UI
- audit logs

Klaro frontend layer:
- consent banner
- preference modal
- script blocking
- service activation after consent
- contextual consent for embeds

Optional external CMP mode:
- allow admins to use an external consent management provider if needed
```

## Admin navigation

Add a new admin section:

```text
Global settings
- Privacy & Compliance
```

Suggested tabs:

```text
Overview
Legal pages
Cookie consent
Third-party services
External CMP
Consent records
Audit log
```

## 1. Consent provider settings

Add admin setting:

```text
Consent provider
- Disabled
- Built-in Eventyay consent using Klaro
- External CMP script
```

Default for existing deployments:

```text
Disabled
```

Default for new installations can remain disabled unless explicitly enabled by the platform administrator.

## 2. Built-in Eventyay consent using Klaro

[...]

PR #5438, opened 2026-09-06
feat: add built-in cookie consent system using Klaro

sanketkokate712:
Refs #5414.

This is the first slice of the privacy/consent system, not the whole issue. #5414 covers 14 areas; landing them in one PR would be unreviewable, so this PR builds the foundation everything else hangs off — configuration, the service registry, and a Klaro consent layer — and leaves the rest for follow-ups. I'd rather get the consent semantics reviewed carefully before building consent records and audit logging on top of them.

## What this adds

**Admin → Global settings → Privacy & Compliance**

- Consent provider: `Disabled` / `Built-in Eventyay consent using Klaro` / `External CMP script`
- Per-category toggles for Functional, Analytics, Marketing and Embedded content
- Cookie Policy / Privacy Policy URLs
- External CMP provider name and script URL
- Overview warnings when the banner is on but no Cookie Policy is published, or a service has no category

**Third-party service registry** — a `ThirdPartyService` model (identifier, provider, purpose, consent category, enabled, privacy policy URL, cookie names). The Klaro configuration is generated from it. Registering a service lists it in the consent modal and lets Klaro clear its cookies when consent is declined or withdrawn. **It does not block the service's scripts by itself:** today only embeds rendered with `{% consent_embed %}` wait for consent. Gating scripts (e.g. a `consent_script` tag emitting Klaro's `data-name` markup) is a follow-up.

**Frontend consent layer** — Klaro is configured from those settings and the registry, with a "Privacy settings" footer link that reopens the preference modal, a Cookie Policy link in the notice, modal and footer, and a `{% consent_embed %}` tag that renders a placeholder instead of a third-party iframe until the visitor consents to that service. The placeholder is only used when the service is published in the Klaro config; otherwise the iframe renders directly, so it can never get stuck.

## Decisions worth reviewing

**Consent is opt-in, enforced in the generated config rather than the UI.** The config sets `default: false`, advertises only categories an admin has enabled, and publishes only services whose category is enabled. A service registered as Analytics while Analytics is off is absent from the config entirely, so Klaro never reports consent for it. The alternative — emitting every service and relying on Klaro's defaults — means one config mistake silently ships a live tracker, so I preferred the version that fails closed.

**Defaults to `Disabled`.** Existing deployments see no behaviour change until an administrator opts in, per the issue.

**Klaro mode requires a published Cookie Policy and Privacy Policy.** Both must be http(s) URLs. Form validation rejects the combination rather than rendering a banner that links to a dead policy page — a banner pointing at nothing is arguably worse than no banner for compliance purposes.

**Provider is one choice, not independent switches**, which is what structurally prevents the built-in banner and an external CMP from both running (issue section 9).

**Klaro 0.7.22 is vendored** under `static/klaro/` (BSD-3-Clause, LICENSE included). `static/` has no bundler — it is served as plain files, like `leaflet/`, `moment/` and `select2/` — so a bare `import` specifier would not resolve in the browser. The config is passed through a JSON script tag rather than an inline script, keeping the page free of inline JS and compatible with a strict CSP.

**Two root templates, not one.** `startpage_base.html` does not extend `base.html`, so the include is in both. I found this because the banner rendered everywhere except the homepage.

## Testing

8 tests in `app/tests/base/test_privacy_consent.py`, covering the blocking semantics rather than the rendering: optional categories off by default, a service staying invisible until its category is enabled, necessary services always active, disabled services never published, and the two provider modes being mutually exclusive.

Verified manually on a local stack: the admin page renders, and with Analytics enabled plus Google Analytics (Analytics) and Stripe (Strictly necessary) registered, the homepage emits

```json
{"purposes": ["necessary", "analytics"],
 "services": [{"name": "google-analytics", "required": false, "default": false},
              {"name": "stripe", "required": true, "default": true}],
 "default": false}
```

## Deliberately not in this PR

Consent record storage, audit logging, admin CRUD for the service registry (registry entries are currently created via shell/fixtures), the legal-pages workflow with versioning and locales, and rollout to the video/schedule webapps. Each is a separate reviewable change and I'd like to confirm the approach here first.

## Question for reviewers

Section 4 asks for server-side consent records with a consent ID, anonymous visitor ID, policy version and source page. That is the one part I did not want to guess at, because storing an anonymous visitor ID plus timestamp and page context creates a new store of personal data whose own lawful basis and retention policy need deciding — and getting that wrong makes the compliance position worse, not better.

Should consent records be stored for all visitors, or only for authenticated users, with anonymous consent left in the Klaro cookie? And is there an existing retention/cleanup mechanism in Eventyay I should hook into rather than adding a new one?

## Summary by Sourcery

Establish the privacy consent foundation with administrator configuration, a categorized third-party service registry, and fail-closed frontend consent enforcement.

New Features:
- Add administrator configuration for disabled, built-in Klaro, or external consent providers, consent categories, and policy URLs.
- Introduce a third-party service registry that classifies services by consent category and supplies their public consent metadata.
- Add frontend consent handling with Klaro, privacy-settings controls, and consent-gated embedded content placeholders.

Bug Fixes:
- Prevent unknown or malformed stored PayPal endpoint values from being silently treated as the live endpoint.

Enhancements:
- Enforce opt-in consent semantics by publishing only enabled categories and their services while keeping necessary services active.
- Keep consent configuration in a CSP-compatible JSON payload and safely escape administrator-provided service content.

Build:
- Vendor Klaro 0.7.22 and its license under the static assets.

Tests:
- Add coverage for consent provider modes, category and service blocking behavior, embed gating, cookie serialization, configuration escaping, and PayPal endpoint validation.

## Summary by CodeRabbit

- **New Features**
  - Added privacy and compliance settings for consent providers, consent categories, legal policy links, and external consent-management platforms.
  - Added admin visibility and configuration for third-party services, consent categories, providers, and cookies.
  - Added built-in Klaro consent support and optional external CMP script integration.
  - Added privacy settings access and consent controls to the site footer.
  - Embedded content remains blocked until consent and is removed when consent is withdrawn.
- **Bug Fixes**
  - Improved protection against unsafe consent configuration content.
  - Invalid PayPal Connect endpoints are now rejected instead of silently defaulting.


Rachit7168 (changes requested):
Resolve Ai comments and merge conflicts !

mariobehling (approved):

sanketkokate712 (commented):
Fixed in 2536d2a3: `consent_config()` now uses `django.utils.html.json_script` instead of `mark_safe(json.dumps(...))`, which escapes `<`, `>` and `&` across the whole payload (so titles, purposes and the policy URLs are all covered). Covered by `test_service_title_cannot_break_out_of_the_config_script`, which feeds a `</script><script>…` title and asserts only one `</script>` survives and the value round-trips intact.

app/eventyay/control/forms/global_settings.py
```diff
@@ -21,21 +24,54 @@
...
 PAYPAL_CONNECT_ENDPOINT_CHOICES = (
-    ('live', _('Live')),
-    ('sandbox', _('Sandbox')),
+    (PAYPAL_CONNECT_ENDPOINT_LIVE, _('Live')),
+    (PAYPAL_CONNECT_ENDPOINT_SANDBOX, _('Sandbox')),
 )
 
+# Values accepted in addition to the choices above, kept for settings written
+# before the endpoint became a live/sandbox switch.
+PAYPAL_CONNECT_ENDPOINT_ALIASES = {
```
Rachit7168: In app/eventyay/control/forms/global_settings.py, there are several stray modifications related to PAYPAL_CONNECT_ENDPOINT_CHOICES and paypal_connect_endpoint_choice(). This appears to be an unintended inclusion (perhaps from a merge/rebase conflict or a dirty working tree). Since this PR is focused strictly on privacy and cookie consent, you should revert these PayPal-related changes to keep the git history clean.

app/eventyay/control/forms/global_settings.py
```diff
@@ -1051,3 +1102,57 @@ def __init__(self, *args, **kwargs):
...
+        self.fields['privacy_consent_provider'] = forms.ChoiceField(
+            label=_('Consent provider'),
+            choices=ConsentProvider.choices,
+            required=True,
+            help_text=_(
+                'Existing deployments stay on "Disabled" until an administrator turns consent on.'
+            ),
+        )
+
+    def clean(self):
```
Rachit7168, line 1122: You might want to also enforce the privacy_policy_url. Under GDPR/ePrivacy laws, consent banners generally need to link to the overarching Privacy Policy as well. If Klaro renders the banner, it should have a valid link to both.

app/eventyay/base/models/privacy.py
```diff
@@ -0,0 +1,117 @@
...
+        """__str__ method."""
+        return self.title
+
+    @property
+    def required(self):
+        """required method."""
+        return self.category == ConsentCategory.NECESSARY
+
+    @property
+    def cookie_name_list(self):
```
Rachit7168, line 102: This will output an array of strings in JSON, which Klaro will treat as exact matches. This works perfectly for exact cookie names. However, some services (like Google Analytics) generate dynamic cookies like _ga_XYZ123. Klaro supports regex for this, but JSON does not natively support JS RegExp objects. You don't need to fix this right now (exact matches are a great MVP), but you should keep it in mind if you receive bug reports about dynamic cookies not being cleared when consent is withdrawn.

app/eventyay/base/templatetags/privacy_consent.py
```diff
@@ -0,0 +1,104 @@
...
+    """consent_provider method."""
+    gs = GlobalSettingsObject()
+    return gs.settings.get('privacy_consent_provider') or ConsentProvider.DISABLED
+
+
+@register.simple_tag
+def external_cmp_script():
+    """external_cmp_script method."""
+    gs = GlobalSettingsObject()
+    if (gs.settings.get('privacy_consent_provider') or '') != ConsentProvider.EXTERNAL:
```
Rachit7168, line 80: The logic inside consent_embed (in privacy_consent.py) and embed_placeholder.html is extremely well done. Properly deferring to the external CMP when ConsentProvider.EXTERNAL is selected (by rendering the iframe directly and letting the external script handle the blocking) prevents dead UI states where a placeholder is rendered but no script exists to unblock it.

Rachit7168 (changes requested):
Thanks for working on this. Please address :

app/eventyay/control/forms/global_settings.py
sanketkokate712: Good catch, reverted in 2ec8bd300. The PAYPAL_CONNECT_ENDPOINT_* changes were real bug fixes (an unknown/malformed endpoint value was silently treated as "live", or crashed the form constructor) that got pulled onto this branch, but you're right that this PR should stay scoped to consent. Tracking that fix separately against dev instead.

app/eventyay/base/models/privacy.py
sanketkokate712: Noted, thanks. Exact-match cookie names are the MVP scope here on purpose; if we start seeing bug reports about dynamic cookie suffixes (like GA4's _ga_XYZ123) we'll look at teaching the config builder to emit Klaro-compatible regex patterns for those services.

app/eventyay/base/templatetags/privacy_consent.py
sanketkokate712: Thanks for the review!

app/eventyay/control/forms/global_settings.py
sanketkokate712: Good point, done in d5c2d4d18 — PrivacySettingsForm.clean() now requires privacy_policy_url the same way it already required privacy_cookie_policy_url, so Klaro can't be enabled without both policy links.

app/eventyay/base/migrations/0074_privacy_consent.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 6.0.7 on 2026-09-06 00:52
+
+import django.core.validators
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('base', '0073_loungemeshaccesstoken_loungemeshserver_and_more'),
```
Rachit7168: Same migration collision as last time, unfortunately. After the latest merge with `dev`, this still depends on `0073_…`, but `dev` already added `0074_bbbserver_disable_ssl_janusserver_disable_ssl_and_more` on that same parent. Two leaf nodes again → `manage.py migrate --noinput` will abort.

Please renumber this to `0075_privacy_consent` and depend on `0074_bbbserver_disable_ssl_janusserver_disable_ssl_and_more` (same dance you did for 0072/0073).

app/eventyay/control/views/global_settings.py
```diff
@@ -769,3 +814,49 @@ def post(self, request, *args, **kwargs):
...
+        gs = GlobalSettingsObject()
+        services = ThirdPartyService.objects.all()
+
+        context['services'] = services
+        context['provider'] = gs.settings.get('privacy_consent_provider', 'disabled')
+        # Surfaced as warnings on the overview so misconfiguration is visible
+        # rather than silently shipping a banner that blocks nothing.
+        #
+        # `category` is non-null and defaulted, so an enabled service is always
+        # classified. What an administrator can still get wrong is enabling a
```
Rachit7168: This comment is out of date now — `category` is `blank=True` with `default=''`, so services *can* be unclassified. The list below also pulls those in (empty string isn't in `enabled_categories`), but the template warning still talks about a category being "turned off". For an empty category that reads weird ("category () is turned off").

Maybe split unclassified vs category-disabled, or at least update the copy/comment so admins get a clear "please assign a category" message.

app/eventyay/control/templates/pretixcontrol/admin/privacy_settings.html
```diff
@@ -0,0 +1,87 @@
...
+
+    {% if missing_cookie_policy %}
+        <div class="alert alert-warning">
+            {% trans "No Cookie Policy URL is configured. The consent banner cannot be enabled until one is published." %}
+        </div>
+    {% endif %}
+    {% for service in unpublished_services %}
+        <div class="alert alert-warning">
+            {% blocktrans trimmed with name=service.title category=service.get_category_display %}
+                {{ name }} is enabled, but its consent category ({{ category }}) is turned off.
```
Rachit7168, line 31: Related to the view comment — when `category` is blank, `get_category_display` doesn't give a useful label, so this warning is confusing. Something like "enabled but has no consent category" for the empty case would be clearer.

app/eventyay/base/configurations/default_setting.py
```diff
@@ -3189,3 +3192,64 @@ def hex_color_field_config(label, default='', help_text=None, widget_class='colo
...
+# --- Privacy & Compliance -------------------------------------------------
+# Consent is opt-in: every optional category defaults to False so that no
+# non-essential script can run before the visitor has actively accepted it.
+DEFAULT_SETTINGS.update(
+    {
+        'privacy_consent_provider': {
+            'default': 'disabled',
+            'type': str,
+            'form_class': forms.ChoiceField,
+            'serializer_class': serializers.ChoiceField,
```
Rachit7168, line 3204: `ChoiceField` / `serializers.ChoiceField` without `choices` here. The admin form overrides it, so the UI is fine, but anything that builds fields from `DEFAULT_SETTINGS` (serializer path especially) will be missing the allowed values. Can you add `form_kwargs` / `serializer_kwargs` with `ConsentProvider.choices`?

app/eventyay/base/models/privacy.py
```diff
@@ -0,0 +1,117 @@
...
+    privacy_policy_url = models.CharField(
+        max_length=500,
+        blank=True,
+        validators=[URLValidator()],
+        verbose_name=_('Privacy policy URL'),
+    )
+    cookie_names = models.TextField(
+        blank=True,
+        verbose_name=_('Cookie names'),
+        help_text=_('One cookie name per line.'),
```
Rachit7168: Small UX note from earlier: Klaro treats these as exact string matches. That is fine for fixed names, but things like GA (`_ga_*`) need a regex pattern or they won't get cleared on withdraw. Worth a short help_text hint, e.g. that a line can be a Klaro cookie pattern / regex, not only a literal name.

app/eventyay/control/views/global_settings.py
```diff
@@ -769,3 +814,49 @@ def post(self, request, *args, **kwargs):
...
+        # config entirely, so the banner never mentions or blocks it.
+        enabled_categories = enabled_consent_categories(gs.settings)
+        context['unpublished_services'] = [
+            service
+            for service in services
+            if service.enabled
+            and not service.required
+            and service.category not in enabled_categories
+        ]
+        context['missing_cookie_policy'] = not gs.settings.get('privacy_cookie_policy_url')
```
Rachit7168, line 821: You now require both Cookie Policy *and* Privacy Policy in the form before Klaro can be enabled — nice. The overview warning still only checks the cookie URL though. Can we surface a missing Privacy Policy the same way, so the yellow banner matches what `clean()` enforces?

Rachit7168 (changes requested):
Thanks for pushing through the earlier feedback — the fail-closed config, `json_script` escaping, embed withdraw, and policy URL checks look solid.

One blocker after the latest `dev` merge though: the privacy migration collides with another `0074` again, which will break `migrate` in CI the same way as before. A couple of smaller follow-ups below as well.

sanketkokate712:
@Rachit7168 thanks for the review. All points from your latest review are fixed, and I merged the latest `dev` (it merged cleanly, with no conflicts).

**Migration collision (blocker), fixed in 814f5146**
- Renamed `0074_privacy_consent` to `0075_privacy_consent`. It now depends on `0074_bbbserver_disable_ssl_janusserver_disable_ssl_and_more`.
- `base` has a single leaf again (`0075_privacy_consent`), `detect_conflicts()` returns nothing, and `makemigrations --check` shows no changes for `base`.

**Blank category warning** (`views/global_settings.py`, `privacy_settings.html`)
- The overview now keeps two separate lists: `unclassified_services` and `category_disabled_services`.
- A service with no category now shows "*X is enabled but has no consent category*", so the "category () is turned off" message is gone. I also rewrote the outdated comment.

**Missing Privacy Policy warning**
- Added `missing_privacy_policy`. The yellow banners now match what `PrivacySettingsForm.clean()` enforces for both policy URLs.

**`privacy_consent_provider` choices** (`default_setting.py`)
- Added `choices=ConsentProvider.choices` to both `form_kwargs` and `serializer_kwargs`.

**Dynamic cookies** (`cookie_names` help text)
- I checked the Klaro bundle we ship. Any cookie string that starts with `^` is compiled into a `RegExp`, and other strings are escaped and matched exactly. So patterns already work without JS changes.
- The help text now says this and gives `^_ga_` for Google Analytics as an example. The migration has the same text.

**Unrelated changes from your earlier review**, fixed in 5ad129c1
- Reverted the one-line boilerplate docstrings that had been added to existing code outside this feature. That includes `clean_payment_paypal_connect_endpoint`.
- `test_global_settings_restructure.py` is now identical to `dev`, and the PR has no PayPal changes left. It's down to 17 files, all about privacy.

**Tests**
- Added tests for the pattern cookies, the setting choices, and the overview warnings (unclassified services, disabled categories, missing Privacy Policy).
- `test_privacy_consent.py`, `test_global_settings_test_email.py` and `test_admin_users.py` all pass (91 passed across the files I ran).
- 4 tests in `test_global_settings_restructure.py` also fail on `dev` with that file unchanged (`anti_abuse_provider` required, tab order, plugin payment field). This PR doesn't cause them.

app/eventyay/base/migrations/0074_privacy_consent.py
sanketkokate712: Migration collision (blocker), fixed in https://github.com/fossasia/eventyay/commit/814f51466c6dde6cc889484bfb1f3e0fc1ed4ec5

Renamed 0074_privacy_consent to 0075_privacy_consent. It now depends on 0074_bbbserver_disable_ssl_janusserver_disable_ssl_and_more.
base has a single leaf again (0075_privacy_consent), detect_conflicts() returns nothing, and makemigrations --check shows no changes for base.

app/eventyay/control/views/global_settings.py
sanketkokate712: Blank category warning (views/global_settings.py, privacy_settings.html)

The overview now keeps two separate lists: unclassified_services and category_disabled_services.
A service with no category now shows "X is enabled but has no consent category", so the "category () is turned off" message is gone. I also rewrote the outdated comment.

app/eventyay/control/templates/pretixcontrol/admin/privacy_settings.html
sanketkokate712: Blank category warning (views/global_settings.py, privacy_settings.html)

The overview now keeps two separate lists: unclassified_services and category_disabled_services.
A service with no category now shows "X is enabled but has no consent category", so the "category () is turned off" message is gone. I also rewrote the outdated comment.

app/eventyay/base/models/privacy.py
sanketkokate712: Dynamic cookies (cookie_names help text)

I checked the Klaro bundle we ship. Any cookie string that starts with ^ is compiled into a RegExp, and other strings are escaped and matched exactly. So patterns already work without JS changes.
The help text now says this and gives ^_ga_ for Google Analytics as an example. The migration has the same text.

Rachit7168 (commented):
Review notes on the Klaro consent foundation — focusing on places where the admin/config story and the actual visitor-facing behaviour diverge.

app/eventyay/base/templatetags/privacy_consent.py
```diff
@@ -0,0 +1,104 @@
...
+
+    enabled = enabled_consent_categories(settings)
+    services = ThirdPartyService.objects.filter(enabled=True)
+
+    return {
+        'elementID': 'klaro',
+        'storageMethod': 'cookie',
+        'cookieName': 'eventyay_consent',
+        'privacyPolicy': settings.get('privacy_policy_url') or '',
+        'cookiePolicy': settings.get('privacy_cookie_policy_url') or '',
```
Rachit7168: **[P1] Cookie Policy URL is required but never shown to visitors**

`PrivacySettingsForm.clean()` refuses to enable Klaro without `privacy_cookie_policy_url`, and this config emits `cookiePolicy` — but Klaro 0.7.22 only reads `privacyPolicy` (and translation `privacyPolicyUrl`) when building the banner policy link. There are zero references to `cookiePolicy` in the vendored Klaro bundle.

So admins must publish a Cookie Policy that the UI never links. Either surface it in the banner/modal (custom Klaro translation / extra link), or stop requiring/storing a value Klaro cannot use.

app/eventyay/base/templatetags/privacy_consent.py
```diff
@@ -0,0 +1,104 @@
...
+    third-party frames. Emitting a placeholder in those modes would leave the
+    content permanently unreachable.
+    """
+    gs = GlobalSettingsObject()
+    provider = gs.settings.get('privacy_consent_provider') or ConsentProvider.DISABLED
+    return {
+        'service': service,
+        'src': src,
+        'title': title,
+        'blocked': provider == ConsentProvider.KLARO,
```
Rachit7168: **[P1] Embeds stay permanently blocked when Klaro is on but Embed is off (or the service is missing)**

`blocked` is true whenever the provider is Klaro, ignoring `privacy_category_embed_enabled`. Placeholders call `manager.getConsent(serviceName)`. If Embed is disabled (or the named service was never registered / filtered out of the config), that service is absent → `getConsent` stays false → the iframe never appears.

That is the same deadlock the docstring carefully avoids for Disabled/External. Gate `blocked` on Klaro **and** an enabled Embed category (and ideally a published matching service), or fall back to a live iframe when Embed cannot be consented to.

app/eventyay/base/models/privacy.py
```diff
@@ -0,0 +1,122 @@
...
+    EXTERNAL = 'external', _('External CMP script')
+
+
+class ThirdPartyService(models.Model):
+    """
+    Admin-managed registry of third-party services and the consent category
+    each one belongs to.
+
+    The registry is what the frontend consent layer is built from: every
+    optional service listed here is blocked until its category is accepted.
```
Rachit7168: **[P1] Registering a service does not by itself block unmarked scripts**

The docstring / help text imply that classifying a service is enough to block it until consent. Klaro only activates/deactivates DOM nodes with `data-name="<service.name>"` (typically `type="text/plain"` + `data-type` / `data-src`). This PR never annotates existing app scripts and adds no loader that does.

Publishing a registry entry only affects the Klaro modal and cookie cleanup for listed names. Unmarked `<script>` / pixels keep running. Please either wire annotated script loading for registered services, or narrow this claim so reviewers/admins do not assume fail-closed blocking for trackers yet — today only `{% consent_embed %}` is actually gated.

app/eventyay/base/configurations/default_setting.py
```diff
@@ -3189,3 +3190,66 @@ def hex_color_field_config(label, default='', help_text=None, widget_class='colo
...
+            'default': '',
+            'type': str,
+            'form_class': forms.URLField,
+            'serializer_class': serializers.URLField,
+            'form_kwargs': dict(required=False, label=_('External CMP script URL')),
+        },
+        'privacy_policy_url': {
+            'default': '',
+            'type': str,
+            'form_class': forms.CharField,
```
Rachit7168: **[P2] Policy URL fields should be URL-validated**

`privacy_cmp_script_url` correctly uses `forms.URLField`, but `privacy_policy_url` / `privacy_cookie_policy_url` are plain `CharField`s. Klaro puts `privacyPolicy` into an `href`, so a non-http value (e.g. `javascript:…`) can end up in the public banner.

Prefer `URLField` (or an allowlist) and only pass http(s) URLs into the public config — same treatment as the CMP script URL above.

sanketkokate712:
@Rachit7168 thanks, all four points from your latest review are fixed in a035cc46. The branch is up to date with `dev`, has no conflicts, and there are still no new `base` migrations on `dev`, so `0075_privacy_consent` is still the only leaf.

**[P1] Cookie Policy URL never shown to visitors**
You were right: Klaro 0.7.22 has no `cookiePolicy` option. I dropped that key.
- Klaro's `htmlTexts` would have been the other option, but it sends every string through `dangerouslySetInnerHTML`, including service descriptions admins type in. Instead, `consent.js` adds a Cookie Policy link to the notice (after the notice text) and to the preference modal (after Klaro's "read our privacy policy" line). A `MutationObserver` puts the link back if Klaro re-renders.
- The URL and translated label come through the JSON config as `eventyayCookiePolicy`, and Klaro ignores unknown keys.
- The footer also shows a server-rendered Cookie Policy link next to "Privacy settings", so the link works even without JS.
- I tested this in headless Chrome with the vendored Klaro: the link appears in both the notice and the modal, with no console errors. The original `consent.js` shows no link.

**[P1] Embeds permanently blocked**
- Added `published_services(settings)`. The Klaro config builder and `consent_embed` now use the same check, so they can't drift apart.
- `blocked` is now true only when the provider is Klaro **and** the service is published in the config. If the Embed category is off, or the service is disabled or not registered, the iframe renders directly.
- I checked for a published service rather than only for the Embed category being on. A service registered under another enabled category can still be consented to, so it shouldn't fall back to a live iframe.

**[P1] Registering a service doesn't block unmarked scripts**
Agreed. The PR only gates `{% consent_embed %}`, so I narrowed the claims instead of implying scripts fail closed:
- the model docstring and the `name` help text (the migration has the same text)
- the settings comment
- a new note above the third-party services table on the admin page
- the PR description, which now says a script-gating tag (Klaro `data-name` markup) is a follow-up

**[P2] Policy URL fields should be URL-validated**
- `privacy_policy_url` and `privacy_cookie_policy_url` are now `forms.URLField` / `serializers.URLField`, so `javascript:…` is rejected on save.
- As a second layer, the public config and the footer only receive absolute http(s) URLs. Values stored some other way (`javascript:`, `data:`, relative paths) are left out, not rendered as links.

**Tests**
- Added 15 tests: embeds when the category is off, the service is disabled, or it isn't registered; the cookie policy link data and the missing `cookiePolicy` key; unsafe URLs never reaching the config or footer; `javascript:` rejected by the form; and the URLField definitions.
- 13 of them fail on the previous commit.
- `test_privacy_consent.py` has 42 passing, and 106 pass across the privacy and global-settings files I ran.
- `makemigrations --check` shows no changes.
- The same 4 `test_global_settings_restructure.py` failures from last time still happen, and that file is still identical to `dev`.

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/a0e3f118-64ed-43bc-a1ac-cb9b964d7987

Great base implementation

sanketkokate712:
> LGTM , Tested
> 
>  simplescreenrecorder-2026-09-18_11.59.52.mp4 
> Great base implementation

Thanks!!

Saksham-Sirohi (approved):
Consent config goes through json_script, bootstrap is an ES module, and policy URLs are filtered with public_http_url. Default provider stays off until an admin enables it.

Rachit7168:
@sanketkokate712 , Please resolve conflicts

sanketkokate712:
> @sanketkokate712 , Please resolve conflicts

working.

sanketkokate712:
> @sanketkokate712 , Please resolve conflicts

Done Resolved all Conflicts ,can i work on new up coming issues having some pr open.

PR #5486, opened 2026-09-07
fix(orga): guard feedback next redirects against open redirect

### LINKED ISSUE #5485: Open redirect in feedback-management views (unvalidated `next` parameter)
### Summary
Three redirects in the feedback-management views (`app/eventyay/orga/views/submission.py`) pass the user-supplied `next` query parameter straight into `redirect()` with no validation:

- `FeedbackBulkAction.post` — empty-selection branch and end of method
- `FeedbackUpdateStatus.get`

An authenticated organizer who follows a crafted link such as
`.../submissions/feedback/bulk/?next=https://evil.example.com`
is redirected to the external URL — an open redirect (CWE-601), usable for phishing.

### Expected
Only internal/same-host `next` URLs should be honored; external URLs should be ignored.

### Note
The sibling method `FeedbackUpdateStatus.post` in the same file already guards this correctly with `url_has_allowed_host_and_scheme(...)` — the other three paths just missed it. So the fix is to apply that same check.

### Severity
Low-to-medium: the endpoints require an authenticated organizer with `base.orga_update_submission`, so this is an authenticated open redirect rather than an unauthenticated one.

### Environment
- Branch: `dev`

https://github.com/user-attachments/assets/f383b076-d241-421c-9718-3c0ae1e69d0b

Rishika3D:
## Summary
Three redirects in the feedback-management views (`orga/views/submission.py`) passed the user-supplied `next` query parameter straight into `redirect()` with no validation:
- `FeedbackBulkAction.post` (empty-selection branch and end of method)
- `FeedbackUpdateStatus.get`

An authenticated organizer following a crafted link like `.../submissions/feedback/bulk/?next=https://evil.example.com` was redirected to the external URL — an open redirect (CWE-601), usable for phishing.

The sibling method `FeedbackUpdateStatus.post` in the same file already guards this with `url_has_allowed_host_and_scheme(...)`. This PR applies the same allow-list check to the other three paths, so an external `next` is ignored (falling back to the event feedback page) while an internal `next` still works.

## Severity
Low-to-medium: the endpoints require an authenticated organizer with `base.orga_update_submission`.

## Proof / test
Adds `tests/talk/orga/views/test_orga_views_feedback_redirect.py` with two cases:
- Against the original code the block test FAILS — the response is `302 Location: https://evil.example.com/phish`.
- Against this fix both tests PASS — the external `next` is rejected, and a safe internal `next` is still honored.

Video attached below.

https://github.com/user-attachments/assets/98f51a0a-43b4-4a08-90b7-196823ce07bb

Fixes #5485

## Summary by Sourcery

Guard feedback-management redirects against open redirects while preserving safe internal navigation.

Bug Fixes:
- Prevent unauthenticated external redirects from feedback-management views by rejecting unsafe user-supplied `next` URLs and falling back to the event feedback page.

Tests:
- Add coverage verifying that external feedback redirect targets are blocked while safe internal targets remain supported.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved redirect safety in feedback workflows by blocking untrusted external destinations.
  - Safe internal redirect destinations continue to work as expected.
  - Secure connections now require secure redirect destinations.
  - When a destination is missing or invalid, users are redirected to the event feedback page.
  - Redirect behavior is now consistent across bulk feedback actions and individual feedback status updates.


Saksham-Sirohi (approved):
Open redirect is closed the same way as `FeedbackUpdateStatus.post`: `url_has_allowed_host_and_scheme(..., allowed_hosts=None)` keeps relative/`next` working and rejects external hosts. Tests cover the bulk path.

app/eventyay/orga/views/submission.py
```diff
@@ -1273,7 +1273,10 @@ def post(self, request, *args, **kwargs):
         
         if not feedback_ids:
             messages.warning(request, _('No items selected.'))
-            return redirect(request.GET.get('next', request.event.orga_urls.feedback))
+            next_url = request.GET.get('next')
+            if next_url and url_has_allowed_host_and_scheme(next_url, allowed_hosts=None):
+                return redirect(next_url)
```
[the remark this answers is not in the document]
Saksham-Sirohi: @Rishika3D i think thus can be addressed

Saksham-Sirohi (changes requested):

mariobehling (approved):

Rachit7168 (approved):
LGTM , Tested

app/eventyay/orga/views/submission.py
```diff
@@ -1273,7 +1273,10 @@ def post(self, request, *args, **kwargs):
         
         if not feedback_ids:
             messages.warning(request, _('No items selected.'))
-            return redirect(request.GET.get('next', request.event.orga_urls.feedback))
+            next_url = request.GET.get('next')
+            if next_url and url_has_allowed_host_and_scheme(next_url, allowed_hosts={request.get_host()}, require_https=request.is_secure()):
```
Saksham-Sirohi, line 1277: Nit: this allow-list check is copy-pasted four times in this class pair. A tiny helper (e.g. `_feedback_next_redirect(request)`) would keep the four call sites from drifting if the policy changes again.

app/tests/talk/orga/views/test_orga_views_feedback_redirect.py
```diff
@@ -0,0 +1,45 @@
...
+
+@pytest.mark.django_db
+def test_feedback_update_status_get_open_redirect_is_blocked(orga_client, event, submission):
+    """FeedbackUpdateStatus.get: external `next` falls back to the feedback page."""
+    with scope(event=event):
+        feedback = Feedback.objects.create(talk=submission, review='ok')
+    url = event.orga_urls.feedback + f'{feedback.pk}/action/?action=noop&next=' + EXTERNAL_NEXT
+    response = orga_client.get(url)
+    assert response.status_code == 302
+    assert response['Location'] == event.orga_urls.feedback
```
Saksham-Sirohi, line 45: Nit (coverage): add one case for `FeedbackUpdateStatus.post` with an external `next`, and optionally an absolute same-host `next` (`https://{testserver}...`) so the `allowed_hosts={request.get_host()}` change stays locked in. Not required to merge.

Saksham-Sirohi (approved):
Open redirect is closed correctly on all four feedback paths: `allowed_hosts={request.get_host()}` + `require_https=request.is_secure()` matches the stronger pattern used elsewhere (e.g. team views), and external `next` falls back to `orga_urls.feedback`. Regression coverage for empty bulk, completion bulk, and `FeedbackUpdateStatus.get` looks good.

Non-blocking nits below. (Same-file `allowed_hosts=None` at the submission success/pending-state redirects is out of scope here—worth a follow-up, not this PR.)

app/eventyay/orga/views/submission.py
Saksham-Sirohi: would help dry changes up does not block us

PR #5491, opened 2026-09-07
fix(cfp): preserve uploaded files after back navigation

### LINKED ISSUE #5443: PDF slides disappear after navigating back in CFP submission wizard
### Describe the bug

When uploading a PDF in the "Slides" field of the CFP submission wizard,
the PDF disappears from the form when navigating back to the previous step.

The PDF appears to be uploaded successfully, but it is no longer displayed
when returning to the form.

### Steps to reproduce

1. Open a CFP submission form.
2. Upload a PDF in the "Slides" field.
3. Click "Next".
4. Click "Back".
5. Observe the "Slides" field.

### Actual behavior

The uploaded PDF is no longer displayed in the Slides field.

The session image remains visible.

I tested this with PDFs of different sizes, including a ~100 KB PDF,
so the issue does not appear to be related to the upload size.

### Expected behavior

When navigating back, the previously uploaded PDF should still be displayed
in the Slides field.

### Additional context

The upload limit is 10 MB.

The browser also displays the warning:

> This file is too large to be saved in the browser's temporary storage. If the form has other errors, you will need to re-select it before submitting.

However, the PDF disappears even when using a very small PDF.

The PDF appears to be saved successfully on the server, so the issue may be
related to restoring or displaying the existing PDF when navigating back.

### Screenshots

Image

SawanTeja:
### Description

Closes #5443

**Bugs Fixed**

* **Disappearing PDFs on Back Navigation:** Fixed an issue where uploaded PDFs in the "Slides" field or other custom file fields disappeared when navigating to the next step and clicking "Back".
* **Remove Button Not Working:** Fixed an issue where the remove button did not correctly remove uploaded files from the form.
* **Incorrect "File Too Large" Warning:** Fixed a UI bug where the browser temporary-storage warning was always displayed, even when no file was uploaded.
* **Upload Limit Text Always Visible:** Fixed an issue where the red upload-limit warning was permanently visible on a fresh form instead of only appearing when the limit was exceeded.
* **Lost Files on Form Validation Errors:** Fixed an issue where uploaded files disappeared from the UI when form submission failed validation due to another field.
* **Temporary File Removal Error:** Fixed an error that could occur when removing an unsubmitted session file with a temporary string ID.

**Technical Changes**

* Updated `FormFlowStep.get_form_initial` in `flow.py` to correctly restore single-file and multiple-file session uploads.
* Updated `get_files` in `flow.py` to respect files explicitly cleared by the user.
* Updated `InfoForm.__init__` in `submission.py` to include active session files alongside existing database resources.
* Updated `SlidesWidget.get_context` in `widgets.py` to use existing resources reliably during validation cycles.
* Added validation in `save_slides_resource` to safely handle temporary session file IDs.
* Updated `slides_input.html` to hide upload warnings by default and display them only when triggered by the client-side validation logic.

https://github.com/user-attachments/assets/bacb7739-e953-43cf-b9bf-76abab539177


app/eventyay/cfp/flow.py
```diff
@@ -277,23 +295,29 @@ def get_form(self, from_storage=False):
...
 
         # Preserve MultiValueDict semantics for proper multi-file field support
         files = MultiValueDict()
         # Add session files first
-        for field, file_obj in session_files.items():
-            files[field] = file_obj
+        for field, file_list in session_files.lists():
+            files.setlist(field, file_list)
         # For each field, new uploads completely replace any existing session files
         for field, file_list in self.request.FILES.lists():
```
Rachit7168, line 317: There is a critical bug here when dealing with multi-file uploads (like `slides_files`). 
If a user previously uploaded `slide1.pdf` and `slide2.pdf` (which are in `session_files`), and then decides to upload `slide3.pdf`, standard HTML multi-file inputs only submit the new file (`slide3.pdf`) in `request.FILES`.

Because you are using `files.setlist(field, file_list)`, it **completely overwrites** the session files. `form.files` will now only contain `slide3.pdf`. 

This causes two cascading failures:
1. When `self.set_files(form.files)` runs, it will overwrite the CFP session with *only* `slide3.pdf`. `slide1` and `slide2` are permanently lost.
2. In `submission.py`, `save_slides_resource()` only creates DB records for files in `slides.resources` (the new uploads). Session files that got pushed into `kept_existing_resources` are ignored and never saved to the DB.

**Recommendation:** 
You should use `files.appendlist` to merge the new uploads with the session files, rather than overwriting them. However, if you do this, you must also ensure `SlidesField.clean()` doesn't double-count them (since `submission.py` also injects session files into `existing_resources` for UI rendering).

app/eventyay/submission/forms/resource.py
```diff
@@ -54,8 +54,9 @@ def get_slides_max_count(event) -> int:
 
 
 def save_slides_resource(submission, slides: SlidesData):
-    if slides.clear_ids:
-        delete_slide_resources(submission, resource_ids=slides.clear_ids)
+    valid_clear_ids = [cid for cid in slides.clear_ids if cid.isdigit()]
```
Rachit7168, line 55: This function only creates DB records for `slides.resources` (which correspond to newly submitted files in `request.FILES`).

During the CFP wizard, temporary files are kept in the session. Because they are injected into `existing_resources` by the view to render the UI, they end up inside `slides.kept_existing_resources` after validation, not `slides.resources`. 

Because this loop ignores `kept_existing_resources`, any session files that the user uploaded in previous steps will **never be saved to the database** when the CFP is finalized. 

**Recommendation:**
You need to also iterate over `slides.kept_existing_resources` and check if they are temporary session files (e.g., checking if `str(resource.pk).startswith('tmp:')`). If they are, you need to extract their `url`/`filename` and create the `Resource` in the DB just like you do for `slides.resources`.

Rachit7168 (changes requested):
Please address : 

app/eventyay/submission/forms/resource.py. Inside SlidesField.clean (where total_count is evaluated)

Because `submission.py` explicitly injects `session_slides` into `self.existing_resources` (so they render in the UI), and the session files *also* exist in `form.files` (if the user hasn't overwritten them), they will end up in BOTH `kept_existing_resources` and `cleaned_resources`.

This means `slides_data.total_count` will count session files **twice**. If the limit is 3, and the user has 2 session files, it will calculate a count of 4 and throw a `ValidationError: You can add at most 3 slide entries.`

**Recommendation:**
Filter out the duplicates in `SlidesField.clean` by ensuring that items aren't pushed into `cleaned_resources` if they are already represented as a temporary session file in `kept_existing_resources`.

mariobehling (approved):

Rachit7168 (approved):
LGTM , Tested

https://github.com/user-attachments/assets/04c7a59e-919a-4b68-bbd6-b93968244a4a

Saksham-Sirohi (approved):

PR #5492, opened 2026-09-07
Add Admin Message Center for platform-wide emails

### LINKED ISSUE #5346: Implement platform admin message center for communication with organisers and registered users
Implement a new **Admin Message Center** that allows platform admins to communicate with organisers and registered users from the platform admin area.

The workflow should follow the improved message center approach already used in the Tickets and Talks components:

* one clear compose page
* audience filters on the same page
* recipient preview
* message editor
* test email
* draft support
* outbox/sent handling
* templates section

This feature is platform-wide and not limited to one event.

## Goal

Platform admins should be able to send platform-level emails to selected users and groups, for example:

* all registered users
* all organisers
* organisers of selected events
* users with specific roles
* users with incomplete setup
* users with upcoming events
* users related to events in a specific status
* selected individual users

Image

Image
Image
Image

## Proposed sidebar structure

Add a new item in the admin sidebar:

```text
Admin
- Message center
  - Outbox
  - Compose
  - Drafts
  - Sent
  - Templates
```

The admin message center should be clearly separated from event-level message centers.

## Compose page

Create a platform admin compose page similar to the Tickets and Talks message center.

Suggested route:

```text
/admin/messages/compose/
```

Suggested page title:

```text
Compose platform email
```

Subtitle:

```text
Send emails to organisers, registered users and selected platform groups.
```

## Compose page structure

Use the same section-based layout as the improved event email composer:

```text
Audience
Reply-To and BCC
Message
Attachment
Delivery
Email preview
Action row
```

Do not use separate tabs for Recipients, Content and Scheduling.

## Audience section

The audience section should allow platform admins to define recipients using filters.

Suggested recipient groups:

```text
Recipient group
- All registered users
- All organisers
- Organisers of selected events
- Event team members
- Speakers
- Reviewers
- Attendees
- Selected users
```

## Suggested filters

Add filters depending on the selected recipient group.

### User filters

```text
Account status
- Verified
- Unverified
- Active
- Banned
- Deleted / anonymized should be excluded by default

User role
- User
- Organiser
- Staff
- Platform admin

Created after
Created before

Last active after
Last active before

Language / locale
```

### Organiser filters

```text
Organiser account
- All organisers
- Selected organisers

Organiser status
- Has active events
- Has draft events
- Has past events
- Has no events

Billing status
- Billing configured
- Billing missing
- Billing validation pending
```

### Event-related filters

```text
Event
- Selected events
- Events by selected organisers

Event status
- Live
- Draft
- Past

Event date range
- Starts after
- Starts before
- Ends after
- Ends before

Event setup status
- Missing ticket setup
- Missing payment setup
- Missing schedule
- Missing video setup

Ticketing status
- Shop enabled
- Shop disabled
- Has paid tickets
- Has free tickets

CfP status
- CfP open
- CfP closed
- Has pending proposals
```

### Recipient controls

Add:

```text
Show recipient list
Clear filters
Exclude platform admins
Exclude inactive users
Exclude users without confirmed email
```

The recipient count should update when filters change.

Example:

```text
284 recipients
```

## Recipient preview

Add **Show recipient list**.

The recipient preview should show:

```text
Name
Email
Account status
Role
Organiser / Event relation
Reason included
```

Requirements:

* de-duplicate recipients
* show why a user is included
* exclude users without valid email addresses from final delivery
* show warnings for skipped recipients

Example:

```text
12 users skipped because they do not have a confirmed email address.
```

## Reply-To and BCC

Add a section directly below Audience:

```text
Reply-To and BCC
```

Fields:

```text
Reply-To
BCC
```

Reply-To should default to the platform sender address if empty.

BCC should support comma-separated addresses.

## Message section

Use the same message editor approach as the other message center pages.

Fields:

```text
Subject *
Message *
```

Support:

* rich text editor
* language versions if platform emails support localization
* placeholder insertion
* preview mode

The message section should include:

```text
Insert placeholder
Placeholder help
```

## Placeholders

Use an **Insert placeholder** button instead of showing all placeholders permanently.

The placeholder drawer/popover should be searchable and grouped.

Suggested groups:

### User

```text
{user_name}
{first_name}
{last_name}
{email}
{account_url}
```

### Organiser

```text
{organiser_name}
{organiser_url}
{organiser_billing_url}
```

### Event

```text
{event_name}
{event_url}
{event_dashboard_url}
{event_start_date}
{event_end_date}
```

### Platform

```text
{platform_name}
{platform_url}
{support_email}
{support_url}
```

Only show placeholders that are valid for the selected recipient group.

## Attachment section

Add optional attachment support if already available in the event message center.

```text
Attachment
[Choose file]
```

Show the same file size and file type restrictions as the existing message center.

## Delivery section

Add delivery options:

```text
Send now
Schedule for later
```

If **Schedule for later** is selected, show:

```text
Date
Time
Timezone
```

Use platform timezone or explicitly show the timezone used.

Add:

```text
Send test email to
[Input] [Send test email]
```

Test email validation should be inline.

If the field is empty, show the error next to the field, not as a global page alert.

## Final action row

Use the same action row pattern as the improved composers:

```text
284 recipients selected     [Save draft] [Preview email] [Send to outbox]
```

If **Send immediately** is supported, place it in the final action row:

```text
284 recipients selected     [ ] Send immediately     [Save draft] [Preview email] [Send to outbox]
```

When **Send immediately** is checked, the button label should change to:

```text
Send email
```

Default button label:

```text
Send to outbox
```

## Outbox

Add an admin message center outbox.

The outbox should show queued platform emails.

Suggested columns:

```text
Subject
Audience
Recipients
Created by
Scheduled for
Status
Created at
Actions
```

Actions:

```text
Preview
Send now
Edit
Cancel
Delete draft / queued email
```

## Drafts

Add draft support.

Drafts should save:

* selected recipient group
* all audience filters
* recipient count at time of saving
* subject
* message body
* language versions
* placeholders
* Reply-To
* BCC
* attachments metadata if supported
* delivery settings
* scheduled date/time

When reopening a draft, recalculate the recipient count and show a notice if it changed.

Example:

```text
Recipient count changed from 284 to 271 because platform data changed since this draft was saved.
```

## Sent

Add a sent email list.

Suggested columns:

```text
Subject
Audience
Recipients
Sent by
Sent at
Status
Actions
```

Actions:

```text
View email
View recipient list
Duplicate
```

The sent list should include emails sent from the admin message center.

## Templates section

Add a **Templates** section in the admin message center.

At this stage, templates should be **read-only**.

The Templates page should list all platform email templates, but must not allow editing yet.

Suggested columns:

```text
Template
Category
Subject
Trigger
Recipient
Status
Actions
```

Actions:

```text
View
Preview
```

Do not show edit buttons at this stage.

## Platform email templates to list

The Templates section should list platform/system emails such as:

```text
Account registration
Email confirmation
Password reset
Account notification emails
Organiser invitation
Event team invitation
Billing validation emails
Platform fee / organiser billing emails
Ticket order emails
Ticket confirmation emails
Ticket cancellation emails
Refund notification emails
CfP submission confirmation
Proposal acceptance email
Proposal rejection email
Speaker schedule update
Reviewer/team member notification emails
Video/event notification emails
System notification emails
```

The exact list should be generated from the existing platform email template registry.

## Template detail view

Clicking a template should open a read-only detail page or drawer.

Show:

```text
Template name
Category
Trigger
Recipient type
Subject
Body preview
Available placeholders
Last updated
Source / system identifier
```

No editing in this issue.

## Permissions

Only platform admins should access the admin message center.

Requirements:

* non-admin users must not access admin message center routes
* organisers must not access platform-wide messaging
* all send actions must be permission-checked server-side
* all message actions should be logged

## Audit logging

Record:

```text
Created by
Updated by
Sent by
Sent at
Recipient filters used
Recipient count
Send mode
Scheduled time
```

This is needed because platform-wide emails can affect many users.

## Safety checks

Before sending, show a confirmation if the recipient count is high.

Example:

```text
You are about to send this email to 284 recipients.

[Cancel] [Send to outbox]
```

For immediate sending:

```text
You are about to send this email immediately to 284 recipients.

[Cancel] [Send email]
```

## Required changes

* Add admin sidebar item **Message center**.
* Add admin message center routes.
* Add platform-wide compose page.
* Add audience filters for organisers, users and event-related groups.
* Add recipient count and recipient preview.
* Add Reply-To and BCC.
* Add message editor with placeholders.
* Add test email support with inline validation.
* Add delivery options.
* Add Save draft support.
* Add Outbox, Drafts and Sent pages.
* Add read-only Templates page.
* List all existing platform email templates.
* Add permission checks.
* Add audit logging.
* Add tests for filters, drafts, preview, outbox and sending.

## Acceptance criteria

* Platform admins can access an admin **Message center**.
* Non-admin users cannot access it.
* Admins can compose platform-wide emails.
* Admins can select recipients using audience filters.
* Recipient count updates based on filters.
* Admins can preview the recipient list.
* Recipients are de-duplicated.
* Users without valid email addresses are handled safely.
* Admins can enter Reply-To and BCC.
* Admins can compose subject and message body.
* Admins can insert valid placeholders.
* Admins can send a test email.
* Missing test email address shows inline validation.
* Admins can save drafts.
* Drafts preserve filters and message content.
* Admins can send emails to the outbox.
* Admins can send immediately if the option is enabled.
* Sent emails appear in the admin message center Sent section.
* Templates page lists all platform email templates.
* Templates are read-only in this issue.
* Template detail view shows subject, body preview, trigger, recipient type and placeholders.
* All send actions are permission-checked.
* Admin message actions are audit-logged.
* The UI follows the improved Tickets and Talks message center layout.

MukundC25:
Closes #5346 | Part 1

>  Templates are currently read-only as per issue, can be improved or made editable in a follow-up

#### Summary

Adds a new **Admin Message Center** allowing platform admins to compose and send emails to organisers, registered users, and selected platform groups from `/admin/messages/`. Follows the same section-based compose layout already used in Tickets and Talks message centers, but operates platform-wide (not event-scoped).

#### Models (`base/models/admin_mail.py`)
- `AdminEmailQueue` — platform-wide email with status tracking (draft/queued/sending/sent/cancelled)
- `AdminEmailQueueRecipient` — per-recipient send tracking with error handling
- `AdminEmailQueueFilter` — persists audience filter state for drafts
- Migration: `0071_admin_message_center`

#### Compose page (`/admin/messages/compose/`)
Section-based layout matching the existing event-level composers:
- **Audience** — 8 recipient groups (all users, all organisers, event organisers, team members, speakers, reviewers, attendees, selected users) with filters for account status, user role, language, event status, date ranges, and exclusion checkboxes
- **Reply-To and BCC**
- **Message** with placeholder support (user, platform groups)
- **Attachment** upload
- **Delivery** — send now, schedule for later, test email with inline validation
- **Action row** — save draft, preview email, send to outbox / send immediately

#### List pages
- **Outbox** — queued emails with send/edit/cancel actions
- **Drafts** — saved drafts with edit/duplicate/delete actions
- **Sent** — sent emails with duplicate action
- **Templates** — read-only list of all platform email templates (system + MailTemplate roles)
- **Template detail** — read-only view of individual template

#### Infrastructure
- AJAX endpoint for live recipient count and preview list
- Recipient resolution with deduplication across all filter combinations
- Celery task (`send_admin_email`) with `select_for_update(skip_locked=True)` and retry logic
- Audit logging via `LogEntry`
- Admin sidebar entry (envelope icon) with Outbox, Compose, Drafts, Sent, Templates children

#### Permissions
- All views use `AdministratorPermissionRequiredMixin` (requires active staff session)
- Non-admin users cannot access any message center routes

#### Files changed (14)

| File | Change |
|------|--------|
| `base/models/admin_mail.py` | **New** — 3 models + enums |
| `base/models/__init__.py` | Register new models |
| `base/migrations/0071_admin_message_center.py` | **New** — migration |
| `control/forms/admin/admin_messages.py` | **New** — compose + recipients forms |
| `control/views/admin_messages.py` | **New** — 11 views + recipient resolution |
| `control/urls_admin.py` | Add 11 URL patterns |
| `control/navigation.py` | Add Message Center sidebar nav |
| `control/tasks.py` | Add `send_admin_email` Celery task |
| `admin/messages/compose.html` | **New** — compose template |
| `admin/messages/outbox.html` | **New** — outbox list |
| `admin/messages/drafts.html` | **New** — drafts list |
| `admin/messages/sent.html` | **New** — sent list |
| `admin/messages/templates.html` | **New** — templates list |
| `admin/messages/template_detail.html` | **New** — template detail |

#### Screen rec

https://github.com/user-attachments/assets/1d6f3066-4ea0-4fd6-a94f-b7ec78e3b06b

## Summary by Sourcery

Enable platform administrators to compose, manage, and deliver targeted platform-wide email campaigns through a new Message Center.

New Features:
- Add an administrator-only platform-wide message center for composing emails to configurable user, organiser, event, speaker, reviewer, attendee, and selected-user audiences.
- Provide audience filtering, recipient previews, email previews, placeholders, test messages, attachments, immediate delivery, scheduling, and high-recipient confirmation.
- Add outbox, drafts, sent-mail, and read-only template views with management actions such as editing, cancelling, deleting, and duplicating messages.

Enhancements:
- Track email queues, filters, per-recipient delivery state, failures, and deduplicated recipients with platform-wide persistence.
- Integrate asynchronous email delivery with scheduling, retries, concurrency protection, HTML sanitisation, and audit logging.
- Add the Message Center to the administrator navigation and protect its routes with administrator permissions.

Tests:
- Add coverage for email queue behaviour, recipient resolution and deduplication, permissions, previews, recipient APIs, duplication, filters, and audit records.

## Summary by CodeRabbit

- **New Features**
  - Added an Admin Message Center with Outbox, Compose, Drafts, Sent, and Email Templates sections.
  - Administrators can compose, preview, schedule, send, cancel, duplicate, and delete messages.
  - Added recipient filtering by audience, events, roles, language, account activity, and status.
  - Added draft saving, attachments, reply-to and BCC support, delivery options, and placeholder assistance.
  - Added recipient previews, delivery confirmations, and safeguards for large audiences.

- **Bug Fixes**
  - Improved email validation, HTML sanitization, attachment handling, and delivery retry behavior.


app/eventyay/control/forms/admin/admin_messages.py
```diff
@@ -0,0 +1,336 @@
...
+        widget=SplitDateTimePickerWidget(),
+        label=_('Schedule for later'),
+        required=False,
+        help_text=_('Leave empty to add to outbox. If set, the email will be sent at this time.'),
+    )
+
+    send_immediately = forms.BooleanField(
+        label=_('Send immediately'),
+        required=False,
+        help_text=_('If checked, the email will be sent immediately instead of being added to the outbox.'),
```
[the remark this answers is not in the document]
MukundC25: Fixed. `send_immediately` is now rendered as an inline checkbox in the action row at the bottom of the compose page

app/eventyay/control/views/admin_messages.py
```diff
@@ -0,0 +1,789 @@
...
+                        'user_role': f.user_role,
+                        'language': f.language,
+                        'event_status': f.event_status,
+                        'created_after': f.created_after,
+                        'created_before': f.created_before,
+                        'last_active_after': f.last_active_after,
+                        'last_active_before': f.last_active_before,
+                        'exclude_admins': f.exclude_admins,
+                        'exclude_inactive': f.exclude_inactive,
+                        'exclude_unconfirmed_email': f.exclude_unconfirmed_email,
```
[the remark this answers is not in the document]
MukundC25: Fixed

app/eventyay/control/tasks.py
```diff
@@ -87,3 +90,54 @@ def delete_organizer_data(organizer_id: int, user_id: int | None = None) -> None
...
+@scopes_disabled()
+def send_admin_email(self, admin_email_id: int) -> None:
+    try:
+        with transaction.atomic():
+            mail = (
+                AdminEmailQueue.objects
+                .select_for_update(skip_locked=True)
+                .filter(pk=admin_email_id)
+                .exclude(status__in=[AdminEmailStatus.SENT, AdminEmailStatus.DRAFT, AdminEmailStatus.CANCELLED])
+                .first()
```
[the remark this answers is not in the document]
MukundC25: Fixed. `AdminMessageCancelView` now uses `select_for_update()` inside `transaction.atomic()`, locking the row before checking and updating status

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,344 @@
...
+                        'html': AdminEmailQueue.make_html(body),
+                        'reply_to': [reply_to_addr] if reply_to_addr else [],
+                        'event': None,
+                        'cc': [],
+                        'bcc': bcc_list,
+                        'attachments': attachments,
+                    },
+                    ignore_result=True,
+                )
+                recipient.sent = True
```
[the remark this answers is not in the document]
MukundC25: out of scope, can be a followup

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,344 @@
...
+                recipient.error = None
+                recipient.save(update_fields=['sent', 'error'])
+            except Exception as exc:
+                logger.exception('Error sending admin email to %s', recipient.email)
+                recipient.error = str(exc)
+                recipient.save(update_fields=['error'])
+
```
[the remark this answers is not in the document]
MukundC25: Fixed. After the per-recipient dispatch loop, `send()` checks for any remaining unsent valid recipients. If any failed dispatch, it resets the queue status

app/eventyay/control/views/admin_messages.py
```diff
@@ -0,0 +1,789 @@
...
+                ).values_list('pk', flat=True)
+            else:
+                event_pks = None
+
+            if event_pks is not None:
+                qs = qs.filter(
+                    Q(teams__all_events=True, teams__organizer__events__pk__in=event_pks)
+                    | Q(teams__limit_events__pk__in=event_pks)
+                ).distinct()
+
```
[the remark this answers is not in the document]
MukundC25: Fixed. The event-status narrowing now branches by recipient group: SPEAKERS are filtered via `Submission.event__pk__in`, ATTENDEES via `OrderPosition/Order` subqueries. Only ORGANISERS, TEAM_MEMBERS, and REVIEWERS use the team-relationship filter.

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,344 @@
...
+                        'event': None,
+                        'cc': [],
+                        'bcc': bcc_list,
+                        'attachments': attachments,
+                    },
+                    ignore_result=True,
+                )
+                recipient.sent = True
+                recipient.error = None
+                recipient.save(update_fields=['sent', 'error'])
```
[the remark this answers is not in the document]
MukundC25: Fixed

app/eventyay/control/views/admin_messages.py
```diff
@@ -0,0 +1,789 @@
...
+                    'scheduled_at': draft.scheduled_at,
+                    'delivery_mode': 'later' if draft.scheduled_at else 'now',
+                })
+                try:
+                    f = draft.filters
+                    initial.update({
+                        'account_status': f.account_status,
+                        'user_role': f.user_role,
+                        'language': f.language,
+                        'event_status': f.event_status,
```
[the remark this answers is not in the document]
MukundC25: fixed

app/eventyay/control/views/admin_messages.py
```diff
@@ -0,0 +1,789 @@
...
+        kwargs['draft_save'] = self.request.POST.get('action') == 'draft'
+        return kwargs
+
+    def get_initial(self):
+        initial = super().get_initial()
+        draft_pk = self.request.GET.get('draft')
+        if draft_pk:
+            draft = AdminEmailQueue.objects.filter(
+                pk=draft_pk, status=AdminEmailStatus.DRAFT
+            ).first()
```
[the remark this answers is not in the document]
MukundC25: Fixed

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,344 @@
...
+                    kwargs={
+                        'to': [recipient.email],
+                        'subject': subject,
+                        'body': body,
+                        'html': AdminEmailQueue.make_html(body),
+                        'reply_to': [reply_to_addr] if reply_to_addr else [],
+                        'event': None,
+                        'cc': [],
+                        'bcc': bcc_list,
+                        'attachments': attachments,
```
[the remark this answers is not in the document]
MukundC25: out of scope

app/eventyay/control/forms/admin/admin_messages.py
```diff
@@ -0,0 +1,336 @@
...
+        required=False,
+        widget=forms.EmailInput(attrs={'placeholder': 'name@domain.com'}),
+    )
+
+    delivery_mode = forms.ChoiceField(
+        label=_('Delivery mode'),
+        choices=DELIVERY_MODE_CHOICES,
+        initial='now',
+        required=False,
+        widget=forms.RadioSelect(),
```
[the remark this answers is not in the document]
MukundC25: The naming is intentional

app/eventyay/control/views/admin_messages.py
```diff
@@ -0,0 +1,1081 @@
...
+
+def _save_filters(mail: AdminEmailQueue, filters: dict) -> AdminEmailQueueFilter:
+    obj, _created = AdminEmailQueueFilter.objects.update_or_create(
+        mail=mail,
+        defaults={
+            'account_status': filters.get('account_status', ''),
+            'user_role': filters.get('user_role', ''),
+            'language': filters.get('language', ''),
+            'created_after': filters.get('created_after'),
+            'created_before': filters.get('created_before'),
```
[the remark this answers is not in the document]
MukundC25: Fixed. `_save_filters` now persists all fields, Reopening a draft restores all from the saved filter record

app/eventyay/control/views/admin_messages.py
```diff
@@ -0,0 +1,1081 @@
...
+            if selected_event_ids:
+                qs = qs.filter(
+                    Q(teams__all_events=True, teams__organizer__events__pk__in=selected_event_ids)
+                    | Q(teams__limit_events__pk__in=selected_event_ids)
+                ).distinct()
+
+        elif recipient_group == AdminRecipientGroup.ATTENDEES:
+            pos_qs = OrderPosition.objects.filter(
+                attendee_email__isnull=False
+            ).exclude(attendee_email='')
```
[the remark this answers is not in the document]
MukundC25: Fixed. `pos_qs` now includes `order__status__in=['p', 'n']`, matching the `order_qs` restriction, Cancelled/expired order attendees are excluded from the recipient list.

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,385 @@
...
+                exclude_inactive=filters.exclude_inactive,
+                exclude_unconfirmed_email=filters.exclude_unconfirmed_email,
+            )
+        return new_mail
+
+    def _resolve_attachment(self) -> list[dict] | None:
+        if not self.attachment:
+            return None
+        try:
+            cf = CachedFile.objects.get(id=self.attachment)
```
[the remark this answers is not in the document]
MukundC25: Fixed. When a scheduled email with an attachment is queued, we extend the `CachedFile.expires` to `scheduled_at + 1 day`, ensuring the file survives until delivery regardless of the periodic cleanup job.

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,385 @@
...
+            self.status = AdminEmailStatus.QUEUED
+            self.save(update_fields=['status'])
+
+        return True
+
+    def _build_context(self, recipient: 'AdminEmailQueueRecipient') -> dict[str, str]:
+        context: dict[str, str] = {
+            'platform_name': str(getattr(django_settings, 'PLATFORM_NAME', 'Eventyay')),
+            'platform_url': str(getattr(django_settings, 'SITE_URL', '')),
+            'support_email': str(getattr(django_settings, 'SUPPORT_EMAIL', '')),
```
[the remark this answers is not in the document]
MukundC25: Fixed

app/eventyay/control/tasks.py
```diff
@@ -87,3 +90,54 @@ def delete_organizer_data(organizer_id: int, user_id: int | None = None) -> None
             protected_labels,
         )
         raise
+
+
+@app.task(bind=True, name='eventyay.control.send_admin_email', max_retries=3, default_retry_delay=60, acks_late=True)
+@scopes_disabled()
+def send_admin_email(self, admin_email_id: int) -> None:
```
Rachit7168, line 97: When mail.send() returns True, but the status remains AdminEmailStatus.QUEUED (which happens if it was only partially sent due to exceptions), this celery task logs a warning and exits normally. Since no exception is raised, Celery considers the task successful and will not retry it. The remaining unsent emails will be permanently stuck in the queue.

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,400 @@
...
+
+class AdminEmailStatus(models.TextChoices):
+    DRAFT = 'draft', _('Draft')
+    QUEUED = 'queued', _('Queued')
+    SENDING = 'sending', _('Sending')
+    SENT = 'sent', _('Sent')
+    CANCELLED = 'cancelled', _('Cancelled')
+
+
+class AdminEmailQueue(models.Model):
```
Rachit7168, line 44: If this send() task fails partially and is later retried by Celery (assuming the retry bug above is fixed), any_dispatched will evaluate to True again for any new successes. This means your BCC list will receive a duplicate copy of the same email for every retry batch.

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,400 @@
...
+        User,
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='admin_email_queue',
+        verbose_name=_('Created by'),
+    )
+
+    recipient_group = models.CharField(
+        max_length=30,
```
Rachit7168, line 55: You are catching the dispatch exception and setting recipient.error = 'Dispatch failed', but discarding the actual exception context. It would be highly beneficial for debugging to include the error message.

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,400 @@
...
+    )
+
+    account_status = models.CharField(max_length=20, blank=True, default='', verbose_name=_('Account status'))
+    user_role = models.CharField(max_length=20, blank=True, default='', verbose_name=_('User role'))
+    language = models.CharField(max_length=20, blank=True, default='', verbose_name=_('Language / locale'))
+    created_after = models.DateTimeField(null=True, blank=True)
+    created_before = models.DateTimeField(null=True, blank=True)
+    last_active_after = models.DateTimeField(null=True, blank=True)
+    last_active_before = models.DateTimeField(null=True, blank=True)
+
```
Rachit7168, line 375: When an admin clicks "Send now" on an email in the outbox that was previously scheduled for a future date, this view correctly queues the send_admin_email Celery task. However, the Celery task will see that mail.scheduled_at > current_time and immediately re-delay itself. The email will not send immediately as the user requested.

Rachit7168 (changes requested):
Thanks for working on this. Please address all the comments and : 

You have inline imports here: from eventyay.base.models.base import CachedFile, import datetime, and from django.utils.timezone import now as _now. These should be moved to the top of the file to adhere to PEP-8. Furthermore, now is already imported at the top of the file as tz_now (line 13), and _now isn't actually being used anywhere in this block

app/eventyay/control/tasks.py
MukundC25: Fixed. `send()` now raises `RuntimeError` when unsent recipients remain after a partial dispatch, which triggers Celery's existing retry path. The task no longer exits silently on partial sends.

app/eventyay/base/models/admin_mail.py
MukundC25: Fixed

app/eventyay/base/models/admin_mail.py
MukundC25: Fixed. Changed `recipient.error = 'Dispatch failed'` to `recipient.error = str(exc)` so the actual exception message is stored for debugging.

app/eventyay/base/models/admin_mail.py
MukundC25: Fixed. `AdminMessageSendView` now clears `scheduled_at` (sets it to `None`) before calling `apply_async`, so the worker's schedule check no longer re-delays it.

MukundC25:
@Rachit7168 please, re-review

Rachit7168 (changes requested):
Please renumber the migration file

app/eventyay/control/views/admin_messages.py
```diff
@@ -0,0 +1,1130 @@
...
+                    Q(teams__all_events=True, teams__organizer__events__pk__in=setup_event_ids)
+                    | Q(teams__limit_events__pk__in=setup_event_ids)
+                ).distinct()
+
+    seen: set[str] = set()
+    result: list[dict] = []
+    reason = str(dict(AdminRecipientGroup.choices).get(recipient_group, recipient_group))
+    skipped = 0
+
+    for user in qs.only('pk', 'email', 'fullname', 'is_active', 'is_staff', 'is_administrator'):
```
Rachit7168: **Performance Issue (N+1 Query):** Because `user.get_full_name()` falls back to `self.wikimedia_username` if `fullname` is empty, Django will execute a separate database query for **every single user** that doesn't have a full name (since `wikimedia_username` isn't included in this `.only()` clause). 

Please add `'wikimedia_username'` to the `.only()` list to prevent the N+1 query bottleneck:
```python
for user in qs.only('pk', 'email', 'fullname', 'wikimedia_username', 'is_active', 'is_staff', 'is_administrator'):

File: app/eventyay/base/models/admin_mail.py [hunk 21]
```diff
@@ -0,0 +1,405 @@
...
+        from eventyay.common.mail import mail_send_task
+
+        reply_to_addr = self.reply_to or getattr(django_settings, 'DEFAULT_FROM_EMAIL', '')
+        bcc_list = [b.strip() for b in self.bcc.split(',') if b.strip()] if self.bcc else []
+        attachments = self._resolve_attachment()
+
+        is_first_attempt = not self.recipients.filter(sent=True).exists()
+
+        any_dispatched = False
+        for recipient in valid_recipients:
```
Rachit7168, line 214: **Major Database Bottleneck:** You are calling `recipient.save(update_fields=['sent', 'error'])` individually inside this loop. If an admin sends an email to 50,000 users, this will result in 50,000 separate `UPDATE` queries executing sequentially inside the Celery task's open database transaction (`with transaction.atomic():`). This will lock rows for a very long time and likely cause transaction timeouts or database throttling.

Please collect the results in memory and use a bulk update instead:
```python
        sent_ids = []
        error_updates = []
        
        for recipient in valid_recipients:
            # ... build context ...
            try:
                mail_send_task.apply_async(...)
                sent_ids.append(recipient.pk)
            except Exception as exc:
                logger.exception('Error dispatching admin email to %s', recipient.email)
                recipient.error = str(exc)
                error_updates.append(recipient)

        # Bulk update at the end of the loop
        if sent_ids:
            AdminEmailQueueRecipient.objects.filter(pk__in=sent_ids).update(sent=True, error=None)

        if error_updates:
            AdminEmailQueueRecipient.objects.bulk_update(error_updates, ['error'])

File: app/eventyay/control/tasks.py [hunk 22]
```diff
@@ -87,3 +90,52 @@ def delete_organizer_data(organizer_id: int, user_id: int | None = None) -> None
...
+                return
+
+            result = mail.send()
+
+            if not result:
+                logger.warning('[AdminMail] AdminEmailQueue ID %s: send returned False.', admin_email_id)
+            else:
+                logger.info('[AdminMail] AdminEmailQueue ID %s: all emails sent successfully.', admin_email_id)
+
+    except MaxRetriesExceededError:
```
Rachit7168, line 134: **State Handling Bug:** If a deterministic error occurs during the dispatch loop (e.g., an edge-case email address causing `apply_async` to fail locally), `send()` correctly raises `RuntimeError` to trigger a retry. 

However, after 3 retries, Celery will catch it here and drop the task. Because the status is never updated in this block, the email will remain stuck in the `QUEUED` / `SENDING` state **forever** in the admin outbox UI.

Please mark the queue as `SENT` (meaning finished processing) even if retries are exceeded, so it clears from the active Outbox queue (any failed recipients are correctly tracked via `recipient.error` anyway).
```python
    except MaxRetriesExceededError:
        logger.error('[AdminMail] Max retries exceeded for AdminEmailQueue ID %s', admin_email_id)
        AdminEmailQueue.objects.filter(pk=admin_email_id).update(
            status=AdminEmailStatus.SENT, sent_at=now()
        )

**Review by Rachit7168 — CHANGES_REQUESTED** [review]
**Scalability Issue (UI crash):** This approach works fine locally with a small dataset, but it will break in production. If there are thousands of users, organizers, or speakers, querying and loading all of them into the `<select>` choices at page load will crash the browser and cause severe database overhead.

Instead of loading all users at once, we should use one of the following approaches:
1. **Add Pagination:** Paginate the dropdown list so it only loads a small chunk of users at a time.
2. **AJAX Search (Recommended):** Use an AJAX-backed search dropdown (similar to the one we use when adding speakers to a session). This way, the browser only fetches results dynamically when the admin types at least 3 characters, showing users based on matching name or email.

This will ensure the page remains fast and stable even with a massive user base.

cc : @Saksham-Sirohi

File: app/eventyay/base/models/admin_mail.py [hunk 23]
```diff
@@ -0,0 +1,405 @@
...
+            except (ValueError, FileNotFoundError, OSError):
+                logger.warning(
+                    'CachedFile %s content unavailable for AdminEmailQueue %s',
+                    self.attachment, self.pk,
+                )
+                return None
+            return [{
+                'name': cf.filename or 'attachment',
+                'content': content,
+                'content_type': cf.type or 'application/octet-stream',
```
ViRUS-0-0, line 168: In `AdminEmailQueue._resolve_attachment()`, attachment files are read into raw binary `bytes` (`content = cf.file.read()`) and passed via `attachments` into `mail_send_task.apply_async(kwargs={'attachments': attachments})`. Modern Celery defaults to the JSON serializer, which does not support Python `bytes` objects and will raise `TypeError: Object of type bytes is not JSON serializable` at runtime whenever an attachment is included. Furthermore, even if serialized with pickle, embedding full file binary payloads into Celery kwargs for hundreds or thousands of recipients will duplicate the file across thousands of broker messages, creating severe RabbitMQ/Redis memory exhaustion and potential OOM crashes.

Resolution: Pass the CachedFile ID (or storage path) to `mail_send_task` rather than loading file contents into memory and passing raw bytes through Celery task parameters. Let the task worker read the CachedFile during mail construction.

app/eventyay/control/tasks.py
```diff
@@ -87,3 +90,52 @@ def delete_organizer_data(organizer_id: int, user_id: int | None = None) -> None
...
+@scopes_disabled()
+def send_admin_email(self, admin_email_id: int) -> None:
+    try:
+        with transaction.atomic():
+            mail = (
+                AdminEmailQueue.objects
+                .select_for_update(skip_locked=True)
+                .filter(pk=admin_email_id)
+                .exclude(status__in=[AdminEmailStatus.SENT, AdminEmailStatus.DRAFT, AdminEmailStatus.CANCELLED])
+                .first()
```
ViRUS-0-0: In `send_admin_email`, the task wraps `mail.send()` inside `with transaction.atomic():`. In `mail.send()`, for each recipient, `mail_send_task.apply_async()` dispatches an email task immediately to the broker, and `recipient.sent = True` is updated. If any recipient fails or dispatches are incomplete, `mail.send()` raises `RuntimeError` at line 289. Raising this exception inside `with transaction.atomic():` rolls back the entire database transaction, resetting all `recipient.sent` flags back to `False`. However, the Celery broker messages have already been queued. When Celery retries the task, `valid_recipients` re-fetches all recipients who were already processed and re-dispatches emails, causing duplicate emails to be repeatedly sent to users.


Resolution: Do not wrap the dispatch loop and broker invocations inside a single database atomic transaction that rolls back on partial failures. Instead, lock the `AdminEmailQueue` status, perform per-recipient state commits outside the atomic block or per batch, and do not raise transaction-rolling exceptions after broker calls have already been emitted.

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,405 @@
...
+
+    def _build_context(self, recipient: 'AdminEmailQueueRecipient') -> dict[str, str]:
+        context: dict[str, str] = {
+            'platform_name': str(getattr(django_settings, 'PLATFORM_NAME', 'Eventyay')),
+            'platform_url': str(getattr(django_settings, 'SITE_URL', '')),
+            'support_email': str(getattr(django_settings, 'SUPPORT_EMAIL', '')),
+            'support_url': str(getattr(django_settings, 'SUPPORT_URL', '')),
+        }
+
+        user = recipient.user
```
ViRUS-0-0, line 310: `AdminComposeForm` configures placeholders for event (`{event_name}`, `{event_url}`, `{event_start_date}`, `{event_end_date}`) and organiser (`{organiser_name}`, `{organiser_url}`). However, `AdminEmailQueue._build_context()` only populates `platform_*` and `user_*` keys, and leaves `account_url` as an empty string. Organiser and event keys are completely omitted from the context dictionary. When platform admins send emails referencing these variables, literal placeholder tags like `{event_name}` and `{organiser_name}` will be sent unreplaced to recipients.

Resolution: Populate default/fallback values for all supported placeholders in `_build_context()`. When recipient resolution is tied to specific events or organisers, resolve the corresponding objects to provide accurate context strings.

app/eventyay/control/forms/admin/admin_messages.py
```diff
@@ -0,0 +1,446 @@
...
+        self.fields['language'].choices = lang_choices
+
+        self.fields['selected_organisers'].queryset = Organizer.objects.all().order_by('name')
+        self.fields['selected_events'].queryset = Event.objects.all().order_by('name')
+        self.fields['selected_users'].queryset = (
+            User.objects.filter(is_active=True)
+            .exclude(email__isnull=True)
+            .exclude(email='')
+            .order_by('email')
+        )
```
ViRUS-0-0: In `AdminComposeForm.__init__`, `self.fields['selected_users'].queryset` is instantiated with `User.objects.filter(is_active=True).exclude(email__isnull=True).exclude(email='').order_by('email')`. On platforms with tens of thousands of users, rendering this form will evaluate the entire user table into memory, construct tens of thousands of `<option>` tags, and cause severe HTTP response latency and browser crashes when rendering Choices.js.

app/eventyay/control/views/admin_messages.py
```diff
@@ -0,0 +1,1130 @@
...
+    user_role = filters.get('user_role', '')
+    language = filters.get('language', '')
+    event_status = filters.get('event_status', '')
+    created_after = filters.get('created_after')
+    created_before = filters.get('created_before')
+    last_active_after = filters.get('last_active_after')
+    last_active_before = filters.get('last_active_before')
+    selected_organiser_ids = filters.get('selected_organisers', [])
+    selected_event_ids = filters.get('selected_events', [])
+    selected_user_ids = filters.get('selected_users', [])
```
ViRUS-0-0, line 71: In `resolve_admin_recipients`, audience filtering checks keys `'selected_organisers'`, `'selected_events'`, and `'selected_users'`. However, `AdminEmailQueueFilter` model persists these fields as `organiser_ids`, `event_ids`, and `selected_user_ids`. When a saved draft or queued email is loaded from the database and converted into a dictionary, these keys do not match. As a result, specific organiser, event, and user filters are silently lost upon editing a draft or executing recipient resolution from persisted state.

app/eventyay/control/forms/admin/admin_messages.py
```diff
@@ -0,0 +1,446 @@
...
+
+        if send_immediately and scheduled_at:
+            raise forms.ValidationError(
+                _('You cannot select "Send immediately" and also specify a scheduled time.')
+            )
+
+        if cleaned.get('delivery_mode') == 'now' and scheduled_at:
+            cleaned['scheduled_at'] = None
+
+        return cleaned
```
ViRUS-0-0: In `AdminComposeForm.clean()`, validation checks if `send_immediately` is selected alongside `scheduled_at`, and clears `scheduled_at` if `delivery_mode == 'now'`. However, it fails to validate that `scheduled_at` is provided when `delivery_mode == 'later'`. Since `scheduled_at` is marked `required=False`, selecting 'Schedule for later' without specifying a time passes validation and defaults to an immediate outbox send instead of notifying the admin.

 Resolution: Add a validation check ensuring that when `delivery_mode == 'later'`, `scheduled_at` must be provided.

app/eventyay/control/tasks.py
```diff
@@ -87,3 +90,52 @@ def delete_organizer_data(organizer_id: int, user_id: int | None = None) -> None
...
+                    '[AdminMail] AdminEmailQueue ID %s: not found, already sent, or locked. Skipping.',
+                    admin_email_id,
+                )
+                return
+
+            current_time = now()
+            if mail.scheduled_at and mail.scheduled_at > current_time:
+                countdown = max(1, int((mail.scheduled_at - current_time).total_seconds()))
+                logger.info(
+                    '[AdminMail] AdminEmailQueue ID %s: scheduled for %s, rescheduling in %s seconds.',
```
ViRUS-0-0: Do not rely on Celery task retry countdown for scheduled delivery. Keep the status as `queued` and register a periodic beat task to poll and trigger `send_admin_email` when `scheduled_at <= now()`.

app/eventyay/control/templates/pretixcontrol/admin/messages/template_detail.html
```diff
@@ -0,0 +1,75 @@
...
+        <table class="table table-condensed">
+          <tbody>
+            <tr>
+              <th class="col-md-2">{% translate "Role" %}</th>
+              <td><code>{{ role }}</code></td>
+            </tr>
+            <tr>
+              <th>{% translate "Subject" %}</th>
+              <td>{{ mail_template.subject }}</td>
+            </tr>
```
ViRUS-0-0: Acceptance criteria for #5346 specify that the Template detail view must show: Template name, Category, Trigger, Recipient type, Subject, Body preview, Available placeholders, Last updated, and Source/system identifier. In `template_detail.html`, Category, Trigger, Recipient type, Placeholders, and Last updated are omitted entirely.

Update `template_detail.html` and `AdminMessageTemplateDetailView` to pass and display category, trigger, recipient type, system identifier, and available placeholder tokens.

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,405 @@
...
+                        kwargs={
+                            'to': [bcc_addr],
+                            'subject': self.subject,
+                            'body': self.message,
+                            'html': AdminEmailQueue.make_html(self.message),
+                            'reply_to': [],
+                            'event': None,
+                            'cc': [],
+                            'bcc': [],
+                            'attachments': attachments,
```
ViRUS-0-0: In `AdminEmailQueue.send()`, when dispatching email copies to addresses listed in `bcc_list`, `'reply_to': []` is hardcoded. If a BCC recipient replies to the email, their client will direct the response to the default platform sender rather than the administrator's configured `reply_to` address.

Resolution: Forward `[reply_to_addr]` in kwargs when sending BCC copies.

app/eventyay/control/templates/pretixcontrol/admin/messages/sent.html
```diff
@@ -0,0 +1,61 @@
...
+                <div class="btn-group" role="group">
+                  <form method="post" action="{% url 'eventyay_admin:admin.messages.duplicate' pk=mail.pk %}">
+                    {% csrf_token %}
+                    <button type="submit" class="btn btn-sm btn-default" title="{% translate 'Duplicate as draft' %}">
+                      <i class="fa fa-copy"></i>
+                    </button>
+                  </form>
+                </div>
+              </div>
+            </td>
```
ViRUS-0-0: The requirements specify that the Sent list should provide actions to 'View email', 'View recipient list', and 'Duplicate'. The implemented `sent.html` only provides a 'Duplicate as draft' action, preventing administrators from inspecting sent message bodies or reviewing recipient delivery states.

Resolution: Add 'View email' preview modal/detail button and 'View recipient list' action button to `sent.html`.

ViRUS-0-0 (changes requested):
THE PR Adds the functionality but it is far from merge ready. Please resolve the comments

arpan7sarkar (commented):
Looking at this with the follow-up #5669 (multi-language admin emails) in mind. Three things worth deciding now rather than later, since `0075` hasn't shipped yet.

### 1. `subject` / `message` could be i18n fields from the start

`base/models/admin_mail.py:61-62` stores them as plain fields:

```python
subject = models.CharField(max_length=500, blank=True, default='', verbose_name=_('Subject'))
message = models.TextField(blank=True, default='', verbose_name=_('Message'))
```

But the form already uses `I18nEmailBodyFormField` for the message. That mismatch is why `_get_message_text()` exists in `control/views/admin_messages.py:533`, and at line 541 it keeps only English and silently drops every other language:

```python
return data.get('en', '') or (next(iter(data.values()), '') if data else '')
```

The ticket-side `EmailQueue` in `plugins/sendmail/models.py` already uses `I18nTextField` for exactly this case.

Switching now would mean:

- extra locales stop being discarded
- `_get_message_text()` goes away entirely
- #5669 needs no `AlterField` migration stacked on a migration that hasn't been released yet

`I18nField.to_python()` wraps a plain string in `LazyI18nString`, so existing rows keep working with **no data migration**.

### 2. BCC copies skip placeholder replacement

```python
for key, value in context.items():
    subject = subject.replace('{' + key + '}', str(value))
    body = body.replace('{' + key + '}', str(value))
```

The BCC block at `admin_mail.py:255-262` passes the unresolved originals instead:

```python
'subject': self.subject,
'body': self.message,
'html': AdminEmailQueue.make_html(self.message),
```

So BCC recipients receive the raw `{user_name}` / `{platform_name}` text rather than the resolved values.

### 3. `send()` bypasses `mail()`, so no language is ever activated

It calls `mail_send_task.apply_async()` directly at `admin_mail.py:230`, rather than going through `base/services/mail.py:mail()`, which wraps rendering in `with language(locale)`.

This is mostly harmless today, since the subject and body are plain strings the admin typed. It matters once #5669 lands and the body is localized per recipient, and it already affects any translatable content added downstream of `mail_send_task`. Might be worth confirming that's intentional for now.

---

Happy to send a PR for any of these if that's easier than folding them in here.

Rachit7168 (changes requested):
Please address review left by Om and Arpan

app/eventyay/base/models/admin_mail.py
MukundC25: Fixed

app/eventyay/control/tasks.py
MukundC25: Fixed

app/eventyay/base/models/admin_mail.py
MukundC25: Fixed. `_resolve_attachment()` now returns only the `CachedFile` ID and metadata, no file bytes. The worker reads the file at send time, not during dispatch

app/eventyay/control/views/admin_messages.py
MukundC25: not a bug

app/eventyay/base/models/admin_mail.py
MukundC25: Fixed. Those placeholders have been removed from the compose page entirely. Since this is a platform-wide broadcast, there's no per-recipient event or organiser to resolve them from

MukundC25:
@arpan7sarkar 

> 1. subject / message could be i18n fields from the start

Good point. But it makes more sense to do the full switch in a PR against #5669, with the language tabs and per-user delivery all as one change rather than a partial one here.

> 2. BCC copies skip placeholder replacement

Fixed. BCC copies now resolve platform-level placeholders before sending.

> 3. send() bypasses mail(), so no language is ever activated

Yes, intentional for now. Since the message is plain text today there's nothing locale-sensitive to activate. Once #5669 adds language support, send() will activate the recipient's language before rendering.

app/eventyay/control/tasks.py
```diff
@@ -87,3 +90,60 @@ def delete_organizer_data(organizer_id: int, user_id: int | None = None) -> None
...
+
+        if mail is None:
+            logger.info(
+                '[AdminMail] AdminEmailQueue ID %s: not found, already sent, or locked. Skipping.',
+                admin_email_id,
+            )
+            return
+
+        current_time = now()
+        if mail.scheduled_at and mail.scheduled_at > current_time:
```
Saksham-Sirohi, line 115: ViRUS: scheduled delivery still uses Celery self.retry(countdown=...) instead of leaving queued + periodic beat poll.

app/eventyay/control/views/admin_messages.py
```diff
@@ -0,0 +1,1183 @@
...
+                    teams__organizer__pk__in=billed_org_ids
+                ).distinct()
+
+    # Ticketing status filter
+    if ticketing_status and recipient_group in (
+        AdminRecipientGroup.EVENT_ORGANISERS,
+        AdminRecipientGroup.EVENT_TEAM_MEMBERS,
+    ):
+        with scopes_disabled():
+            from eventyay.base.models.product import Product
```
Saksham-Sirohi: Inline imports of Product / Event_SettingsStore remain inside resolve_admin_recipients (Rachit/agents.md: imports at module top unless circular).

Saksham-Sirohi (changes requested):
Prior review items still open on this HEAD.

Many ViRUS UX/filter issues improved (select2 users, delivery_mode later validation, sent actions, template detail, BCC reply_to). Blockers remain: Celery countdown scheduling and residual inline imports. Reaffirm CHANGES.

arpan7sarkar:
> @arpan7sarkar
> 
> > 1. subject / message could be i18n fields from the start
> 
> Good point. But it makes more sense to do the full switch in a PR against #5669, with the language tabs and per-user delivery all as one change rather than a partial one here.
> 
> > 2. BCC copies skip placeholder replacement
> 
> Fixed. BCC copies now resolve platform-level placeholders before sending.
> 
> > 3. send() bypasses mail(), so no language is ever activated
> 
> Yes, intentional for now. Since the message is plain text today there's nothing locale-sensitive to activate. Once #5669 adds language support, send() will activate the recipient's language before rendering.

Ok @MukundC25  got it then once this pr will get merged i will be working on adding that feature.

app/eventyay/control/tasks.py
MukundC25: current approach follows the same pattern as the ticket/sendmail `send_queued_mail` task uses the same `self.retry(countdown=...)` approach for scheduled delivery

app/eventyay/control/tasks.py
```diff
@@ -87,3 +90,60 @@ def delete_organizer_data(organizer_id: int, user_id: int | None = None) -> None
...
+            countdown = max(1, int((mail.scheduled_at - current_time).total_seconds()))
+            logger.info(
+                '[AdminMail] AdminEmailQueue ID %s: scheduled for %s, rescheduling in %s seconds.',
+                admin_email_id,
+                mail.scheduled_at,
+                countdown,
+            )
+            self.retry(countdown=countdown, args=[admin_email_id], throw=False)
```
Saksham-Sirohi: Still using Celery `self.retry(countdown=…)` to wait until `scheduled_at`, with `max_retries=3`. A long-scheduled mail (or repeated early delivery) burns retries; on MaxRetriesExceededError below you then stamp status=SENT. Prefer relying solely on `apply_async(..., eta=mail.scheduled_at)` (already done in the view) and if the task wakes early, re-queue with eta without counting as a failure retry — never mark SENT on schedule exhaustion.

app/eventyay/control/tasks.py
```diff
@@ -87,3 +90,60 @@ def delete_organizer_data(organizer_id: int, user_id: int | None = None) -> None
...
+
+        if not result:
+            logger.warning('[AdminMail] AdminEmailQueue ID %s: send returned False.', admin_email_id)
+        else:
+            logger.info('[AdminMail] AdminEmailQueue ID %s: all emails sent successfully.', admin_email_id)
+
+    except MaxRetriesExceededError:
+        logger.error('[AdminMail] Max retries exceeded for AdminEmailQueue ID %s', admin_email_id)
+        AdminEmailQueue.objects.filter(pk=admin_email_id).update(
+            status=AdminEmailStatus.SENT,
```
Saksham-Sirohi: Marking AdminEmailQueue SENT on MaxRetriesExceededError is incorrect — operators will think the campaign delivered when it may not have. Use a failed/cancelled terminal status (or leave QUEUED) and surface the error.

app/eventyay/base/models/admin_mail.py
```diff
@@ -0,0 +1,412 @@
...
+        if not valid_recipients:
+            self.status = AdminEmailStatus.SENT
+            self.sent_at = now()
+            self.save(update_fields=['status', 'sent_at'])
+            return True
+
+        self.status = AdminEmailStatus.SENDING
+        self.save(update_fields=['status'])
+
+        from eventyay.common.mail import mail_send_task
```
Saksham-Sirohi, line 204: Inline import still present (`from eventyay.common.mail import mail_send_task` inside send()). Move to module top per project import rules (unless this is strictly for a circular import — if so, document why).

Saksham-Sirohi (changes requested):
Prior review items still open on this HEAD.

Prior Celery countdown + inline-import blockers still open on fe669bf4. Compose correctly uses eta=mail.scheduled_at, but send_admin_email still self.retry(countdown=…) under max_retries=3, and MaxRetriesExceededError incorrectly marks the queue SENT. Inline `from eventyay.common.mail import mail_send_task` remains inside admin_mail.send(). Many UX items were fixed, but scheduling correctness is still not merge-ready.

app/eventyay/base/models/admin_mail.py
MukundC25: circular import

Rachit7168 (approved):
LGTM, Tested 

Requires follow-up PRs for some things; will open an issue for it once this is merged. Also, the pagination for the user search is not addressed as well. Could be a follow-up. Current is good as a base implementation

Saksham-Sirohi (approved):
Prior blockers cleared on this HEAD (eta requeue for scheduled mail; MaxRetriesExceeded no longer marks SENT; mail_send_task local import justified with circular-import comment, same pattern as QueuedMail). LGTM with Rachit7168 approval on this SHA.

PR #5497, opened 2026-09-07
fix: correct pyproject.toml to discover all eventyay subpackages

ArnavBallinCode:
Changes `packages = ['eventyay']` to use `packages.find` so that subpackages like `eventyay.base` are correctly included when building wheels and sdists.

Previously, running `pip install` or building a wheel of the core repository would only include the top-level files in `eventyay/`, completely missing nested modules like `eventyay.base`. This caused `ModuleNotFoundError: No module named 'eventyay.base'` when plugins attempted to import core logic after installing the core repository.

## Summary by CodeRabbit

* **Chores**
  * Updated package discovery configuration to automatically include supported application packages during builds.

### DIFF
--- a/app/pyproject.toml
+++ b/app/pyproject.toml
@@ -289,8 +289,9 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 [tool.ruff.lint.isort]
 lines-after-imports = 2
 
-[tool.setuptools]
-packages = ["eventyay"]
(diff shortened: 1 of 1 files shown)


Saksham-Sirohi (approved):
Explicit `packages = ["eventyay"]` only ships the top-level package, so nested modules like `eventyay.base` were missing from wheels/sdists. `packages.find` with `include = ["eventyay*"]` and `namespaces = false` matches how this tree is laid out. Looks good.

PR #5436, opened 2026-09-06
feat(interpretation): integration of live captions UI and improvement of WHEP audio reliability

Nirupon123:
## What does this PR do?
This PR introduces the initial integration for live captions and refactors the WebRTC WHEP client for the interpretation feature. It provides a new set of stage tools in the video player for managing captions and AI Text-to-Speech (TTS), alongside improving connection reliability and fixing local development routing issues for interpretation streams.

- depends on: [https://github.com/fossasia/voxbento/pull/450](https://github.com/fossasia/voxbento/pull/450) and [https://github.com/fossasia/eventyay-interpretation/pull/109](https://github.com/fossasia/eventyay-interpretation/pull/109)

- fixing : #5433

### User Interface (`app/eventyay/webapp/video/src/views/rooms/item.vue`)
- **Caption Controls:** Introduced a new `.cc-controls` tool group under the video stage.
- **CC Toggle:** Added a closed-caption button (`mdi-closed-caption`) to toggle the visibility of live transcripts.
- **Dedicated Caption Dropdown:** Added a secondary `AudioTranslationDropdown` specifically for selecting the caption language, which appears when CC is enabled.
- **AI TTS Toggle:** Added a robot icon toggle (`mdi-robot`) to easily enable or disable AI Text-to-Speech playback.
- **Smart Language Syncing:** Added logic to automatically sync the caption language with the chosen interpretation audio language by default, while still allowing the user to manually override the caption language if desired.
- **Audio State Refactor:** Extracted the audio update logic into a new `recomputeInterpretationAudio()` method. Added fallback checks so that if the translation configuration is set to `'Original'` (or lacks a URL/YouTube ID), the interpretation audio is correctly nulled out, seamlessly returning the user to the default stream.

### WebRTC Integration (`app/eventyay/webapp/video/src/lib/webrtc/whep.js`)
- **ICE Gathering:** Added a Promise to wait for `iceGatheringState === 'complete'` (with a 500ms fallback timeout) before sending the SDP offer.
- **NAT Workaround:** Added logic to rewrite the SDP answer (`c=IN IP4 ...`) to use `window.location.hostname`. This fixes media routing issues when running the stack in Docker/WSL.
- **Abort Controller:** Implemented an `AbortController` to cleanly cancel pending fetch requests in `connect()` if `disconnect()` is called rapidly.
[...]

PR #5422, opened 2026-09-04
Feature: get different stream servers working

Saksham-Sirohi:
## Description

This PR unifies, modernizes, and stabilizes video streaming and conferencing integrations across Eventyay. It ensures robust support for multiple stream and meeting providers—**BigBlueButton (BBB)**, **Jitsi Meet**, **Janus WebRTC**, and **Zoom**—along with centralized TURN/STUN server management and a streamlined attendee in-room experience.

### Context & Motivation
- **Legacy Cleanup**: The previous `StreamingServer` abstraction was deprecated and lacked proper integration with modern room types and per-event permissions.
- **In-Room Experience**: Attendees clicking on Zoom meetings previously encountered external links that opened in new browser tabs, navigating away from the event platform. Leaving meetings also left users stranded or bounced them into organizer management views rather than the video attendee dashboard.
- **Maintainability & Security**: Inline styles and JavaScript in templates were refactored into dedicated, reusable static assets complying with Content Security Policy (CSP) guidelines and repository standards.

---

#### 1. Video Server Architecture & Legacy Model Cleanup
- **Migration & Model Removal**: Removed deprecated `StreamingServer` model and its corresponding forms/views (`app/eventyay/base/migrations/0069_delete_streamingserver.py`).
- **Unified Server Management**: Standardized service layers and management forms for **BigBlueButton**, **Janus**, **Jitsi**, and **TURN** servers (`app/eventyay/control/forms/server_management.py`, `video_settings.html`).
- **Live Video Modules**: Enhanced room channel modules (`bbb.py`, `januscall.py`, `jitsi.py`, `room.py`, `zoom.py`) with proper scoping, permission enforcement, and URL resolution.
- **Local Dev Tools**: Added a seeding management command (`seed_video_servers.py`) and a mock BBB container setup (`tools/mock_bbb_server.py`, `tools/Dockerfile.bbb`, `deployment/docker-compose.video.yml`) for local testing.

#### 2. Zoom Modern In-Room Integration
- **Embedded In-Room Player**: Replaced external tab redirects with an embedded in-room player (`#zoom-embedded-frame`) supporting camera, microphone, display capture, and clipboard permissions.
- **Meeting SDK v3.11.2 Support**: Integrated Zoom Meeting SDK signature generation (`generate_signature`) with automatic embedded fallback if credentials or CDN assets are unavailable.
- **Static Asset Separation**:
[...]

Issue #472, opened 2024-12-11
Create an editable start page


mariobehling:
At the moment there is no editable start page with information about the event. Please create an editable start page similar to other platforms. In the admin area add a page to edit the images and text of each section.

![Screenshot from 2024-12-11 05-01-55](https://github.com/user-attachments/assets/0bae6b38-24dc-4c9f-8db7-e38e32f37fa7)

Hopin used to have a nice page, e.g. here https://web.archive.org/web/20210831232426/https://hopin.com/
Instead of black use blue and gray as it fits.

Ensure to include a section with "events"
![Screenshot from 2024-12-11 05-03-19](https://github.com/user-attachments/assets/cc7bbac4-9ce4-44d2-ac40-ad5740d75690)

r3yc0n1c:
hi, I'd like to work on this

VRamazing:
Hi, @odkhang Can i take this up?

mariobehling:
This has been resolved. Hence closing this.

PR #5504, opened 2026-09-07
feat(interpretation): support human and AI audio streams

### LINKED ISSUE #5499: feat(interpretation) : accomodate multiple audio streams (from human interpretation booth and AI TTS) from voxbento
**Context:** The frontend must allow attendees to select between Human and AI streams from a single dropdown, and intelligently switch its underlying playback engine between WebRTC (WHEP) for human streams and a WebSocket Web Audio API scheduler for AI streams.

**Tasks:**

- **Update AudioTranslationDropdown UI:** Parse the updated payload provided by the backend to distinguish between Human and AI tracks. Visually update the dropdown options to clearly display the specific stream type (e.g., "Spanish (Human)", "German (AI)").
- **Re-introduce the WebSocket Audio Player:** Eventyay currently expects all interpretation streams to be WebRTC. Re-introduce the previously removed `audio-scheduler.js` and `tts-parser.js` logic to connect to the existing Voxbento TTS endpoint.
- Ensure the scheduler properly parses the custom binary frames (`[1-byte version][4-byte length][JSON header][audio bytes]`) sent by Voxbento.
- Implement exponential backoff for the TTS WebSocket (reusing the logic from `LiveCaptions.vue`).
- **Smart Audio Routing in MediaSource.vue:**
  - Update `recomputeInterpretationAudio()` in `item.vue` to pass the entire translation config (including `tts_ws_url` and `whip_url`) to Vuex.
  - **Routing:** If the active track has a `whip_url/whep_url` (Human), initialize the WebRTC player. If the track has a `tts_ws_url` (AI), initialize the WebSocket connection and pipe the binary messages to `audio-scheduler.js`.
  - **Teardown:** Implement rigorous teardown mechanisms. Switching from Human to AI (or vice versa) must completely destroy the previous player, close the WebRTC peer connection or WebSocket, and clear any audio buffers to prevent ghost audio or memory leaks.

sanketkokate712:
Fixes #5499

Attendees can now choose between a human interpreter and the Voxbento AI voice from the same interpretation dropdown. Each option says which one it is, so "Spanish (Human)" and "German (AI)" sit side by side and you know what you are picking.

The two kinds of stream arrive over different transports, so the player picks the right one based on the track you select:

- Human booths keep using the existing WHEP/WebRTC client.
- AI tracks open a WebSocket to Voxbento and play through the Web Audio API.

To make the AI side work I added back the two pieces that had been removed:

- `tts-parser.js` decodes Voxbento's binary frame format (`[1-byte version][4-byte length][JSON header][audio bytes]`), buffering across message boundaries since a frame can be split over several WebSocket messages.
- `audio-scheduler.js` owns the socket, schedules the decoded audio, and reconnects with exponential backoff if the connection drops.

Switching between a human and an AI track fully tears down whatever was playing before, closing the peer connection or socket and clearing buffers, so you don't get ghost audio or a leaked connection when someone flips between languages a few times.

One thing worth flagging for review: the dropdown validator used to require a `url` or `youtube_id` on every track, which silently dropped any track that only carried a stream endpoint. That meant AI tracks would never have shown up at all. It now also accepts tracks with just a `tts_ws_url` or `whep_url`.

**Testing:** `eslint` and `vite build` both pass on the video webapp. I have not been able to exercise this against a live Voxbento endpoint, so the binary frame decoding and the audio scheduling are the parts most worth a careful look from someone who can point it at a real stream.

## Summary by Sourcery

Enable attendees to choose and play human or AI interpretation audio streams from the same language selector.

New Features:
- Support selecting human interpreter and Voxbento AI audio streams for the same language.
- Play AI interpretation audio through a WebSocket and Web Audio pipeline with frame buffering, scheduling, and reconnection.

Bug Fixes:
- Preserve and correctly restore selected interpretation streams when multiple stream types share a language.
- Allow interpretation entries that provide only TTS WebSocket or WHEP stream endpoints.

Enhancements:
- Tear down active human and AI playback cleanly when switching streams or leaving the video.
- Label interpretation options by stream type and prefer human streams when deduplicating language entries.

Tests:
- Add coverage for AI audio scheduling, teardown, reconnect backoff, retry limits, backoff reset, and cancellation.

## Summary by CodeRabbit

- **New Features**
  - Added support for real-time text-to-speech audio interpretation alongside existing WHEP and YouTube playback.
  - TTS audio now supports smooth playback, pause/resume, automatic reconnection, and cleanup when switching or stopping streams.
  - Audio translation options can distinguish between multiple streams for the same language.

- **Bug Fixes**
  - Improved translation stream matching for TTS, WHEP, and WHIP sources.
  - Prevented duplicate language options while retaining the most suitable available stream.


mariobehling (approved):

ArnavBallinCode (changes requested):

ArnavBallinCode (changes requested):
Please also address all copilot comments

Nirupon123 (changes requested):
image

- same language cant be AI and booth both at the same time
- also please take of the copilot commets 
- Thank you so much for your effort. Please fix the issues, love to review again

sanketkokate712:
Hi @Rachit7168 @ArnavBallinCode @Nirupon123, I've pushed another round of fixes. I went back to the Voxbento source ([`fossasia/voxbento`](https://github.com/fossasia/voxbento)) to check this PR against the real TTS protocol instead of my assumptions, and that turned up two bugs that would have kept AI audio from ever playing. Details below.

## What changed

### 1. Merged latest `dev` and resolved conflicts (`c8c9a98`)
The PR conflicted with #5422 and #5436 in `MediaSource.vue` and `AudioTranslationDropdown.vue`. The resolution keeps both sides:
- From `dev`: the `youtube_id` fallback, forcing `play()` after WHEP connects (autoplay fix), and the dropdown emitting the full selected entry.
- From this PR: routing between WHEP and TTS, full teardown on switch, and `whepUrl`/`ttsWsUrl` in the `languageChanged` payload.

### 2. TTS frames are now decoded with Voxbento's real wire format (`a742ec9`)
Voxbento packs every TTS message in [`portal/websockets/manager.py`](https://github.com/fossasia/voxbento/blob/main/portal/websockets/manager.py) as:

```python
frame = struct.pack(">BI", 1, len(header_bytes)) + header_bytes + audio_bytes
```

Its listener ([`listener-event.js`](https://github.com/fossasia/voxbento/blob/main/portal/static/js/listener-event.js)) decodes the audio as **16-bit little-endian mono PCM at 24 kHz**. The previous code got this wrong in two ways:

| | Before | Now |
|---|---|---|
| 4-byte length field | Read as the length of the **whole frame**, then brace-scanned for the JSON end. Audio bytes spilled into the "next frame" and got thrown away as garbage. | Read as the **JSON header length**. Everything after the header is audio, one frame per WebSocket message (as in Voxbento's own `tts-parser.js`). |
| Audio samples | Treated as 8-bit unsigned (`(b - 128) / 128`) and played at the AudioContext rate (usually 48 kHz). | Decoded as `Int16` LE → float and played at 24 kHz (or `header.sample_rate` if the server sends one). |

The old tests built frames the same wrong way the parser read them, so they passed without catching this. I rewrote them to encode frames exactly like the Python server.

Other scheduler fixes in the same commit:
- **Jitter buffer:** when the queue runs dry, playback re-anchors 0.25 s ahead, like Voxbento's `audio-scheduler.js`, so network jitter doesn't cause back-to-back gaps.
- **Backoff:** capped at 10 s, matching `LiveCaptions.vue`, as the issue asks.
- **Teardown:** stops audio that is already queued (`source.stop()`), not only the socket, so switching tracks can't leave half a sentence playing.
- **Pause:** if the main player is paused before the first frame arrives, the newly created AudioContext stays suspended.
- **Failed first connect:** previously this disconnected the scheduler for good. It now keeps retrying with backoff, and switching tracks or leaving the room still tears it down.

### 3. Human/AI tracks stay playable, one entry per language (`5bba7e6`)
- **Silent regression from the merge:** `dev`'s new `recomputeInterpretationAudio()` in `item.vue` nulled any selection without `url`/`youtube_id`. WHEP-only and TTS-only tracks carry neither, so after the merge **no human or AI track would have played**. It now also accepts `whepUrl`/`ttsWsUrl`.
- **Same language as both AI and booth (@Nirupon123's review):** each language appears once. When the backend sends both a booth and an AI stream for the same language, the **human interpreter is kept**. Before, whichever entry happened to come first won. CodeRabbit suggested keeping both, but I followed the maintainer review here.

## Screenshots

These come from a small harness that mounts this PR's actual `AudioTranslationDropdown.vue`, `interpretation-streams.js`, `audio-scheduler.js` and `tts-parser.js`. They are driven by a mock Voxbento TTS WebSocket that packs frames byte for byte like `manager.py` and sends synthetic speech-like PCM16 audio at 24 kHz. The waveform comes from an `AnalyserNode` attached to the Web Audio graph, so it shows audio that was actually played.

**1. Dropdown:** the backend sent Spanish twice (AI + Human). Only the human booth is shown, and each option is labelled Human/AI.

Dropdown options

**2. German (AI) selected:** the TTS WebSocket opens, frames are decoded (JSON header caption, 16800 samples @ 24000 Hz) and queued back to back (0.25→0.95 s, 0.95→1.65 s, …).

AI track playing

**3. Switch from German (AI) to French (Human):** the socket closes (server: open 0, closed 1), the AudioContext is closed, no audio is scheduled after teardown, the waveform goes flat, and routing moves to the WHEP client.

Switch AI to Human teardown

## Testing
- `node --test src/lib/audio-scheduler.test.js`: **12/12 pass**. Covers frame parsing (header length, bad version, truncation), PCM16 decoding, 24 kHz gapless queueing, ignoring error/empty frames, teardown stopping queued audio and dropping late messages, pause before the first frame, backoff with the 10 s cap, the retry limit, backoff reset, retrying after a failed first connect, and cancelling a pending retry.
- `eslint` on all changed files: clean.
- `vite build` would not finish locally: V8 ran out of native memory on my machine, which has little free memory at the moment. I'm relying on CI for the full build.

**Not covered:** the harness has no media server, so the WHEP (human) path is shown only as a routing decision; that code is unchanged from `dev`. It's still worth a check against a live Voxbento instance before merging.

ArnavBallinCode (changes requested):
Thanks for working through the review feedback and fixing the protocol/parsing issues. The changes look much more aligned with the actual VoxBento protocol now.

One important thing I’d still like to see before we merge this: **please actually set up and test the complete flow against a running VoxBento instance and Eventyay**, rather than relying only on the mock WebSocket harness.

The current tests are useful for validating the parser and scheduler in isolation, but they don't prove that the complete integration works end to end. In particular, we need to verify the real flow:

Eventyay interpretation stream → VoxBento TTS endpoint → WebSocket → TTS frames → AudioScheduler → Web Audio → attendee hears the AI interpretation.

Please set up a real VoxBento booth/stream, connect it to the Eventyay video room, and test at least:

* AI interpretation stream actually appears in the dropdown
* Selecting it establishes the real WebSocket connection
* Real VoxBento audio is decoded and audible
* Switching between streams works without ghost/overlapping audio
* Disconnect/reconnect works with the real endpoint
* Human/WHEP interpretation still works
* The complete flow works after refreshing/rejoining the room

A mock server can confirm that the JavaScript behaves according to the mocked protocol, but it can’t catch integration/configuration issues between Eventyay and the actual VoxBento deployment. Since this PR is specifically adding VoxBento support, I think we should validate that real integration before approving it.

Once you’ve tested this end to end, please add a short screen recording showing the actual Eventyay + VoxBento setup and the AI interpretation playing. That will also make the PR much easier to validate.

Thanks for the work so far. Please re request review only once the real end-to-end test is done and we will take another look.

app/eventyay/webapp/video/src/interpretation-streams.js
```diff
@@ -22,5 +22,20 @@ export function pluginLanguageStreams(room) {
 	const usable = Array.isArray(streams)
 		? streams.filter(entry => isUsableAudioTranslationEntry(entry))
 		: []
-	return ensureOriginalLanguageEntry(usable)
+
+	// A language is offered once: either a human booth or the AI voice, never both.
```
Saksham-Sirohi: Confirms one entry per language with human preferred. Keep this product rule documented; do not reintroduce duplicate language keys without distinct selection identities.

app/eventyay/webapp/video/src/lib/audio-scheduler.js
```diff
@@ -0,0 +1,190 @@
...
+const JITTER_BUFFER_SEC = 0.25;
+// Same backoff shape as LiveCaptions.vue: 1s, 2s, 4s, ... capped at 10s.
+const BASE_RETRY_DELAY_MS = 1000;
+const MAX_RETRY_DELAY_MS = 10000;
+
+/**
+ * Plays Voxbento AI interpretation audio: owns the TTS WebSocket, decodes
+ * each frame and queues it for gapless playback through the Web Audio API.
+ */
+export class AudioScheduler {
```
Saksham-Sirohi, line 14: Code-level scheduler/parser fixes look sound, but the open E2E request stands: please post evidence against a real VoxBento instance (dropdown → WS → audible AI audio, switch/teardown, reconnect, human WHEP still works, refresh/rejoin) before re-requesting review.

Saksham-Sirohi (changes requested):
Parser/scheduler/reconnect fixes look sound, and one stream per language with human preferred matches the dual-stream feedback. Still blocked on the open request for real VoxBento↔Eventyay E2E evidence before merge.

sanketkokate712:
@ArnavBallinCode @Saksham-Sirohi I've now tested this end to end against a real VoxBento instance. Recording below.

https://drive.google.com/file/d/1Mot8xlhvQb6sblQI9Vw3cFKkHinrG2QZ/view?usp=sharing

**Setup.** Eventyay runs this branch. VoxBento runs from `main` in Docker (portal + MediaMTX) using only local providers: faster-whisper `base` for the floor, NLLB-600M for translation, and Supertonic for TTS. Eventyay was connected through the normal OAuth flow (VoxBento developer app, then "Connect VoxBento" in the plugin), and the room was synced through the plugin's usual sync, so every URL below comes from the real integration. Nothing was hand-written. I skipped Jitsi. The floor audio is a recorded English talk, published to the same RTSP path the floor-bot uses (`voxdemo/1/floor`), so everything downstream of the floor-bot is the real pipeline.

**The real test caught a gap the mock couldn't.** The interpretation plugin never sent a `tts_ws_url`. It also sends a human booth's WHEP URL in `youtube_id`, not in `whep_url`. With the plugin as it is today, AI tracks never appear, and real human booths lost the "Human" label and the human-preferred dedupe. I fixed both sides:
- fossasia/eventyay-interpretation#120 adds `stream_type` (human/ai) from #111 and builds `tts_ws_url` for AI languages from VoxBento's current route: `/ws/tts/{voxbento_room_id}/{lang}/{event}-{room}-floor`.
- `fb43481` on this branch classifies streams in one place, treats a non-YouTube URL in `youtube_id` as a human booth, and honours `stream_type`. It also keeps the one-entry-per-language, human-preferred rule, now documented in `interpretation-streams.js`.

**Checklist results:**

- [x] **AI stream appears in the dropdown.** Options are Original, Spanish (Human), German (AI), French (AI), taken straight from the plugin payload after sync.
- [x] **Selecting it opens the real WebSocket** to `ws://localhost:8100/ws/tts/1/de/voxdemo-1-floor`.
- [x] **Real VoxBento audio is decoded and audible.** In a headless run I hooked `AudioBufferSourceNode.start` and got 15 s of German speech scheduled in the first 30 s. VoxBento sends each segment twice: first text-only with 0 audio bytes, then the audio. The scheduler skips the empty frame as intended.
- [x] **Switching without ghost audio.** German → French closes the German socket, closes its AudioContext and stops the 2 buffers that were still queued. Going to Original schedules 0 new buffers afterwards.
- [x] **Disconnect/reconnect.** After `docker restart voxbento-portal-1`, the client retried at 1 s, 2 s, 4 s, 8 s and 10 s, reconnected, and German audio resumed once VoxBento's floor pipeline was running again.
- [x] **Human/WHEP still works.** Spanish (Human) plays the booth over WHEP from MediaMTX. In the video I'm the interpreter, publishing over WHIP from MediaMTX's publish page. Switching to an AI track tears the peer connection down.
- [x] **Refresh/rejoin.** After a reload, German (AI) plays again.

I also merged the latest `dev`; there were no conflicts. Tests: the interpretation tests pass 16/16 under `node --test` (`interpretation-streams.test.js`, whose 4 new cases use the real plugin payload shapes, plus `audio-scheduler.test.js`), eslint is clean, and the plugin suite passes 70/70.

One thing to flag for later: when fossasia/voxbento#466 simplifies the TTS route to `/ws/tts/{booth_id}`, only the URL builder in the plugin needs to change. The frontend just uses whatever `tts_ws_url` it receives.

app/eventyay/webapp/video/src/interpretation-streams.js
sanketkokate712: Agreed. The rule is documented at the dedupe in `interpretation-streams.js`, including the reason: the dropdown and the stored selection are keyed by language, so a second stream for the same language would first need a distinct selection identity. I won't reintroduce duplicate language keys without that.

app/eventyay/webapp/video/src/lib/audio-scheduler.js
sanketkokate712: Thanks. The real VoxBento↔Eventyay E2E evidence is now posted here, with a recording: https://github.com/fossasia/eventyay/pull/5504#issuecomment-5737304007

It covers every item you listed: dropdown → WS → audible AI audio, switch/teardown, reconnect, human WHEP, and refresh/rejoin. It was run against VoxBento `main` with a real OAuth connect and plugin room sync. Testing surfaced a plugin-side gap (no `tts_ws_url` was ever sent), which fossasia/eventyay-interpretation#120 fixes.

Nirupon123 (changes requested):
thanks for the contribution @sanketkokate712  i have seen your changes in eventyay-interpretation repo, added a comment there by looking the video that looks really cool and well implemented as we need but can you confirm if a user can hear audio from different languages and see the captions in other language, like a mix up situation?? and please take a look in the eventyay-interpretation comment.

Nirupon123:
@codex review

sanketkokate712:
> thanks for the contribution @sanketkokate712 i have seen your changes in eventyay-interpretation repo, added a comment there by looking the video that looks really cool and well implemented as we need but can you confirm if a user can hear audio from different languages and see the captions in other language, like a mix up situation?? and please take a look in the eventyay-interpretation comment.

Could you please clarify what specific fixes are required at this stage and whether the current direction of my implementation aligns with the project's expectations, especially considering the latest end-to-end testing and the points raised in the recent reviews?

sanketkokate712:
@Nirupon123 thanks for the review. I've taken over the TTS work across all three repos as you suggested, and this recording runs every component on its new branch:

- VoxBento: fossasia/voxbento#496, which builds on your #472 (your commit is kept as the first commit)
- Plugin: fossasia/eventyay-interpretation#120
- Eventyay: this PR

**Video:** https://drive.google.com/file/d/1152vpzv3f-lSLxn0im9Gi8hlsry_VHaq/view?usp=sharing

**Your question: can someone hear one language and read captions in another?** Yes. The audio and caption dropdowns are independent. In the video I listen to German (AI) and set captions to French. German audio keeps playing on its own socket (`/ws/tts/voxdemo-1-ai-de`), and the French text comes from VoxBento's French caption channel (`/ws/captions/voxdemo-1-fr`). VoxBento's own listener page does the same: picking another caption language no longer stops the AI audio there. That was a bug in the draft.

**"Same language can't be AI and booth at the same time"** is now enforced at every layer:
- Eventyay room editor: each language row is either "Human interpreter" or "AI voice (VoxBento)".
- Plugin: one entry per language. AI languages are synced separately from human booths.
- VoxBento: a human booth always wins. The room sync, the admin form and the booth list all ignore AI for a language that has a human booth.
- Attendee dropdown: one entry per language, human preferred, now matched on the language code.

**New in this PR since the last recording**
- `74da48c` sends the VoxBento listener token on the AI audio socket, since VoxBento now authenticates `/ws/tts` like captions. It also matches duplicate languages by code.
- `2918738` adds the "Interpreted by" choice (Human interpreter / AI voice) to the room editor's interpretation list.

**Testing:** `node --test`: 21 passed (interpretation streams, audio scheduler, new editor-entry tests). eslint is clean and `vite build` passes. End to end against the new VoxBento branch, a scripted browser run passed all 14 checks, both with and without VoxBento's `BOOTH_ACCESS_TOKEN`: the dropdown lists Spanish (Human), German (AI) and French (AI); German plays with the listener token; German audio continues with French captions; switching tears down cleanly; Spanish plays over WHEP; Original stays silent. Saving French as Human and then as AI in the room editor moved it between a VoxBento interpreter booth and an AI booth each time.

PR #5529, opened 2026-09-08
fix(billing): define BILLING_REMINDER_SCHEDULE

### LINKED ISSUE #5526: Monthly billing crashes  `settings.BILLING_REMINDER_SCHEDULE` is never defined
**Describe the bug**

`settings.BILLING_REMINDER_SCHEDULE` doesn't exist. It's read in `collect_billing_invoice()` but never defined in `config/settings.py`.

```python
# app/eventyay/eventyay_common/tasks.py:128
billing_invoice = BillingInvoice(
    ...
    reminder_schedule=settings.BILLING_REMINDER_SCHEDULE,  
    ...
)
billing_invoice.save()  
```

Python evaluates all the kwargs before it constructs the object, so it dies on 140 and never reaches the save. That's the only place `BillingInvoice` is created in the whole codebase, so no invoice ever makes it into the DB.

Two other things make it worse:

`monthly_billing_collect()` (tasks.py:152) only catches `DatabaseError`, so the AttributeError isn't caught at all. And the call sits inside `for organizer: for event:` — so the first event that hits it takes down the rest of the run too.

It also won't show up on a fresh install. The guards above line 128 return early for events with no paid orders, so this only starts failing once there are real sales.

**Expected behavior**

Should create a `BillingInvoice` row for every event that had paid orders last month, with `reminder_schedule` set from config.

**Screenshots**

n/a, it's a Celery task.

**Additional context**

Nothing surfaces the failure — `process_auto_billing_charge`, `billing_invoice_notification`, `retry_failed_payment` and `check_billing_status_for_warning` all just query an empty table and log a clean run.

Fix is probably just adding it to settings.py. The model's help_text suggests the shape:

```python
BILLING_REMINDER_SCHEDULE = [14, 28]
```

Might be worth a Django system check too so a missing setting fails at startup instead of quietly inside a worker.

To confirm:

```bash
grep -rn "BILLING_REMINDER_SCHEDULE" app/eventyay/
```

Two hits, both reads, nothing defining it.

HuzaifaAbdulRehman:
Fixes #5526

`collect_billing_invoice()` reads `settings.BILLING_REMINDER_SCHEDULE` when it builds a BillingInvoice. That name was never set in `config/settings.py`, so the first event with paid orders last month raised AttributeError. `monthly_billing_collect()` only catches DatabaseError, so the rest of the organisers never ran.

I added `BILLING_REMINDER_SCHEDULE = [14, 28]`. Those are the calendar days the reminder task already uses. The shape comes from the model's help_text.

I left the catch block alone. Once the setting exists the crash is gone.

Tests in `tests/eventyay_common/test_billing_reminder_schedule.py` check the setting and that a paid order last month creates an invoice with that schedule. 2 passed on Postgres 16.

## Summary by Sourcery

Define the billing reminder schedule and verify that monthly billing invoice collection stores it correctly.

Bug Fixes:
- Define the billing reminder schedule so monthly invoice collection no longer fails when processing events with paid orders from the previous month.

Tests:
- Add coverage verifying the configured reminder days and their persistence on newly collected billing invoices.

## Summary by CodeRabbit

- **New Features**
  - Billing invoices now include reminder dates scheduled for the 14th and 28th of the month.
  - Unpaid invoice reminders follow a consistent twice-monthly schedule.

- **Tests**
  - Added coverage confirming the configured reminder dates.
  - Verified that the schedule is applied when billing invoices are created.

### DIFF
--- a/app/eventyay/config/settings.py
+++ b/app/eventyay/config/settings.py
@@ -1564,6 +1564,9 @@ EVENTYAY_OBLIGATORY_2FA = conf.obligatory_2fa
 EVENTYAY_SESSION_TIMEOUT_RELATIVE = 3600 * 3
 EVENTYAY_SESSION_TIMEOUT_ABSOLUTE = 3600 * 12
 
+# Calendar days of the month used for unpaid billing-invoice reminders.
+BILLING_REMINDER_SCHEDULE = [14, 28]
+
 # TODO: The `pdftk` tool should be auto-detected.
 PDFTK = ''
 
--- a/app/tests/eventyay_common/test_billing_reminder_schedule.py
+++ b/app/tests/eventyay_common/test_billing_reminder_schedule.py
@@ -4,7 +4,7 @@ from decimal import Decimal
 import pytest
 from dateutil.relativedelta import relativedelta
 from django.conf import settings
-from django_scopes import scopes_disabled
+from django_scopes import scope
 
 from eventyay.base.models import BillingInvoice, Order
 from eventyay.eventyay_common.tasks import collect_billing_invoice
@@ -26,7 +26,7 @@ def test_collect_billing_invoice_stores_reminder_schedule(event):
     last_month_date = _last_month_start()
     order_time = last_month_date + relativedelta(days=5)
 
-    with scopes_disabled():
+    with scope(organizer=event.organizer, event=event):
         Order.objects.create(
             event=event,
             code='BILL1',
--- a/app/tests/eventyay_common/test_billing_reminder_schedule.py
+++ b/app/tests/eventyay_common/test_billing_reminder_schedule.py
@@ -1,0 +1,45 @@
+from datetime import UTC, datetime
+from decimal import Decimal
+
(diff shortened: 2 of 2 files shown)


ArnavBallinCode (commented):
Thank you for your interest in contributing to the project.

We are closing this issue/PR because it does not currently align with the project roadmap, existing architecture, or an open issue that has been confirmed by the maintainers. We are receiving a number of externally generated issues and pull requests that appear to be created without sufficient project context, and we need to keep the issue tracker and review queue focused.

For new contributors, please start by working on existing confirmed issues before opening new issues or submitting related pull requests. This helps ensure that your contribution matches an actual project need and can be reviewed effectively.

Recommended next steps:

* Pick an existing open issue from the issue tracker.
* Submit a focused PR that directly addresses that issue.
* Avoid opening broad, speculative, or AI-generated issues/PRs.

Thank you again for your interest. We welcome contributions that are based on existing project issues and discussed with the maintainers first.

PR #5534, opened 2026-09-09
fix: keep start page events with talks_testmode=False

Rachit7168:
## Summary
- Featured/Upcoming/Past start-page queries used `.exclude(_settings_objects__key='talks_testmode', _settings_objects__value='True')`, which Django compiles as two independent `EXISTS` checks.
- That wrongly hid live events that had a `talks_testmode=False` settings row **and** any other setting with value `True` (common with payments/plugins), e.g. OpenTechSummit India on eventyay.com.
- Replace with a same-row `NOT EXISTS` via `Event.without_talks_testmode()`, and cover the regression in start-page tests.

## Test plan
- [x] `pytest tests/tickets/presale/test_startpage_optimization.py` (5 passed in docker)
- [ ] Reproduce locally: featured event + `talks_testmode=False` + any True setting still appears on `/` and `/upcoming/`
- [ ] Event with `talks_testmode=True` still excluded from start page
- [ ] After deploy, confirm ots2026 appears when Featured/Start page are ON

## Summary by Sourcery

Correct start-page event filtering so only events explicitly using talks test mode are excluded.

Bug Fixes:
- Keep live events with `talks_testmode=False` visible on featured, upcoming, and past start-page listings even when other settings are enabled.
- Continue hiding events whose talks component is explicitly in test mode.

Enhancements:
- Centralize start-page filtering to exclude only events with a matching `talks_testmode=True` setting.
- Add regression coverage for mixed settings and explicit talks test mode behavior.

Tests:
- Add start-page tests covering visibility with mixed settings and exclusion in talks test mode.

## Summary by CodeRabbit

* **Bug Fixes**
  * Events with talks test mode enabled are no longer shown in public event listings.
  * Events remain visible when talks test mode is disabled, even if other settings are enabled.
  * Filtering is consistently applied across start pages, upcoming events, past events, and followed events.
  * Event visibility is now handled consistently when multiple event settings are configured.

* **Tests**
  * Added coverage for correct event visibility in featured and upcoming listings.

### DIFF
--- a/app/eventyay/presale/views/startpage.py
+++ b/app/eventyay/presale/views/startpage.py
@@ -74,11 +74,10 @@ class StartPageView(TemplateView):
                 return ctx
 
             today_datetime = timezone.localtime().replace(hour=0, minute=0, second=0, microsecond=0)
-            base_qs = (
+            base_qs = Event.without_talks_testmode(
                 Event.objects.select_related('organizer')
                 .prefetch_related('_settings_objects')
                 .filter(live=True, testmode=False)
-                .exclude(_settings_objects__key='talks_testmode', _settings_objects__value='True')
             )
             future_filter = Q(date_to__gte=today_datetime) | Q(date_to__isnull=True, date_from__gte=today_datetime)
             past_filter = Q(date_to__lt=today_datetime) | Q(date_to__isnull=True, date_from__lt=today_datetime)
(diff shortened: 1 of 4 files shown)


ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):

app/eventyay/presale/views/startpage.py
```diff
@@ -74,11 +74,10 @@ def get_context_data(self, **kwargs):
                 return ctx
 
             today_datetime = timezone.localtime().replace(hour=0, minute=0, second=0, microsecond=0)
-            base_qs = (
+            base_qs = Event.without_talks_testmode(
```
Saksham-Sirohi, line 77: variable name is inappropriate; please fix it in a follow-up once tested

PR #5535, opened 2026-09-09
fix(common): show one ticketing action on event status page

### LINKED ISSUE #5533: Show either “Set up tickets” or “Manage tickets” based on ticket setup state
On the event status page, the **Event ticketing** section currently shows both:

```text
Set up tickets
Manage Tickets
```

This is confusing because these actions represent different states. The page should show only the action that matches the actual ticket setup status of the event.

Image

## Current behavior

On:

```text
Common event dashboard → Event status
```

the **Event ticketing** card shows both buttons at the same time:

```text
Set up tickets
Manage Tickets
```

This happens even when ticketing already appears to be configured and public sales are enabled.

## Expected behavior

The event status page should show only one primary ticketing action depending on whether tickets already exist.

### Case 1: No tickets configured

If the event has no ticket types configured yet, show:

```text
Set up tickets
```

Do not show:

```text
Manage Tickets
```

### Case 2: Tickets already configured

If the event has one or more ticket types configured, show:

```text
Manage Tickets
```

Do not show:

```text
Set up tickets
```

## Required changes

* Check whether the event has any ticket types configured.
* If no tickets exist, show only **Set up tickets**.
* If at least one ticket exists, show only **Manage Tickets**.
* Ensure the button label and target route match the event state.
* Keep the ticketing status badge, for example **Public sales**, separate from the setup/manage action.
* Apply the same logic regardless of ticketing mode:

  * private ticket test mode
  * public ticket sales
  * ticketing disabled/unpublished, if applicable

## Suggested behavior

```text
No ticket types exist:
[Set up tickets]

One or more ticket types exist:
[Manage Tickets]
```

## Acceptance criteria

* Events without tickets show **Set up tickets** only.
* Events with existing tickets show **Manage Tickets** only.
* The two buttons are never shown at the same time.
* **Set up tickets** links to the ticket creation/setup flow.
* **Manage Tickets** links to the ticket management page.
* The ticketing status badge remains visible independently of the button.
* Tests cover events with zero tickets and events with one or more tickets.

siddharthg2309:
## Summary

- render only one ticketing action in the **Event ticketing** card instead of both
- show **Set up tickets** while the event has no ticket type, **Manage Tickets** once at least one exists
- expose `has_ticket_products` from `EventLive.get_context_data` and branch the template on it
- leave the ticketing status badge and the ticketing-mode cards untouched

Both links were previously rendered unconditionally; only their button classes
swapped on `ticketing_ready`, so the card always offered two competing primary
actions. The new condition reads ticket types alone and never looks at
`testmode`, `private_testmode`, `tickets_published` or `live`, so the button is
identical in private test mode, public test mode, public sales and unpublished.
Reusing the existing short-circuit keeps the query count unchanged.

Fixes #5533

## Screenshots

**Event with no ticket types — only "Set up tickets" renders:**

image

**Event with one ticket type in public sales — only "Manage Tickets" renders, and the status badge is unaffected:**

image

## Testing

- `pytest tests/eventyay_common/test_event_live_ticket_actions.py -q`
  - 15 passed
- reverting the two source files makes 11 of those 15 fail, so they pin the behaviour rather than the implementation
- `pytest tests/eventyay_common/ -q`
  - 6 failed, 177 passed — identical failures with and without the change
    (`test_event_clone`, `test_event_private_testmode` mode-POST, `test_two_factor_auth`).
    They reproduce on unmodified `dev` in my local environment and are unrelated to this change.
- `ruff check` on `eventyay_common/views/event.py` reports the same 36 findings before and after
- `ruff check` and `ruff format --check` pass on the new test file
- manually verified on a local dev server: an event with no ticket type shows only **Set up tickets**;
  deleting/creating a ticket type flips the card between the two actions

 ## Risk

Adds one context variable and turns two unconditional links into an if/else in a
single template block. No model, URL, permission or migration change. The
`ticketing_ready` flag keeps its previous meaning and its other uses in the same
template are untouched, as is the status badge. Reverting the commit restores
both buttons.

Worth flagging separately: `QuickSetupView` refuses to run when
`products.exists() or quotas.exists()`, so an event holding a quota but no
ticket type shows **Set up tickets** and is redirected back with an info
message. That state predates this change, so I kept the condition to ticket
types as the issue specifies rather than widening it.

## Summary by Sourcery

Render a single context-appropriate ticketing action on the event status page based on whether ticket products exist.

Bug Fixes:
- Show exactly one ticketing action on the event status page: offer “Set up tickets” before any ticket products exist and “Manage Tickets” afterward.

Enhancements:
- Expose whether an event has ticket products in the live-event context while preserving the existing ticketing readiness and status badge behavior.

Tests:
- Add coverage for ticketing actions across all ticketing modes, including events without quotas and preservation of the status badge.


Rachit7168 (changes requested):
Image

Does not solve the issue properly

Rachit7168 (approved):
Looks good — the if/else on `has_ticket_products` is the right fix.

My earlier screenshot was from `dev` (both buttons still rendered). Sorry about that.


https://github.com/user-attachments/assets/963bb3c9-c253-4a2b-afe5-ec43faea2985

ArnavBallinCode (approved):

PR #5542, opened 2026-09-09
fix(orga): reject GET on state-changing views

### LINKED ISSUE #5541: Multiple organiser and control endpoints mutate state on GET, bypassing CSRF protection
##Description

Several endpoints under `orga/` and `control/` perform state-changing operations when accessed with a GET request.

Django's `CsrfViewMiddleware` only enforces CSRF protection for unsafe methods, so these GET requests can trigger database changes without a CSRF token.

This is not a permission bypass — the existing permission checks are still enforced. The problem is that a legitimately privileged user's browser can be made to perform these actions by visiting a third-party page, and some of these URLs are also exposed through normal links.

##Confirmed endpoints

| View | Location | Effect of GET |
|---|---|---|
| `ScheduleToggleView` | `eventyay/orga/views/schedule.py:185` | Publishes/unpublishes the public schedule and triggers the Celery sync task |
| `SubmissionSpeakersDelete` | `eventyay/orga/views/submission.py:250` | Removes a speaker from a proposal |
| `PhaseActivate` | `eventyay/orga/views/event.py:303` | Activates a review phase |
| `SubmissionTypeDefault` | `eventyay/orga/views/cfp.py:846` | Sets/clears the CfP default session type |
| `CfPQuestionToggle` | `eventyay/orga/views/cfp.py:654` | Activates/hides a CfP question |
| `EventActionDiscard` | `eventyay/control/views/event.py:1317` | Marks a `RequiredAction` as resolved |
| `EventAdminToken` | `eventyay/control/views/admin_views.py:259` | Generates and persists an admin JWT secret |
| `SubuserView` | `eventyay/orga/views/person.py:100` | Clears the caller's `is_superuser` flag |

##Root causes

There are two different patterns.

1. Discarded super().dispatch() result

Four views call super().dispatch() but ignore the returned response:

```
class ScheduleToggleView(EventPermissionRequired, View):
    permission_required = 'base.update_event'

        super().dispatch(request, *args, **kwargs)
        is_public = not self.request.event.get_feature_flag('show_schedule')
        self._set_schedule_public(self.request.event, is_public)
```

These views don't implement get() or post(). Normally View.dispatch() would return 405 Method Not Allowed, but that response is discarded and the mutation continues.

As a result, a GET can successfully change state while also producing a misleading Method Not Allowed log entry.

The affected views are:

- ScheduleToggleView
- SubmissionSpeakersDelete
- PhaseActivate
- SubmissionTypeDefault

They should instead move the mutation into a real post() handler:

```
def post(self, request, *args, **kwargs):
    is_public = not request.event.get_feature_flag('show_schedule')
    self._set_schedule_public(request.event, is_public)
    ...
    return redirect(request.event.orga_urls.schedule)
```

2. Explicit GET handlers performing writes

Other affected views directly modify state inside get().
For example, CfPQuestionToggle contains:
Legacy GET: toggle active

```
if request.method == http.HTTPMethod.GET:
    question.active = not question.active
    question.save(update_fields=['active'])
```

This should be moved to the POST path and the legacy GET mutation removed.

CSRF / accidental triggering

A privileged user's browser can be made to request one of these URLs from another page:

```/orga/event/<org>/<event>/schedule/toggle">```

No CSRF token is required because the request is a GET.

This can cause actions such as hiding a live event schedule, changing review state, hiding CfP fields, or removing a speaker.
ScheduleToggleView is particularly concerning because changing the schedule visibility also triggers the Celery task that synchronises the visibility with the tickets component.

##Templates

Some affected actions are currently exposed as normal links, for example:

```
- orga/templates/orga/submission/speakers.html:114
- orga/templates/orga/cfp/submissiontype/list.html:61,65
- orga/templates/orga/cfp/question/update.html:20,23
- orga/templates/orga/cfp/talkquestion/detail.html:44,47
- orga/templates/orga/cfp/talkquestion/update.html:20
- orga/templates/orga/settings/review.html:269
- control/templates/pretixcontrol/event/index.html:90
- control/templates/pretixcontrol/event/actions.html:10
- eventyay_common/templates/eventyay_common/event/index.html:37
- eventyay_common/templates/eventyay_common/event/meetup_dashboard.html:36
- orga/templates/orga/base.html:210

```
These should be converted to POST forms or the existing confirmation flow.

The schedule template is already using the correct pattern:

```
<form method="post" action="{{ request.event.orga_urls.toggle_schedule }}" class="d-inline">
    {% csrf_token %}
    <button type="submit" class="dropdown-item">
        {% translate "Hide schedule" %}
    </button>
</form>
```

However, ScheduleToggleView currently has no post() handler, so the view needs to be fixed as well.

## Expected behavior

- State-changing actions should only be performed through POST.
- POST requests should require a valid CSRF token.
- GET requests should return 405 Method Not Allowed and must not modify state.
- Destructive actions should use ActionConfirmMixin where appropriate.
- No Method Not Allowed warning should be generated on a successful request.

## Suggested fix

1. Add post() handlers and remove the four discarded super().dispatch() overrides.
2. Convert state-changing get() handlers to post().
3. Remove the legacy GET mutation in CfPQuestionToggle.
4. Use ActionConfirmMixin for destructive actions such as SubmissionSpeakersDelete and potentially PhaseActivate.
5. Convert the affected template links to POST forms/confirmation links.
6. Add regression tests confirming that GET returns 405 and does not change database state.
7. Add a lint/test check to prevent discarded super().dispatch() calls inside dispatch() overrides.

HuzaifaAbdulRehman:
Fixes #5541

## Summary

Several orga and control views wrote to the database on GET. Django skips CSRF on GET, so a logged-in organiser who loads one of those URLs from another page can hide a schedule, drop a speaker, or flip a review phase.

The discarded-dispatch pattern is in `ScheduleToggleView.dispatch` at `app/eventyay/orga/views/schedule.py:184` on `dev` (`43c97f9`). It calls `super().dispatch()` and then mutates anyway. `CfPQuestionToggle` at `cfp.py:654` still had a legacy GET write. The issue list is right. Two of its line numbers are off by one.

I moved the eight listed views to `post()`. GET now returns 405 and does not write. Templates submit POST forms with a CSRF token. `SubmissionType.urls.default` (`type.py:74`) now has a trailing slash so it matches the URLConf.

I skipped `ActionConfirmMixin` on speaker-remove and phase-activate, and I skipped a lint for discarded `super().dispatch()`. Those were extra to the CSRF hole. GET to these URLs is now a no-op. The organiser buttons still work.

## Test plan

- [x] WSL pytest, 21 CSRF cases: 21 passed in 95s
- [ ] GET on each listed URL returns 405 and does not write

## Summary by Sourcery

Require POST requests for state-changing organiser and control actions to close CSRF vulnerabilities from unsafe GET endpoints.

Bug Fixes:
- Prevent state-changing organiser and control views from mutating data when accessed with GET requests by requiring POST submissions.

Enhancements:
- Update affected interface actions to use CSRF-protected POST forms while preserving organiser workflows.
- Maintain JSON-based question toggles and correct submission-type default URL generation.

Tests:
- Add coverage confirming GET requests return 405 without changing state and POST requests continue to perform the intended actions.
- Add template checks to prevent nested forms in updated organiser pages.

## Summary by CodeRabbit

- **Security Improvements**
  - State-changing organizer and event-management actions now use CSRF-protected POST forms instead of GET links.
  - Direct GET requests to these actions are rejected to help prevent unintended changes.
  - Event admin access now requires a protected form submission.

- **Bug Fixes**
  - Submission-type default URLs now include a trailing slash.
  - Invalid requests receive a clear “Method Not Allowed” response.
  - Corrected form placement to prevent nested-form issues.

- **Tests**
  - Added coverage for POST behavior, CSRF protection, rejected GET requests, unchanged state, and form structure.

### DIFF
--- a/app/eventyay/base/models/type.py
+++ b/app/eventyay/base/models/type.py
@@ -71,7 +71,7 @@ class SubmissionType(OrderedModel, PretalxModel):
     class urls(EventUrls):
         """URL patterns for submission type views."""
         base = edit = '{self.event.cfp.urls.types}{self.pk}/'
-        default = '{base}default'
+        default = '{base}default/'
         delete = '{base}delete/'
         prefilled_cfp = '{self.event.cfp.urls.public}?submission_type={self.slug}'
 
--- a/app/eventyay/control/templates/control/event_list.html
+++ b/app/eventyay/control/templates/control/event_list.html
@@ -45,9 +45,12 @@
         </td>
         <td class="text-right video-admin-actions">
             {% if event.domain and event.admin_token %}
-                <a href="{% url 'eventyay_admin:video_admin:event.admin' pk=event.pk %}" class="btn btn-default btn-xs">
-                    <span class="fa fa-lock"></span> {% trans "Admin access" %}
-                </a>
+                <form method="post" action="{% url 'eventyay_admin:video_admin:event.admin' pk=event.pk %}" class="d-inline">
+                    {% csrf_token %}
+                    <button type="submit" class="btn btn-default btn-xs">
+                        <span class="fa fa-lock"></span> {% trans "Admin access" %}
+                    </button>
+                </form>
             {% endif %}
         </td>
         <td class="text-right video-admin-actions">
--- a/app/eventyay/control/templates/pretixcontrol/event/actions.html
+++ b/app/eventyay/control/templates/pretixcontrol/event/actions.html
@@ -7,10 +7,12 @@
         {% for action in actions %}
             <li class="list-group-item logentry">
                 <p>
-                    <a href="{% url "control:event.requiredaction.discard" event=request.event.slug organizer=request.event.organizer.slug id=action.id %}"
-                            class="btn btn-default btn-xs pull-right flip">
-                        {% trans "Hide message" %}
-                    </a>
+                    <form method="post" action="{% url 'control:event.requiredaction.discard' event=request.event.slug organizer=request.event.organizer.slug id=action.id %}" class="pull-right flip">
+                        {% csrf_token %}
+                        <button type="submit" class="btn btn-default btn-xs">
+                            {% trans "Hide message" %}
+                        </button>
+                    </form>
                     <small><span class="fa fa-clock-o"></span> {{ action.datetime|date:"SHORT_DATETIME_FORMAT" }}</small>
                 </p>
                 {{ action.display|safe }}
--- a/app/eventyay/control/templates/pretixcontrol/event/index.html
+++ b/app/eventyay/control/templates/pretixcontrol/event/index.html
@@ -87,10 +87,12 @@
                 {% for action in actions %}
                     <li class="list-group-item logentry">
                         <p>
-                            <a href="{% url "control:event.requiredaction.discard" event=request.event.slug organizer=request.event.organizer.slug id=action.id %}"
-                               class="btn btn-default btn-xs pull-right flip">
-                                {% trans "Hide message" %}
-                            </a>
+                            <form method="post" action="{% url 'control:event.requiredaction.discard' event=request.event.slug organizer=request.event.organizer.slug id=action.id %}" class="pull-right flip">
+                                {% csrf_token %}
+                                <button type="submit" class="btn btn-default btn-xs">
+                                    {% trans "Hide message" %}
+                                </button>
+                            </form>
                             <small><span
                                     class="fa fa-clock-o"></span> {{ action.datetime|date:"SHORT_DATETIME_FORMAT" }}
                             </small>
--- a/app/eventyay/control/views/admin_views.py
+++ b/app/eventyay/control/views/admin_views.py
@@ -257,6 +257,9 @@ class EventAdminToken(AdministratorPermissionRequiredMixin, DetailView):
             raise Http404("Event not found")
 
     def get(self, request, *args, **kwargs):
+        return self.http_method_not_allowed(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
         event = self.get_object()
 
         # Ensure JWT configuration exists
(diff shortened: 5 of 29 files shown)


Saksham-Sirohi (commented):
Closes CSRF-via-GET hole: ScheduleToggleView, speaker remove, review-phase activate, CfP/talk question toggles, submission-type default, control event actions moved to POST; templates use csrf_token forms. Nested-form pitfall avoided — question toggle forms live in warnings block before the edit <form>; review phase activate forms are sibling forms with form= attribute. GET→405 + no-mutation tests added. Existing EventPermissionRequired / control permissions preserved; no new unscoped event ORM. No human approval yet. Looks ready from my side; still needs another human approval before merge.

Rachit7168 (commented):
LGTM , Tested 

This is a security improvement. Restricting these state-changing actions to POST requests and implementing proper CSRF token forms in the templates effectively secures these endpoints against Cross-Site Request Forgery (CSRF).

Rachit7168 (approved):
LGTM , Tested

This is a security improvement. Restricting these state-changing actions to POST requests and implementing proper CSRF token forms in the templates effectively secures these endpoints against Cross-Site Request Forgery (CSRF).

Rachit7168 (approved):
Great work on this! 🚀 

This is a solid security improvement. Restricting these state-changing actions to `POST` requests and implementing proper CSRF token forms in the templates effectively secures these endpoints against Cross-Site Request Forgery (CSRF). 

A few things I really liked about this PR:
- **Clean Implementation:** Dropping `get`/`dispatch` and relying on Django's base `View` behavior to return `405 Method Not Allowed` for `GET` requests to `post` methods is exactly the right way to handle this.
- **Template Cleanup:** Swapping out the anchor `<a>` tags for inline `<form>` elements with buttons was done neatly without disrupting the UI structure.
- **Test Coverage:** Updating the test suites to explicitly assert `405` responses on `GET` requests ensures we don't accidentally regress on this in the future.

I've reviewed the diffs and everything looks good to me. LGTM! Thanks for the contribution! 🎉

PR #5543, opened 2026-09-09
fix: require POST for state-changing organiser and control endpoints

### LINKED ISSUE #5541: Multiple organiser and control endpoints mutate state on GET, bypassing CSRF protection

tarone-saloni:
Fixes #5541

## Description

This PR updates all 8 endpoints covered by the issue, plus one additional endpoint discovered during the review.

### One thing the issue got wrong

The issue mentions that these endpoints are not permission bypasses. That's true for the 8 endpoints listed there, but OutboxSend `(orga/views/mails.py) ` is different.

For the single-mail path, dispatch() returns before `super().dispatch()` is called. Because of that, `permission_required = 'base.send_queuedmail' `is never checked for:

``` /orga/event/<org>/<event>/mails/outbox/<pk>/send ```

EventPermissionMiddleware._handle_orga_url only checks whether the user is authenticated, so any logged-in user could send a queued mail using GET.

I moved this to the same POST/confirmation flow and added `test_reviewer_cannot_send_queued_mail `to cover it.

### Fixes

Removed dispatch() overrides that were dropping super().dispatch()

For these views, the state-changing code is now in post():

- ScheduleToggleView
- SubmissionSpeakersDelete
- PhaseActivate
- SubmissionTypeDefault

GET now returns 405 and doesn't perform the action.

Moved state-changing GET handlers to POST

## Changed:

- CfPQuestionToggle
- EventActionDiscard
- EventAdminToken
- SubuserView

CfPQuestionToggle still returns the same JSON response for AJAX requests. It now accepts a form-encoded POST with active instead of using the old GET branch.

EventAdminToken was also changed from DetailView to `SingleObjectMixin`, View. It doesn't need to render a template on GET, and the old template_name was actually pointing to the unrelated "clear event data" page.

## Confirmation flow

ActionConfirmMixin is now used for:

- SubmissionSpeakersDelete
- PhaseActivate
- single-mail OutboxSend
- This matches the existing pattern used by EventDelete, MailDelete and OrganizerDelete.

The template links themselves stay the same; they now take the user to the confirmation page before the action is performed.

Templates:
Converted the affected links from the issue to CSRF-protected POST forms in 10 templates.

`orga/schedule/index.html` didn't need a template change because it was already posting. The view just didn't have a post() handler for it.

## Not included
I didn't add the suggested lint/test check for discarded `super().dispatch()` calls.

That would need a separate flake8 plugin or an AST-based test, so I'd rather keep that as a follow-up instead of adding it to this PR.

## Other things I found

1. `SubmissionType.urls.default` was missing its trailing `slash. APPEND_SLASH` hid this for GET, but with POST it would result in a 301 and the request body would be lost. Without fixing this, the new POST button would have silently failed.

2. `LocaleSet (cfp/views/locale.py, presale/views/locale.py)` also changes request.user.locale on GET. This is intentional language-switcher behavior and only changes the user's UI language, so I left it as-is. If we want to make the sweep completely exhaustive, this can be changed separately.

3. `MailSettingsRendererPreview` and the Gmail OAuth callbacks looked suspicious initially, but after checking them they don't need changes.

4. There's also an unrelated issue with `trigger_public_schedule()`: the Celery sync calls it with an `organiser_slug` kwarg that it doesn't accept, and the resulting TypeError is swallowed by the broad except Exception. I didn't change this here; it should probably be a separate issue.

## Tests
Added `tests/talk/orga/views/test_orga_views_state_changing_get.py` with:

-8 tests checking that GET doesn't change state
-1 test checking that a reviewer cannot send a queued mail
Also updated 11 existing tests from GET to POST.

I ran the 7 affected test files against a clean dev checkout and compared the failures with the baseline. There are no new failures.

The existing failures are:

-NoReverseMatch from tests missing the organizer slug (#3613)
-missing frontend build artifacts

### DIFF
--- a/app/eventyay/base/models/type.py
+++ b/app/eventyay/base/models/type.py
@@ -71,7 +71,7 @@ class SubmissionType(OrderedModel, PretalxModel):
     class urls(EventUrls):
         """URL patterns for submission type views."""
         base = edit = '{self.event.cfp.urls.types}{self.pk}/'
-        default = '{base}default'
+        default = '{base}default/'
         delete = '{base}delete/'
         prefilled_cfp = '{self.event.cfp.urls.public}?submission_type={self.slug}'
 
--- a/app/eventyay/control/templates/control/event_list.html
+++ b/app/eventyay/control/templates/control/event_list.html
@@ -45,9 +45,12 @@
         </td>
         <td class="text-right video-admin-actions">
             {% if event.domain and event.admin_token %}
-                <a href="{% url 'eventyay_admin:video_admin:event.admin' pk=event.pk %}" class="btn btn-default btn-xs">
-                    <span class="fa fa-lock"></span> {% trans "Admin access" %}
-                </a>
+                <form action="{% url 'eventyay_admin:video_admin:event.admin' pk=event.pk %}" method="post" class="d-inline">
+                    {% csrf_token %}
+                    <button type="submit" class="btn btn-default btn-xs">
+                        <span class="fa fa-lock"></span> {% trans "Admin access" %}
+                    </button>
+                </form>
             {% endif %}
         </td>
         <td class="text-right video-admin-actions">
--- a/app/eventyay/control/templates/pretixcontrol/event/actions.html
+++ b/app/eventyay/control/templates/pretixcontrol/event/actions.html
@@ -7,10 +7,13 @@
         {% for action in actions %}
             <li class="list-group-item logentry">
                 <p>
-                    <a href="{% url "control:event.requiredaction.discard" event=request.event.slug organizer=request.event.organizer.slug id=action.id %}"
-                            class="btn btn-default btn-xs pull-right flip">
-                        {% trans "Hide message" %}
-                    </a>
+                    <form action="{% url 'control:event.requiredaction.discard' event=request.event.slug organizer=request.event.organizer.slug id=action.id %}"
+                          method="post" class="pull-right flip">
+                        {% csrf_token %}
+                        <button type="submit" class="btn btn-default btn-xs">
+                            {% trans "Hide message" %}
+                        </button>
+                    </form>
                     <small><span class="fa fa-clock-o"></span> {{ action.datetime|date:"SHORT_DATETIME_FORMAT" }}</small>
                 </p>
                 {{ action.display|safe }}
--- a/app/eventyay/control/templates/pretixcontrol/event/index.html
+++ b/app/eventyay/control/templates/pretixcontrol/event/index.html
@@ -87,10 +87,13 @@
                 {% for action in actions %}
                     <li class="list-group-item logentry">
                         <p>
-                            <a href="{% url "control:event.requiredaction.discard" event=request.event.slug organizer=request.event.organizer.slug id=action.id %}"
-                               class="btn btn-default btn-xs pull-right flip">
-                                {% trans "Hide message" %}
-                            </a>
+                            <form action="{% url 'control:event.requiredaction.discard' event=request.event.slug organizer=request.event.organizer.slug id=action.id %}"
+                                  method="post" class="pull-right flip">
+                                {% csrf_token %}
+                                <button type="submit" class="btn btn-default btn-xs">
+                                    {% trans "Hide message" %}
+                                </button>
+                            </form>
                             <small><span
                                     class="fa fa-clock-o"></span> {{ action.datetime|date:"SHORT_DATETIME_FORMAT" }}
                             </small>
--- a/app/eventyay/control/views/admin_views.py
+++ b/app/eventyay/control/views/admin_views.py
@@ -27,6 +27,7 @@ from django.views.generic import (
     UpdateView,
     View,
 )
+from django.views.generic.detail import SingleObjectMixin
 
 from eventyay.base.models import (
     BBBCall,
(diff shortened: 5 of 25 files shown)


Rachit7168 (changes requested):
Please address coderabbit's review

PR #5550, opened 2026-09-10
chore: sync base model migrations

Rachit7168:
## Summary
Adds the missing `base` migration that keeps Django’s migration history in sync with current `base` models.

This does **not** change product behavior or data. It fixes model/migration drift that shows up on both existing and fresh databases when running:

```bash
python manage.py makemigrations --check --dry-run
# or
python manage.py makemigrations base --dry-run
```

## What was missing
On current `dev`, all existing migrations were already applied (`migrate` reported nothing to apply), but Django still detected **unmigrated model state** in `base`.

In other words:
- Not a case of “migration file exists but wasn’t applied”
- A case of “models changed (or auto-generated state drifted) and no migration was committed”

Reproducible locally with Docker and on a fresh DB.

### Detected by Django before this PR
`makemigrations` wanted a new `base` migration depending on `0070_alter_voucher_budget_alter_voucher_value`, with:

1. **Gmail OAuth index renames** on `GmailOAuthCredential`
2. **Metadata field updates** on a few `base` models

The `socialmedia` plugin also showed similar drift (`platform_username` help_text). That is **out of scope** here and belongs in the plugin repo.

## Why the Gmail indexes drifted
`0064_gmailoauthcredential` created indexes with explicit names:

- `base_gmailo_event_i_6f0d0d_idx` on `['event', 'is_active']`
- `base_gmailo_is_acti_0d8f8f_idx` on `['is_active']`

The current model defines the same indexes **without** explicit `name=` values:

```python
indexes = [
    models.Index(fields=['event', 'is_active']),
    models.Index(fields=['is_active']),
]
```

Django’s auto-generated names for those indexes are now:

- `base_gmailo_event_i_133776_idx`
- `base_gmailo_is_acti_22cf9d_idx`

So migration state still had the old names, while current model state expects the new ones. Some environments already had the new physical index names, which is why a plain `RenameIndex` could fail with “relation … does not exist”.

## What this migration does
`app/eventyay/base/migrations/0071_sync_model_state.py`

### 1. Sync Gmail OAuth index names
- Updates Django state: old index names → new index names
- Updates the database only when needed:
  - skip if the new name already exists
  - skip if the old name is missing
  - otherwise rename using Django’s backend-aware `schema_editor.rename_index()`
    - PostgreSQL: rename in place
    - SQLite: drop + recreate (SQLite does not support `ALTER INDEX ... RENAME TO`)

### 2. Sync metadata-only field definitions
These `AlterField` operations align migration state with model text. They do **not** introduce new defaults or schema type changes:

| Model | Field | What changed in state |
|---|---|---|
| `InvoiceVoucher` | `max_usages` | `help_text` / `verbose_name` (default remains `1`) |
| `InvoiceVoucher` | `price_mode` | `choices` / `verbose_name` (default remains `'none'`) |
| `StreamSchedule` | `stream_type` | `choices` / `verbose_name` (default remains `'youtube'`) |
| `Team` | `can_video_manage_content` | `help_text` / `verbose_name` (default remains `False`) |

So this is a **state sync**, not a billing/access behavior change.

## Out of scope
- `socialmedia` plugin migration for `SocialMediaAccount.platform_username` help_text  
  (`socialmedia/migrations/0005_...`) — plugin-owned, not Eventyay core

## Risk notes
- Low risk: mostly migration-state alignment
- Index rename is conditional and backend-aware
- Defaults in the `AlterField` ops match existing model defaults; they are included because Django serializes the full field definition

## Test plan
- [x] `docker compose exec web python manage.py migrate base` applies `0071_sync_model_state`
- [x] `docker compose exec web python manage.py makemigrations base --check --dry-run` → `No changes detected in app 'base'`
- [x] Index rename path skips when new names already exist (local Postgres case)
- [x] SQLite-safe rename via Django `schema_editor.rename_index()` (no raw `ALTER INDEX`)
- [ ] Fresh DB can migrate cleanly through `0064` → `0071`
- [ ] Confirm `socialmedia` drift remains unchanged / out of scope

## Summary by Sourcery

Synchronize base model migration state and Gmail OAuth indexes without changing application behavior or persisted data.

Bug Fixes:
- Synchronize the base app’s migration state with the current model definitions to eliminate recurring `makemigrations` drift.

Enhancements:
- Safely reconcile Gmail OAuth index names across PostgreSQL and SQLite while accommodating databases that already use either naming scheme.
- Align voucher, stream schedule, and team field metadata without changing their defaults or product behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database upgrade reliability by safely synchronizing existing Gmail authentication indexes.
  - Updated voucher, schedule, and team data definitions to keep stored data aligned with the application.
  - Added reversible migration handling to support safer upgrades and rollbacks across supported database systems.


Saksham-Sirohi (approved):
Gmail index rename is idempotent and the `AlterField` ops are state metadata. Generated on Django 6.0.7 against the 5.2 runtime; the operations look compatible. This and #5365 both add `0071_*` depending on `0070` — renumber whichever lands second. Not blocking the migration content.

ArnavBallinCode (approved):

PR #5571, opened 2026-09-11
feat: enforce email.bulk.monthly capabilities

ArnavBallinCode:
## Description

This PR implements the missing enforcement hook for the `email.bulk.monthly` tier capability.

### What is restricted?
It hooks into `ComposeMailBaseView.post()`, which means it successfully restricts **mass bulk emails sent via the Organizer dashboard**:
- Mass emails sent to **Speakers / Sessions**
- Mass emails sent to **Event Team members / Staff**

### What is NOT restricted?
It intentionally does **not** restrict transactional emails:
- Automated post-purchase ticket emails
- Queued transactional emails (password resets, payment reminders, etc.)
- (Currently, mass emails sent to Attendees via the `sendmail` plugin are handled separately and bypass this specific base view hook).

Depends on plugin PR: https://github.com/fossasia/eventyay-business/pull/58

## Summary by Sourcery

Enforce monthly bulk email entitlements across Organizer bulk email sending flows.

New Features:
- Enforce the `email.bulk.monthly` entitlement for Organizer bulk email composition and draft delivery.

Enhancements:
- Record bulk email entitlement usage atomically when drafts are moved to the outbox.
- Add coverage for entitlement-denied bulk email composition and draft delivery.

Tests:
- Add tests verifying denied bulk emails are not saved or sent and drafts remain unchanged.

## Summary by CodeRabbit

- **New Features**
  - Added entitlement checks when moving draft bulk emails to the outbox.
  - Approved emails now record usage when queued for delivery.
  - Draft emails can still be saved without passing the entitlement check.

- **Bug Fixes**
  - Prevented bulk emails from being queued when entitlement is insufficient.
  - Blocked transitions leave drafts unchanged and unsent.
  - Added a fallback error message when entitlement details are unavailable.

### DIFF
--- a/app/eventyay/orga/views/mails.py
+++ b/app/eventyay/orga/views/mails.py
@@ -14,6 +14,7 @@ from django.views.generic import FormView, ListView, TemplateView, View
 from django_context_decorator import context
 from i18nfield.strings import LazyI18nString
 
+from eventyay.base.entitlements import check_entitlement
 from eventyay.base.models.mail import MailTemplate, QueuedMail, get_prefixed_subject
 from eventyay.base.signals import entitlement_usage_recorded
 from eventyay.common.exceptions import SendMailException
@@ -617,6 +618,18 @@ class ComposeMailBaseView(EventPermissionRequired, FormView):
             message = form.empty_audience_draft_error if is_draft else form.empty_audience_error
             form.add_error(None, message)
             return self.render_to_response(self.get_context_data(form=form))
+
+        if not is_draft:
+            recipients_count = len(form.get_recipients())
+            decision = check_entitlement(
+                self.request.event.organizer,
+                'email.bulk.monthly',
+                event=self.request.event,
+                quantity=recipients_count
+            )
+            if not decision.allowed:
+                form.add_error(None, decision.message)
+                return self.render_to_response(self.get_context_data(form=form))
 
         with transaction.atomic():
             result = form.save()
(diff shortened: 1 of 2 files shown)


Rachit7168 (changes requested):
Image

Cant change and save another teir

Rachit7168 (approved):
LGTM , tested

Saksham-Sirohi (approved):
Looks correct for the missing `email.bulk.monthly` enforcement hook.

On `dev`, compose already **records** usage on non-draft send but never **checked** the cap first, and `DraftToOutbox` did neither — so drafts were a free bypass. This PR closes both gaps: deny before `form.save()`, and check+record (with idempotency key) when activating a draft. Denial tests for compose and draft→outbox look good; drafts themselves stay unmetered until activation.

Ship together with fossasia/eventyay-business#58 (monthly usage aggregation on check) — without that plugin change, `check_entitlement` still won’t see prior usage in the billing month.

Non-blocking notes:
- Check uses `len(form.get_recipients())` while recording uses `len(result)`; prefer one recipient list (or check `len(result)` inside the same atomic block if save is side-effect-free enough) so the metered quantity can’t drift.
- Classic check-then-record TOCTOU remains if two concurrent bulk sends race; atomic reserve belongs in the business plugin if you need hard caps under concurrency.
- Attendee `sendmail` plugin path remains out of scope as documented — track as follow-up if that surface should share the same monthly cap.

Issue #5596, opened 2026-09-13
Missing Data when Validation Fails in CFP


SawanTeja:
When submitting the Call for Speakers (CFP) form with a missing mandatory field, validation fails and the page reloads, causing three data-retention issues:
1. **Text fields are cleared:** Form auto-save scripts incorrectly clear text data upon seeing a validation error.
2. **Images disappear:** Uploaded images do not display the `Currently: <filename>` fallback link, making it appear as if the image was lost. 
3. **Slides error:** If the user uploaded a slide, fixing the missing text field and clicking "Continue" again triggers an incorrect validation error: `"You can add at most 1 slide entries."` (caused by the browser silently resubmitting the file alongside the session file).

**Expected behavior**
When a form validation error occurs:
- Text fields should retain the user's input.
- File upload fields (like Session Image) should clearly show `Currently: <filename>` to indicate the file was safely preserved in the session.
- Re-submitting the form to correct errors should not trigger false-positive limit errors for multi-file fields like Slides.

https://github.com/user-attachments/assets/215f8555-54ec-4947-b670-1555e7a4b157

**Additional context**
The issue is easily reproducible by filling out the CFP Session/Profile form, uploading an image and a slide, leaving one mandatory text field blank, and clicking "Continue".

Rachit7168:
Completed via #5597

PR #5597, opened 2026-09-13
fix(cfP): file retention on form validation failures

SawanTeja:
**Fixes:** #5596 

## Description
This PR resolves data-loss and file retention bugs in the CFP form that occur when a user fails validation due to missing mandatory fields.

### Changes made:
* **Prevented text data loss:** Added a check in `formAutoSave.js` to ensure auto-saved local storage is not cleared on page load if Django validation errors (`.alert-danger`) are present on the page.
* **Fixed "lost" image illusion:** Updated `GenericFlowStep.post()` to properly inject session file metadata into `form.initial` and invalidate the Django `_bound_fields_cache`. This allows the `ClearableFileInput` widget to correctly render the `Currently: <filename>` link instead of an empty file input on validation failures.
* **Resolved duplicate slide errors:** Modified `get_form()` to deduplicate incoming multi-file uploads (by name and size) against existing session files. This prevents browsers that natively retain form state from silently submitting duplicates and incorrectly triggering the `You can add at most 1 slide entries` error.

## Testing Instructions
1. Fill out the CFP Session/Profile form.
2. Upload a Session Image and a Slide file.
3. Intentionally leave a mandatory text field blank and click "Continue".
4. **Verify:** Text inputs are retained, the image field displays `Currently: <filename>`, and clicking "Continue" a second time does not throw a slide limit error.

https://github.com/user-attachments/assets/e1820be8-3047-444e-9369-71465d6a7721

## Summary by Sourcery

Preserve CFP form data and uploads when validation fails, while preventing duplicate file submissions from causing spurious validation errors.

Bug Fixes:
- Preserve auto-saved form data when validation errors are displayed, preventing text input loss on refresh.
- Retain uploaded files across CFP validation failures and restore their visible filenames in file fields.
- Prevent duplicate multi-file submissions from triggering erroneous slide-count validation errors.

Enhancements:
- Improve CFP form handling for session-backed uploads and invalid submissions.

## Summary by CodeRabbit

* **Bug Fixes**
  * File uploads are now preserved when form validation fails, allowing users to review and resubmit them without re-uploading.
  * Duplicate uploads are filtered out when their names and sizes match existing uploads.
  * Replacing a previously uploaded single file now correctly updates the saved upload.
  * Saved form data is retained when validation errors are present, preventing accidental loss of entered information.

### DIFF
--- a/app/eventyay/cfp/flow.py
+++ b/app/eventyay/cfp/flow.py
@@ -318,11 +318,20 @@ class FormFlowStep(TemplateFlowStep):
             if field.endswith('_files'):
                 for f in file_list:
                     existing_files = files.getlist(field)
-                    is_dup = any(
-                        getattr(existing_file, 'name', None) == getattr(f, 'name', None) and 
-                        getattr(existing_file, 'size', None) == getattr(f, 'size', None) 
-                        for existing_file in existing_files
-                    )
+                    is_dup = False
+                    for existing_file in existing_files:
+                        if getattr(existing_file, 'name', None) == getattr(f, 'name', None) and \
+                           getattr(existing_file, 'size', None) == getattr(f, 'size', None):
+                            
+                            f_content = f.read()
+                            existing_content = existing_file.read()
+                            f.seek(0)
+                            existing_file.seek(0)
+                            
+                            if f_content == existing_content:
+                                is_dup = True
+                                break
+
                     if not is_dup:
                         files.appendlist(field, f)
             else:
@@ -391,6 +400,13 @@ class FormFlowStep(TemplateFlowStep):
                                 current = [current]
                             existing[field] = current + new_entries
                         else:
+                            old_entry = existing.get(field)
+                            if old_entry and isinstance(old_entry, dict) and 'tmp_name' in old_entry:
+                                try:
+                                    self.file_storage.delete(old_entry['tmp_name'])
+                                except Exception:
+                                    pass
+                            
                             existing[field] = new_entries if len(new_entries) > 1 else new_entries[0]
                             # Inject initial data so the widget can render the "Currently: link"
                             if hasattr(form, 'initial'):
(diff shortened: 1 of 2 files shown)


app/eventyay/cfp/flow.py
```diff
@@ -317,7 +317,23 @@ def get_form(self, from_storage=False):
         for field, file_list in self.request.FILES.lists():
             if field.endswith('_files'):
                 for f in file_list:
-                    files.appendlist(field, f)
+                    existing_files = files.getlist(field)
+                    is_dup = False
+                    for existing_file in existing_files:
+                        if getattr(existing_file, 'name', None) == getattr(f, 'name', None) and \
```
Rachit7168, line 323: The logic you added to retain files on validation failure is roughly 45 lines long and is duplicated exactly in both FormFlowStep.post() and ProfileStep.post().

Extract this logic into a helper method on the base FormFlowStep class to keep the code DRY and maintainable.

app/eventyay/cfp/flow.py
```diff
@@ -317,7 +317,23 @@ def get_form(self, from_storage=False):
...
+                           getattr(existing_file, 'size', None) == getattr(f, 'size', None):
```
Rachit7168, line 324: In FormFlowStep.get_form(), you are checking for duplicate files by reading the entire file contents into memory. 

Suggestion: Reading the entire contents of a file into memory at once is dangerous and could easily lead to a MemoryError or severe memory spikes on the server if the user uploads large files (like presentation decks or videos).

In the context of a form upload, if both the file name and the exact file size in bytes match, it is standard practice to safely assume it is the exact same file. You should remove the f.read() content comparison entirely:

Rachit7168 (changes requested):
Please address the comments and the issue is that image is not being saved after validation error. The other fields are still shown

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/8c66808f-a5ef-4b7d-8908-1a5145fe098e

Saksham-Sirohi (approved):

PR #5604, opened 2026-09-13
fix(settings): remove unused GeoIP country lookup

Rachit7168:
## Summary
- Part of #1546
- `get_country_from_request()` was the only `HAS_GEOIP` / GeoIP2 consumer and had **no callers**
- Live forms already use `guess_country(event)`
- Delete the dead helper and related imports instead of adding unused settings

**Before:** Undefined `settings.HAS_GEOIP` risked `AttributeError` if the dead helper were ever called.

**After:** Dead GeoIP path is removed; no GeoIP settings needed.

## Test plan
- [ ] Confirm `get_country_from_request`, `HAS_GEOIP`, and `GEOIP_PATH` no longer appear in the codebase
- [ ] Spot-check question/invoice country defaults still use `guess_country(event)`

## Summary by CodeRabbit

* **Refactor**
  * Removed the request-based country lookup functionality from the event question forms.
  * Country defaults are no longer determined using the requester’s IP address, providing a more consistent form experience without GeoIP-based detection.


Aqil-Ahmad (approved):

ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):
`get_country_from_request` has no callers. Invoice and question forms already default country through `guess_country(event)`. Removing the dead GeoIP path is the right fix; no `HAS_GEOIP` setting is needed.

PR #5605, opened 2026-09-13
fix(shift-schedule-editor): fix desktop modal width, shift card clipping, and mobile label

### LINKED ISSUE #5601: [Shift Schedule Editor] Modals and Shift cards broken after mobile layout update
### Description

The recent mobile layout update to the schedule editor introduced two visual regressions in the shift schedule editor on desktop.

### Cause

The mobile layout PR changed how the schedule editor handles sizing and overflow. These changes had unintended side effects on the desktop layout of the shift schedule editor — specifically the edit/assign modals and the shift card role rows.

Image

Image

### Current Behavior

- Opening the edit or assign members modal causes it to stretch nearly full screen width, clipping behind the left sidebar and cutting off the modal content on the left side.
- On the shift cards in the grid, the role capacity badge (e.g. `1/9 assigned`) overflows and gets clipped at the right edge of the card — only part of the text is visible.
- For unassigned shifts, the mobile collapsible header shows 'Unassigned Sessions' as a shared component.

### Expected Behavior

- Modals should appear centred on screen at a reasonable fixed width, fully visible and not obscured by the sidebar.
- The capacity badge on each role row inside a shift card should always be fully visible. If the role name is long, it should truncate with an ellipsis rather than pushing the badge outside the card.
- The header should appear as Unassigned Shifts as a context-aware string.

MukundC25:
Closes #5601 

**Summary**

Fixes regressions introduced by the mobile layout update to the schedule editor, all limited to the shift schedule editor or the shared adapter layer.

**What changed**

- Desktop modals (edit shift, assign members) were expanding to nearly full viewport width and clipping behind the sidebar. Stylus was intercepting the CSS `min()` function call and silently outputting `95vw` instead of `min(680px, 95vw)`. Fixed by using `unquote()` so the native CSS function is emitted correctly.

- The capacity badge (`N/N assigned`) on shift cards was overflowing the right edge of the card on narrow grid columns. The role name side of the row now truncates with an ellipsis, and the badge and restricted tag are `flex-shrink: 0` so they are always fully visible.

- The mobile collapsible drawer showed "Unassigned Sessions" in both the talk and shift editors. The shift editor now correctly shows "Unassigned Shifts", following the same `caps.showRoles` pattern already used for the filter placeholder label.

https://github.com/user-attachments/assets/bcb194ce-349d-4cb4-8508-f236b46ae170

## Summary by Sourcery

Fix schedule editor regressions across desktop layout, shift card contents, mobile labels, and save error handling.

Bug Fixes:
- Correct desktop schedule editor modal sizing, prevent shift card capacity information from being clipped, and display the appropriate unassigned-item label on mobile.
- Show a dismissible error message when saving an editor item fails and handle non-JSON API error responses more clearly.

Enhancements:
- Improve schedule editor save handling so failed saves preserve the editor state while successful saves refresh schedule data reliably.

## Summary by CodeRabbit

* **Bug Fixes**
  * Mobile unassigned panel titles now use the correct translated label.
  * Saving changes handles failures more reliably, keeps the editor open, and displays a dismissible error message.
  * Clearer errors are shown when unexpected server responses prevent data from loading.
* **Style**
  * Long role names now display more cleanly with improved spacing and truncation.
  * Role badges and restricted labels maintain their layout more consistently.
  * The editor modal width now adapts correctly to smaller screens.


Saksham-Sirohi (approved):
Shift UI stays separated. The unassigned label and the full schedule refetch are gated on `caps.showRoles`, and the card clipping is only in the teamshifts adapter. Talk mode still says Unassigned Sessions, does not replace `schedule.value` after save, and still skips title writes when the slot has a `submission`. The shared modal width and save-error banner apply to both editors, but they do not change talk payloads or endpoints. Talk API responses are `JsonResponse`, so the content-type check does not reject them.

PR #5618, opened 2026-09-14
feat(privacy): manage third-party services from the admin

sanketkokate712:
Part of #5414. This builds on #5438, so until that one merges the diff below also shows its commits. The new work here is the last commit.

While testing #5438 I noticed there was no way to actually fill the third-party service registry from the admin. The Privacy & Compliance page listed services, but you could only add, change or remove one from the Django shell. This PR adds the missing pages.

**What changed**

- Add, edit and delete pages for third-party services, linked from the services table on Privacy & Compliance.
- The three registry fields from the issue that the model didn't have yet: data processed, country / region, and DPA status. They show up in the table too.
- The consent category is required when you add a service. The model still lets it be blank, but a service without a category never makes it into the banner, which means it's never blocked. I'd rather stop that at the form than warn about it later.
- The service identifier can't be edited after the service is created. Blocked scripts are matched on that identifier, so renaming a live service would quietly stop blocking them.

**Before**, the table was read-only:

**After**:

**Testing**

`tests/base/test_privacy_service_admin.py` covers add, edit and delete, the required category, the locked identifier, a new service actually reaching the consent banner, and that all three pages return 403 without an admin session. I ran it together with the existing consent tests and all 26 pass.

To try it locally, open Admin → Privacy & Compliance and use **Add service**.

**One thing I noticed along the way**

When I generated the migration, `makemigrations` also wanted to alter `question.type` and `talkquestion.variant`. Those choices changed on `dev` without a migration. I left them out so this PR only touches the privacy models, but someone should probably add that migration separately.

## Summary by Sourcery

Enable administrators to configure privacy consent and manage the third-party services that are shown and controlled by the consent layer.

New Features:
- Add an administrator-managed third-party service registry with create, edit, and delete workflows.
- Add privacy and consent configuration for built-in or external consent management, including policy links and optional consent categories.
- Gate third-party embeds and scripts behind visitor consent, with support for withdrawal and external CMP callbacks.

Bug Fixes:
- Prevent insecure external CMP scripts and invalid stored privacy links from being emitted to public pages.
- Preserve unknown or malformed PayPal endpoint values for validation instead of silently coercing them to the live endpoint.

Enhancements:
- Extend third-party service metadata with processed data, processing region, DPA status, cookie names, and consent classification.
- Require consent categories for new services and prevent editing identifiers used to match blocked scripts.
- Expose Privacy & Compliance administration in global navigation and add contextual warnings for incomplete consent configuration.

Tests:
[...]

PR #5626, opened 2026-09-14
feat(privacy): record privacy configuration changes in an audit log

sanketkokate712:
Part of #5414. Depends on #5438 and #5618.

Only the last commit belongs to this PR. The first one is #5438 squashed and the second is #5618. Both drop out when I rebase onto dev once those land, which leaves this as a single commit.

Right now nobody can tell who changed the consent setup or when. If someone switches the consent provider, turns a category off, or disables a third-party service, nothing is recorded. For a feature whose whole point is compliance, that felt like a gap, and section 14 of the issue asks for it.

**What changed**

- Changing the consent provider, turning a consent category on or off, editing the external CMP settings, and adding, editing, disabling or deleting a third-party service each writes a log entry. The entry stores the administrator and, where it applies, the old and new values.
- There's a new **Privacy audit log** page, linked from the top of Privacy & Compliance.
- It uses the existing `LogEntry` table instead of adding a new model, so there's no migration.

Saving the settings form without changing anything writes nothing. I got that wrong at first: an unset setting reads back as `None` but the form saves it as `''`, so the first save logged a change nobody made. There's a test for it now.

**Before**, the page didn't exist:

**After**, these entries came from real changes made through the admin forms:

**Testing**

`tests/base/test_privacy_audit_log.py` covers each kind of change, the old and new values, a no-op save logging nothing, the page listing entries, and the page returning 403 without an admin session. It passes together with the consent and service-admin tests (33 in total).

One design note for review: global settings aren't a database object, so a settings change is attached to the administrator who made it. The page filters on the action type rather than the object, so this doesn't affect what it shows. Happy to change that if there's a better anchor.

## Summary by Sourcery

Implement privacy and consent management with administrator controls, consent-aware public behavior, and comprehensive audit logging.

New Features:
- Add administrator-managed privacy and consent configuration, including consent providers, categories, legal links, external CMP settings, and third-party service records.
- Provide a public consent-management integration with Klaro, external CMP support, service categorization, cookie handling, and consent-gated embeds.
- Add Privacy & Compliance administration pages for configuring settings and managing third-party services.
- Add a paginated Privacy audit log page showing administrators, timestamps, and privacy configuration changes.

Bug Fixes:
- Prevent no-op privacy settings saves from generating misleading audit entries.
- Harden PayPal endpoint handling so unknown or malformed values are rejected rather than silently mapped to the live endpoint.
- Safely serialize administrator-provided consent configuration to prevent script injection.

Enhancements:
- Record provider, category, CMP, legal URL, and third-party service lifecycle changes using the existing LogEntry audit infrastructure.
- Require valid consent categories and preserve third-party service identifiers after creation.

Tests:
- Add coverage for privacy consent configuration, public rendering behavior, service administration, audit logging, permissions, and PayPal endpoint validation.

Chores:
- Add privacy models, settings, migrations, templates, navigation, routes, and frontend assets.

## Summary by CodeRabbit

- **New Features**
  - Added configurable privacy consent management with Klaro or external CMP support.
  - Added privacy settings for consent categories, legal policy links, and third-party services.
  - Added privacy-aware embed placeholders that load content only after consent.
  - Added admin tools to create, edit, disable, and remove services.
  - Added privacy change audit logs for administrators.

- **Bug Fixes**
  - Improved PayPal endpoint validation to reject unknown or malformed values instead of defaulting unexpectedly.


Rachit7168 (changes requested):
If this PR depends on other PR or is a stacked PR. Then please convert this to draft

PR #5627, opened 2026-09-14
fix(vouchers): add explicit All products option and drop Optional hints

### LINKED ISSUE #5622: Remove "Optional" labels and add "All products" option to Product dropdown
## Description

In the **Voucher details** form under Tickets, the **Valid until** and **Product** field labels currently display an "Optional" sub-label. Additionally, the **Product** selection dropdown requires explicit option support for **"All products"** to apply a voucher across every available product item.

## Current Behavior

* The field label for **Valid until** includes the text `Optional`.
* The field label for **Product** includes the text `Optional`.
* The **Product** select dropdown list does not include an explicit **"All products"** selectable item.

## Expected Behavior

* Remove the `Optional` label text from under **Valid until**.
* Remove the `Optional` label text from under **Product**.
* Add **"All products"** as an explicit selectable option in the **Product** dropdown menu (selecting this applies the voucher globally across all products).

## Acceptance Criteria

* [ ] Remove `Optional` sub-text from the **Valid until** field label.
* [ ] Remove `Optional` sub-text from the **Product** field label.
* [ ] Add **"All products"** option to the Product dropdown list.
* [ ] Ensure selecting **"All products"** correctly sets the underlying payload/state to target all available event products.

## Screenshot Reference

Image

sanketkokate712:
Fixes #5622

## Summary

- The **Valid until** and **Product** labels in the voucher form no longer show the "Optional" hint.
- The **Product** dropdown now has **All products** as a real option. It is the first entry in the list and is selected by default for new vouchers and for vouchers that aren't tied to a product.
- Choosing **All products** saves the voucher with no product, variation, or quota. That means it applies to every product in the event, which is how an empty selection already worked. Existing vouchers don't need a migration.

## Changes

- `control/forms/renderers.py`: `ControlFieldRenderer` skips the "Optional" hint for any form field with `hide_optional = True`. Other forms are not affected.
- `control/forms/vouchers.py`: adds an `ALL_PRODUCTS` value (`'all'`). The form now shows it as the initial choice and treats it like an empty selection when cleaning. It also sets `hide_optional` on `valid_until` and `productvar`. `VoucherBulkForm` inherits all of this.
- `control/views/typeahead.py`: `productvarquota_select2` returns **All products** first on page 1 when the search is empty or matches the label.
- `tests/tickets/control/test_voucher_product_all.py`: new tests for the labels, the dropdown results, and saving with **All products**.

## Screenshots

**Voucher details: no "Optional" under Valid until / Product, and All products is preselected**

Voucher details form

**Product dropdown with the explicit All products option**

Product dropdown

**A voucher saved with All products (no product, variation, or quota in the DB) keeps the selection when reopened**

Saved voucher

## Testing

- `pytest tests/tickets/control/test_voucher_product_all.py tests/tickets/control/test_typeahead.py`: 7 passed
- Checked by hand on the local dev server: created a voucher with **All products**. It was saved with `product_id`, `variation_id` and `quota_id` set to `None`.

## Summary by Sourcery

Make voucher-wide product selection explicit and remove unnecessary optional hints from the voucher form.

New Features:
- Add an explicit “All products” voucher option that is selected by default and applies vouchers across every event product.

Bug Fixes:
- Ensure selecting “All products” clears product, variation, and quota associations when saving vouchers.

Enhancements:
- Allow form fields to suppress the “Optional” label hint, hiding it for voucher validity and product fields while preserving it elsewhere.

Tests:
- Add coverage for voucher labels, product typeahead results, and saving vouchers with the “All products” selection.

## Summary by CodeRabbit

* **New Features**
  * Vouchers can apply to all products without selecting a specific product, variation, or quota.
  * “All products” is available as a voucher product option and appears in relevant product searches.

* **Bug Fixes**
  * Saving an all-products voucher correctly leaves product, variation, and quota unset.
  * Voucher forms no longer display “Optional” for fields that are not applicable, while retaining it where appropriate.
  * Product searches now show “All products” only for relevant queries.


app/eventyay/control/views/typeahead.py
```diff
@@ -481,6 +482,10 @@ def productvarquota_select2(request, **kwargs):
 
     choices = []
 
+    all_products_label = str(_('All products'))
+    if page == 1 and query.lower() in all_products_label.lower():
```
Rachit7168: `query.lower() in all_products_label.lower()` is a bit too loose as a substring check.

Empty query is fine (`'' in 'all products'` is true), but something like `duct` or `pro` also matches and prepends **All products** to unrelated searches.

Can we tighten this? e.g. only add it when the query is empty, or when the label actually starts with the query (or equals a known token), instead of any substring hit.

app/eventyay/control/forms/vouchers.py
```diff
@@ -103,7 +108,9 @@ def __init__(self, *args, **kwargs):
         choices = []
```
Rachit7168, line 108: If we hit `Product.DoesNotExist` above we `pass` without setting `productvar`, so edit forms for a voucher whose product was deleted won't get the All products initial value.

Probably rare, but should that `except` also fall back to `ALL_PRODUCTS`?

Rachit7168 (commented):
Nice approach overall — explicit All products + hiding the Optional hints reads much clearer. One note on how the option is matched in typeahead.

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/54395ac5-ad87-4138-8e49-15d42af48c7f

Saksham-Sirohi (approved):

PR #5628, opened 2026-09-14
fix: Send test email from admin settings without page reload or tab change

### LINKED ISSUE #5619: Send test email from admin settings without page reload or tab change
In the admin email settings, when an admin uses **Test email connection** and clicks **Send test email**, the test email is sent successfully, but the page reloads and redirects the user to the **Meta Data** tab.

This is confusing because the admin loses their position in the **Email** settings area.

The test email action should run inline without refreshing the page, and the result message should appear directly inside the email settings section.

Image

## Current behavior

On:

```text
Admin → Global settings → Settings → Email
```

When the admin enters a recipient address and clicks:

```text
Send test email
```

the system:

* sends the test email
* reloads the page
* moves the user away from the **Email** tab
* opens or redirects to the **Meta Data** tab
* forces the admin to navigate back to the email settings area

## Expected behavior

Clicking **Send test email** should not reload the page and should not change the active tab.

The admin should remain on:

```text
Global settings → Settings → Email
```

After the email is sent, show the result message inline in the same section:

```text
Test email sent to <recipient-email> — check inbox.
```

Example:

```text
Test email sent to mariobehling@gmail.com — check inbox.
```

## Required behavior

* Send the test email asynchronously.
* Keep the current tab active.
* Do not reload the full page.
* Do not redirect to **Meta Data** or any other tab.
* Show a loading state while the test email is being sent.
* Show success or error feedback inline below the test email field.
* Preserve all current email settings form values on the page.
* Do not trigger saving of the whole Global Settings form unless explicitly required.
* Validate the recipient email field before sending.

### Loading

```text
Sending test email…
```

### Success

```text
Test email sent to <recipient-email> — check inbox.
```

### Error

```text
Test email could not be sent. Please check the email settings and try again.
```

### Invalid email

```text
Enter a valid recipient email address.
```

## Required changes

* Replace the current full-page form submission behavior with an AJAX/asynchronous request.
* Add or update the backend endpoint for sending test emails so it returns a structured JSON response.
* Keep the user on the active **Email** tab after the request completes.
* Render success and error messages inline in the **Test email connection** section.
* Preserve the recipient email field value after sending.
* Ensure the action works for all supported email configuration modes, including SMTP and Gmail / Google Workspace API.
* Add tests for success, failure, invalid email, and active-tab preservation.

## Acceptance criteria

* Clicking **Send test email** does not reload the page.
* The active tab remains **Email**.
* The user is not redirected to **Meta Data**.
* A success message appears inline after a successful test email.
* Error messages appear inline if sending fails.
* Invalid recipient email addresses are validated before sending.
* Current email settings form values are not lost.
* The test email action works for SMTP configuration.
* The test email action works for Gmail / Google Workspace API configuration, if enabled.
* Tests verify that the test email action does not change the active tab.

arc12at:
## Closes #5619
 
This pr fixes the admin email settings, where an admin uses Test email connection and clicks Send test email, the test email is sent successfully, but the page reloads and redirects the user to the Meta Data tab.

## Videos

https://github.com/user-attachments/assets/8532a311-b885-492c-b7b7-a73c538af190

## Summary by Sourcery

Send administrator test emails inline while preserving the email settings tab and providing clear validation and delivery feedback.

Bug Fixes:
- Allow administrators to send test emails asynchronously without reloading the settings page or switching tabs.
- Keep non-AJAX test-email submissions on the email settings tab instead of redirecting to the metadata tab.
- Return structured JSON feedback for AJAX test-email requests, including validation and configuration errors.

Enhancements:
- Extract test-email and Turnstile interactions into dedicated client-side scripts with recipient validation, loading states, and inline feedback.

Tests:
- Add coverage for successful, invalid, and failing AJAX email tests, multiple recipients, Gmail delivery, redirect behavior, permissions, and active-tab preservation.

## New Features
- Added a test-email workflow with recipient validation, loading feedback, and success or error messages.
- Added Turnstile testing using the currently entered secret key.
- Added JSON responses for asynchronous email tests while preserving standard form behavior.

## Bug Fixes
- Improved navigation back to the email settings section after testing.
- Improved handling of invalid recipients, missing configuration, and email delivery errors.
- Prevented duplicate email-test submissions while a request is in progress.
- Improved the test-email field layout.


app/eventyay/control/views/global_settings.py
```diff
@@ -222,10 +222,15 @@ class GlobalSettingsTestEmailView(AdministratorPermissionRequiredMixin, View):
...
 
-    def _respond(self, request, level, message):
-        """Redirect back to the email tab with inline feedback. Does not save settings."""
+    def _respond(self, request: HttpRequest, level: str, message: str) -> HttpResponse:
+        """Return JSON for AJAX callers; redirect with session feedback for plain form POSTs."""
+        if (
+            request.headers.get('Accept', '').startswith('application/json')
+            or request.headers.get('X-Requested-With') == 'XMLHttpRequest'
+        ):
+            return JsonResponse({'status': level, 'message': str(message)})
```
Rachit7168, line 233: Nice — this is the right place to keep AJAX on the email tab.

One leftover though: the `gmail_api` branch still does `messages.error(...)` + plain `redirect(...)` when no account is connected, so it never hits `_respond`. With the new JS that means `fetch` follows the redirect, `response.json()` blows up, and the user just gets the generic error (and no `#tab-email`).

Can that path go through `_respond` too?

Rachit7168 (commented):
Tab hash fix looks correct (`#tab-email` matches `tab-{{ group_key }}`). Left notes on the JSON path and one leftover redirect that still skips it.

app/eventyay/static/control/js/admin-test-email.js
```diff
@@ -0,0 +1,119 @@
...
+                'X-CSRFToken': getCsrfToken(form),
+                'Content-Type': 'application/x-www-form-urlencoded',
+                'Accept': 'application/json',
+                'X-Requested-With': 'XMLHttpRequest',
+            },
+            body,
+        });
+
+        if (!response.ok) {
+            throw new Error(`HTTP ${response.status}`);
```
Rachit7168, line 74: If the server returns a non-2xx JSON body (or HTML from that gmail redirect case), we throw here and only show `data-error-text` — never the actual `message`.

Maybe try parsing JSON on failure too, and only fall back to the generic text when that isn't available?

app/eventyay/static/control/js/admin-test-email.js
```diff
@@ -0,0 +1,119 @@
...
+    const form = document.getElementById('admin-test-email-form');
+    const input = document.getElementById('admin-test-email-input');
+    const feedbackEl = document.getElementById('admin-test-email-feedback');
+
+    if (!form || !input || !feedbackEl) {
+        return;
+    }
+
+    const triggerSend = (e) => {
+        e.preventDefault();
```
Rachit7168, line 102: Worth a quick `if (btn.disabled) return;` at the start of `sendTestEmail` / `triggerSend`.

A fast double Enter can still fire two requests before the first call sets `disabled`.

Rachit7168 (approved):
LGTM, Tested 

https://github.com/user-attachments/assets/d0ca89ec-97d7-4be4-bf4d-a18392864245

ArnavBallinCode (approved):

PR #5631, opened 2026-09-14
fix: password input field visibility toggle does not unmask password characters

### LINKED ISSUE #5623: Password input field visibility toggle does not unmask password characters
## Description

When toggling the eye icon visibility button on the **Password** input field, the icon toggles between open and slashed states, but the input content remains masked (`*****`) instead of revealing the plain text password.

## Current Behavior

* Clicking the eye icon changes the toggle icon visual state, but the field input continues to display obscured characters (`*****`) regardless of toggle state.

## Expected Behavior

* Clicking the eye icon to the "visible" state should dynamically switch the input `type` from `password` to `text`, revealing the plain text password characters.
* Clicking the eye icon to the "hidden" state should revert the input `type` back to `password`, masking the characters (`*****`).

## Steps to Reproduce

1. Enter a password into the optional **Password** input field.
2. Click the eye icon on the right end of the input box.
3. Observe that the password remains masked (`*****`).

## Acceptance Criteria

* [ ] Bind password input field `type` dynamically to the visibility toggle state (`type="text"` when visible, `type="password"` when hidden).
* [ ] Ensure toggling properly reveals and hides the plain text content.

## Screencast Reference

https://github.com/user-attachments/assets/32741c98-370a-4aff-95e2-2c8514f652fa

jagritixjha:
Resolves  #5623 
---

**What caused the problem**
---
Previously, when the "eye" icon was clicked on a secret key settings field (e.g., SMTP password), the javascript simply toggled the input field `type` between `password` and `text`. However, because the server intentionally redacts sensitive passwords in the initial HTML (rendering them as `*****`), toggling visibility merely exposed the redacted asterisks, not the actual password. Sending the plaintext password in the initial HTML load is a major security risk, so the previous simple toggle approach was fundamentally flawed.

**What has been changed**
---
1. **Backend Step-up Authentication**: Added a new endpoint (`RevealSecretSettingView`) under `/admin/global/settings/reveal-secret/`. This endpoint requires an active staff session and re-authenticates the user's account password via `check_password()` before returning the requested plaintext secret. It only allows revealing whitelisted keys (e.g., `smtp_password`).
2. **Widget Data Binding**: Updated `SecretKeySettingsWidget` to embed the setting key (e.g. `smtp_password`) into `data-setting-key` so the frontend knows what to fetch.
3. **Frontend Rewrite (Reveal-on-demand)**: Overhauled `secret-toggle.js` and `_payment-settings.scss`. When the eye icon is clicked on a redacted field:
   - A step-up authentication modal prompts the user to confirm their account password.
   - Upon successful verification, the true plaintext secret is fetched from the server and displayed in the input field.
   - A live 30-second countdown badge is displayed. After 30 seconds (or if the user clicks "Hide"), the field reverts to its masked state (`*****`) and the plaintext is purged from the DOM.
4. **Modal Styling**: The new password confirmation modal was styled to match the native `pretix-confirm-dialog` UI (blur backdrop, correct typography, matching SCSS variables) and features a loading spinner during API verification. The eye icon is positioned perfectly in the rightmost corner of the input field using robust absolute positioning.

**Testing**
---
Added `app/tests/tickets/control/test_global_settings_reveal.py`:
- `test_reveal_secret_success`: supplying the correct admin password successfully returns the unmasked `secret_value`.
- `test_reveal_secret_invalid_password`: entering the wrong password correctly yields a `403` with an `invalid_password` error.
- `test_reveal_secret_key_not_allowed`: requesting a non-whitelisted key (e.g., `some_other_setting`) returns a `403 forbidden`.
- `test_reveal_secret_empty_value`: attempting to reveal a key that has no stored value gracefully returns a `404 not_set`.
- `test_reveal_secret_requires_staff`: attempting to hit the endpoint as a normal (non-staff) user returns a `403` or `302`.

All five pass. 
All of them fail on the `development` branch because the endpoint and step-up logic did not exist. 

## Manual frontend testing

Confirmed that:

* The eye icon remains correctly positioned inside the input field.
* Newly entered passwords use the simple visibility toggle.
* The 30-second countdown automatically masks the secret.
* Incorrect passwords display inline errors without closing the modal.

https://github.com/user-attachments/assets/f5bcabeb-b5c4-4517-aff5-02ccac1a64d4

## Summary by Sourcery

Secure secret setting visibility toggles by requiring step-up authentication before revealing stored credentials.

New Features:
- Add on-demand secret setting reveal with password re-authentication for approved credentials.
- Automatically remask revealed secrets after 30 seconds or when manually hidden.

Bug Fixes:
- Fix secret-field visibility toggles so redacted saved credentials no longer expose only placeholder characters.

Enhancements:
- Provide a consistent confirmation modal, countdown indicator, and improved secret-field toggle styling.

Tests:
- Add coverage for successful reveals, invalid passwords, unauthorized users and keys, and unset secrets.

## Summary by CodeRabbit

- **New Features**
  - Added secure, on-demand revealing of saved administrative secrets after password confirmation.
  - Revealed secrets automatically hide again after 30 seconds and can be hidden immediately.
  - Added countdown feedback, clear prompts, error handling, and accessible controls for secret visibility.

- **Bug Fixes**
  - Improved password and secret toggles to target the correct visible input fields.
  - Prevented hidden inputs and saved placeholder passwords from being revealed incorrectly.
  - Password fields now switch reliably between hidden and visible modes.


app/eventyay/control/views/global_settings.py
```diff
@@ -769,3 +769,89 @@ def post(self, request, *args, **kwargs):
...
+                'Secret reveal re-auth failed for user %s (key=%s)',
+                request.user.pk,
+                key,
+            )
+            return JsonResponse(
+                {'error': 'invalid_password', 'detail': str(_('The password you entered was invalid.'))},
+                status=403,
+            )
+
+        gs = GlobalSettingsObject()
```
Rachit7168: This endpoint hardcodes `GlobalSettingsObject()` and requires `is_staff`. However, the `SecretKeySettingsWidget` is also used for Organizer settings (e.g., `smtp_password` in `app/eventyay/orga/forms/email.py`). 

Because `data-reveal-secret-url` is hardcoded globally on the `<body>` tag in `base.html`:
1. If a non-staff Organizer clicks "Show secret key" on their own SMTP settings, they will get an unhandled 403 error.
2. If a staff member views an Organizer's SMTP settings and clicks "Show", the endpoint will blindly return the **Global** `smtp_password` instead of the Organizer's password. The JavaScript will then populate the organizer's input field with the global password!

To fix this, the reveal endpoint needs to be context-aware. It should accept parameters indicating the scope (e.g., `scope=global`, `scope=organizer`, `organizer_slug=...`) and query the correct `SettingsObject`. Permissions must also be checked dynamically based on the requested scope.

app/eventyay/static/pretixcontrol/js/ui/secret-toggle.js
```diff
@@ -1,49 +1,322 @@
...
+            maskField(input, btn);
         } else {
             input.type = 'password';
-            icon.classList.remove('fa-eye-slash');
-            icon.classList.add('fa-eye');
-            button.setAttribute('aria-pressed', 'false');
-            button.setAttribute('aria-label', showLabel);
+            input.setAttribute('type', 'password');
+            const icon = btn.querySelector('i');
+            if (icon) { icon.classList.remove('fa-eye-slash'); icon.classList.add('fa-eye'); }
```
Rachit7168: **ReferenceError: `button` and `showLabel` are undefined.**

This code will crash when a user clicks the toggle to hide a locally-typed password. The parameter is named `btn`, not `button`, and `showLabel` is no longer in scope since it was refactored. 

You can remove this entire `else` block and just call `maskField(input, btn);` instead, as `maskField` is already designed to properly mask both server-revealed and locally-typed values.

app/eventyay/base/forms/__init__.py
```diff
@@ -243,17 +243,18 @@ def render(self, name, value, attrs=None, renderer=None):
...
             f'aria-label="{show_label}" '
             f'data-label-show="{show_label}" '
             f'data-label-hide="{hide_label}" '
+            f'data-key="{name}" '
             'aria-pressed="false">'
             '<i class="fa fa-eye"></i>'
             '</button>'
             '</div>'
-        )
+        ).format(key=name)
```
Rachit7168: **Code Quality / Crash Risk:**
Concatenating an f-string (`f'{output}'`) with a normal string and then calling `.format()` on the result is highly discouraged. 

If the `output` string happens to contain `{` or `}` anywhere inside it (which is very possible depending on field values or attributes), calling `.format(key=name)` will try to parse those braces as kwargs, raise a `KeyError`, and crash the form render with a 500 error. 

Instead of mixing them, please remove `.format(key=name)` entirely and just use f-strings for the whole block:
```python
        toggle_html = (
            f'<div class="secret-key-wrapper" data-setting-key="{name}">'
            f'{output}'
            # ...

**Review by Rachit7168 — CHANGES_REQUESTED** [review]
Please address following comments and could we do a quick pass on the UI to match our design standards? Swapping the heavy black border for a light grey border and a soft drop shadow will make the modal look much cleaner. Also, aligning the lock icon inline with the title, adding consistent inner padding, and using a neutral dark grey for the descriptive text will really help polish the overall layout and contrast. The UI should feel smoother and clean like in recent UI change in Video

File: app/eventyay/control/templates/pretixcontrol/base.html [hunk 4]
```diff
@@ -466,5 +466,98 @@ <h4 class="pretix-confirm-dialog__title" id="pretix-confirm-dialog-title"></h4>
...
+                max-width: 90vw !important;
+                background: #ffffff !important;
+                overflow: hidden !important;
+                outline: none !important;
+            }
+            .secret-reveal-modal::backdrop {
+                background: rgba(0, 0, 0, 0.4) !important;
+                backdrop-filter: blur(2px) !important;
+            }
+            .secret-reveal-modal__header {
```
Rachit7168, line 486: Avoid using important and inline CSS or JS

app/eventyay/control/views/global_settings.py
```diff
@@ -769,3 +769,108 @@ def post(self, request, *args, **kwargs):
...
+
+        if scope == 'global':
+            if not (request.user.is_staff or request.user.is_superuser):
+                return JsonResponse({'error': 'forbidden', 'detail': 'Administrator access required.'}, status=403)
+            gs = GlobalSettingsObject()
+        elif scope == 'organizer':
+            organizer_slug = request.POST.get('organizer', '')
+            if not organizer_slug:
+                return JsonResponse({'error': 'forbidden', 'detail': 'Organizer slug required.'}, status=403)
+            from eventyay.base.models import Organizer
```
Rachit7168, line 835: Move imports to top

app/eventyay/static/pretixcontrol/js/ui/secret-toggle.js
```diff
@@ -1,49 +1,321 @@
 /**
- * Secret key toggle functionality using vanilla JavaScript
- * Handles password visibility toggle for secret key input fields
+ * Reveal-on-demand for SecretKeySettingsWidget fields.
+ *
+ * Flow:
+ * 1. User clicks the "Show" button (.secret-reveal-btn) on a masked field.
+ * 2. If the field shows '*****' (a saved-and-redacted secret), a password modal
```
Rachit7168, line 6: No need for this much comments, keep it short and clear

app/eventyay/static/pretixcontrol/scss/_payment-settings.scss
```diff
@@ -28,16 +28,18 @@
...
 .secret-key-wrapper {
   position: relative;
-  display: inline-block;
+  display: block;
   width: 100%;
 
   input {
-    padding-right: 35px;
+    // Extra right padding so the button never overlaps text
+    padding-right: 80px !important;
```
Rachit7168, line 39: Same here : Avoid important

Rachit7168 (changes requested):
Thanks for working on the issue. Please address :

jagritixjha:
@Aqil-Ahmad @Rachit7168 ,
All outstanding reviews are addressed ptal

Rachit7168 (changes requested):
image

This is a loading animation in the " Show password " button that keeps loading no matter the state. Please remove that!

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/97dd605b-556a-49aa-86a3-caaf9954bc58

Saksham-Sirohi (approved):
Step-up reveal is allowlisted, re-checks the account password, and returns secrets only after permission checks with Cache-Control: no-store. The toggle is an external ES module, not inline script.

PR #5705, opened 2026-09-18
fix(admin): show PayPal Connect settings registered by the plugin

### LINKED ISSUE #5704: PayPal Connect settings registered by the plugin are not shown in admin ticketing settings
### Problem

The admin page **Global settings → Ticketing** shows only three PayPal fields: Client ID, Secret key and Endpoint. Any other global setting that the PayPal plugin registers through `register_global_settings` is silently dropped, even though the form accepts and saves it.

`pretixcontrol/admin/ticketing_settings.html` renders the three fields it knows by name and the generic loop below then skips everything that starts with `payment_paypal_connect_`:

```
{% bootstrap_field form.payment_paypal_connect_client_id layout='control' %}
{% bootstrap_field form.payment_paypal_connect_secret_key layout='control' %}
{% bootstrap_field form.payment_paypal_connect_endpoint layout='control' %}
...
{% if not "payment_stripe_connect_" in field_name and not "payment_paypal_connect_" in field_name %}
```

### Impact

`eventyay-paypal` registers `payment_paypal_connect_partner_payer_id`, the platform merchant ID that is required to ask PayPal about a seller onboarding (`GET /v1/customer/partners/{partner_payer_id}/merchant-integrations`). Because the field is never rendered, an administrator has no way to fill it in and the event settings page reports:

> Eventyay cannot ask PayPal about your onboarding because the platform payer ID is missing. Please ask your administrator to add it to the global PayPal Connect settings.

So a seller whose PayPal onboarding did not redirect back to Eventyay cannot be connected at all.

### Expected behaviour

PayPal Connect settings contributed by the plugin appear in the PayPal section of the ticketing settings page, like every other plugin setting does.

### Steps to reproduce

1. Install `eventyay-paypal`.
2. Open Global settings → Ticketing as a system administrator.
3. The PayPal section shows only Client ID, Secret key and Endpoint; no platform payer ID field.

Rachit7168:
Fixes #5704

### Problem

The PayPal section of Global settings → Ticketing rendered only the three fields the template knows by name, and the generic loop underneath skipped every remaining `payment_paypal_connect_*` field. Settings that `eventyay-paypal` registers through `register_global_settings` were added to the form and saved correctly, but never displayed, so an administrator could not enter them.

The concrete case is `payment_paypal_connect_partner_payer_id`, the platform merchant ID the plugin needs to ask PayPal about a seller onboarding. Without it, a seller whose PayPal onboarding did not redirect back to Eventyay cannot be connected.

### Change

- `GlobalTicketingSettingsForm` now collects the PayPal Connect fields that the template does not render by name into `paypal_extra_fields`.
- `ticketing_settings.html` renders those fields at the end of the PayPal section, so plugin-provided PayPal settings show up without hard-coding plugin field names in core.
- Added a test that a PayPal Connect setting registered through the signal is rendered.

### Verification

- Added `test_ticketing_page_renders_paypal_settings_added_by_plugins`, which registers a PayPal Connect setting through the signal and asserts it is rendered. It fails without this change; CI runs it.
- Template compiles with the Django engine; no other page uses `paypal_extra_fields`.

## Summary by Sourcery

Render plugin-registered PayPal Connect settings in the global ticketing configuration.

Bug Fixes:
- Display PayPal Connect settings registered by plugins in the global ticketing settings form so administrators can configure them.

Enhancements:
- Keep plugin-provided PayPal Connect fields discoverable without hard-coding their names in the core template.

Tests:
- Add coverage verifying that a PayPal Connect setting registered through the global settings signal is rendered.

## Summary by CodeRabbit

- **New Features**
  - PayPal Connect settings can now display additional payment fields provided by installed plugins.
  - Plugin-provided fields appear automatically in the ticketing settings page alongside the standard PayPal options.

- **Bug Fixes**
  - Improved compatibility with plugins that add custom PayPal Connect configuration fields.


Saksham-Sirohi (approved):
`paypal_extra_fields` picks up plugin-registered `payment_paypal_connect_*` settings that the named template fields do not cover, and the existing payment-gateways loop still skips that prefix, so there is no double render. That surfaces `payment_paypal_connect_partner_payer_id` for the PayPal onboarding status lookup.

ArnavBallinCode (approved):

PR #5724, opened 2026-09-19
fix: show orphaned shifts from deleted rooms in schedule editor

### LINKED ISSUE #137: [RFC] Bridge rooms between Talks schedule and TeamShifts locations

MukundC25:
Closes https://github.com/fossasia/eventyay-teamshifts/issues/137 | Paired with https://github.com/fossasia/eventyay-teamshifts/pull/163

**Description:**

Fixes the shift schedule editor hiding shifts that belong to deleted rooms. Previously, `deletedRoomSessions` returned an empty array in shifts mode, making orphaned shifts invisible — they didn't appear in the grid or the unassigned panel.

**What changed**
- Removed the early `if (isShifts) return []` from `deletedRoomSessions` in `App.vue`
- Orphaned shifts now appear in the "Deleted Room Shifts" recovery panel so organizers can drag them into another room
- Talks mode behavior is identical (no functional change for talks)

https://github.com/user-attachments/assets/8d23a4e0-d4f6-45a1-bd8f-9fa164d2f9e9

## Summary by Sourcery

Preserve access to TeamShifts assignments by surfacing shifts from unavailable rooms and preventing their linked rooms from being removed prematurely.

New Features:
- Show shifts associated with deleted or unscheduled rooms in the schedule editor’s recovery panel for reassignment.

Bug Fixes:
- Prevent rooms with assigned TeamShifts shifts from being marked unscheduled or deleted, avoiding unavailable orphaned shifts.

Enhancements:
- Add room-list and deletion-page warnings identifying rooms with linked TeamShifts shifts and explaining the required reassignment or deletion steps.

## Summary by CodeRabbit

- **Bug Fixes**
  - Deleted-room break sessions now appear in the unassigned panel for shift-based schedules.
  - Shift-session details now display assigned roles correctly, without irrelevant speaker or track information.
  - Rooms associated with existing shifts can no longer be deleted or marked as unscheduled.
  - Updated messaging clarifies that listed shifts come from rooms that were deleted or unscheduled.

- **User Interface**
  - Room lists and deletion confirmations now warn when assigned shifts are linked to a room, including the number of affected shifts.


app/eventyay/base/services/room.py
```diff
@@ -302,6 +301,16 @@ def soft_delete_room(event, room, by_user=None):
         # submission after validation and before deleted=True is committed.
         room = Room.objects.select_for_update().get(pk=room.pk)
         validate_room_can_be_deleted(room)
+
+        shift_location = getattr(room, 'shift_location', None)
+        if shift_location is not None and shift_location.shifts.exists():
```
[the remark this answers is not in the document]
Rachit7168: @MukundC25 , Please take a look

Rachit7168 (changes requested):
Please resolve pending review and comment left by sourcery

app/eventyay/base/services/room.py
```diff
@@ -302,6 +301,19 @@ def soft_delete_room(event, room, by_user=None):
...
+        except ObjectDoesNotExist:
+            shift_location = None
+        if shift_location is not None and shift_location.shifts.exists():
+            raise ValidationError(
+                _(
+                    'This room is used by shifts in the TeamShifts schedule. '
+                    'Reassign or delete those shifts before deleting this room.'
+                )
+            )
+
```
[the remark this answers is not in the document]
MukundC25: outdated

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/372b2001-422e-4d53-b376-6fde6e42ab7b

app/eventyay/webapp/schedule-editor/src/App.vue
```diff
@@ -560,13 +560,14 @@ const unscheduled = computed<SessionData[]>(() => {
...
   const isShifts = mode === 'shifts' || mode === 'public-shifts'
-  if (isShifts) return []
   return schedule.value.talks
     .filter(
       (session) =>
         session.code &&
         session.start &&
-        (!session.room || !roomsLookup.value[lookupKey(session.room)]),
+        (isShifts
+          ? (session.room && !roomsLookup.value[lookupKey(session.room)])
```
Aqil-Ahmad, line 569: **Dragging an orphaned shift back onto the sidebar now permanently deletes it.**

Making `deletedRoomSessions` non-empty in shifts mode newly exposes these sessions to `stopDragging()`:

```js
if (draggedSession.value.code && !draggedSession.value.deletedRoom) {
  // …unassign (room = undefined) + saveTalk
} else if (schedule.value?.talks.find(...)) {
  schedule.value.talks = schedule.value.talks.filter(...)
  await api.deleteTalk({ id: Number(draggedSession.value.id) })
}
```

Every session produced here carries `deletedRoom: true`, so the first branch is skipped and the `else if` runs.

Scenario: an organizer picks up a shift from the new "Shifts from Unavailable Rooms" panel, drags it toward the grid (`pointerleave` on `#unassigned`), changes their mind, moves back over the sidebar (`pointerenter` sets `isUnassigning = true`) and releases. `stopDragging()` calls `api.deleteTalk` and the Shift row is destroyed — no confirmation, no undo. Before this change `deletedRoomSessions` was empty in shifts mode, so this path was unreachable for shifts; the panel's stated purpose is recovery, so silent deletion is the opposite of the intent.

Suggest treating `deletedRoom` shifts like the unassign branch (or simply ignoring an unassign drop for them) rather than falling through to delete.

app/eventyay/base/models/room.py
```diff
@@ -141,6 +145,14 @@ def validate_is_unscheduled_change(room) -> None:
     """Raise ValidationError if the room cannot be marked as unscheduled."""
     if room.pk and room_has_linked_submissions(room):
         raise ValidationError({'is_unscheduled': UNSCHEDULED_LINKED_SUBMISSIONS_MESSAGE})
+    if room.pk:
+        with scope(event=room.event):
+            try:
+                shift_location = room.shift_location
```
Aqil-Ahmad: `room.shift_location` is a reverse accessor contributed by the optional `teamshifts` app. When that app is not in `INSTALLED_APPS`, the attribute does not exist at all and this line raises a plain `AttributeError`, which `except ObjectDoesNotExist` does not catch.

`teamshifts` is genuinely optional: `EVENTYAY_PLUGINS_EXCLUDE` (`config/settings.py:379`, from `conf.plugins_exclude`) removes it from `INSTALLED_APPS`, and the codebase already guards for this elsewhere — `control/forms/organizer_forms/team_form.py:76` uses `apps.is_installed('teamshifts')`. The same happens if the deployed `eventyay-teamshifts` is an older revision without the `linked_room` field added in the paired PR (the plugin is pinned to `branch = "main"`, so core and plugin are not upgraded atomically).

Scenario: on such a deployment, saving any room with "unscheduled" ticked goes `RoomForm` → `Room.clean()` → here → uncaught `AttributeError` → HTTP 500.

`RelatedObjectDoesNotExist` subclasses `AttributeError`, so widening the except to `AttributeError` (or gating the whole block on `apps.is_installed('teamshifts')`) fixes both cases.

app/eventyay/orga/views/schedule.py
```diff
@@ -571,6 +590,19 @@ def get_context_data(self, **kwargs):
             ctx['schedule_room_url'] = schedule_editor_room_url(
                 self.request.event, self.object
             )
+
+            try:
+                shift_location = self.object.shift_location
```
Aqil-Ahmad: Same optional-plugin problem as in `base/models/room.py`: `self.object.shift_location` raises a plain `AttributeError` — not `ObjectDoesNotExist` — when the `teamshifts` app is absent from `INSTALLED_APPS` (possible via `EVENTYAY_PLUGINS_EXCLUDE`) or when the deployed plugin predates the `linked_room` field from the paired PR.

Scenario: on such a deployment, opening the room delete confirmation page (`/orga/event/<slug>/schedule/rooms/<pk>/delete/`) raises `AttributeError` and returns HTTP 500, so rooms can no longer be deleted at all.

Note the inconsistency with the `list` branch a few lines above, which does defend against a missing plugin. Catching `AttributeError` here (it is a superclass of `RelatedObjectDoesNotExist`) or gating on `apps.is_installed('teamshifts')` covers both.

app/eventyay/orga/views/schedule.py
```diff
@@ -540,6 +541,24 @@ def get_generic_title(self, instance=None):
 
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data(**kwargs)
+
+        if self.action == 'list':
+            try:
+                from teamshifts.models import ShiftLocation
```
Aqil-Ahmad, line 548: `except ImportError` does not cover the failure mode it is guarding against. If the `eventyay-teamshifts` distribution is installed (it is a hard dependency in `app/pyproject.toml`) but the app is excluded from `INSTALLED_APPS` via `EVENTYAY_PLUGINS_EXCLUDE`, `from teamshifts.models import ShiftLocation` does not raise `ImportError` — Django raises `RuntimeError: Model class teamshifts.models.ShiftLocation doesn't declare an explicit app_label and isn't in INSTALLED_APPS`.

Scenario: an operator sets `plugins_exclude = ("teamshifts",)`; the room list page then returns HTTP 500 instead of degrading gracefully.

The established pattern in this codebase is `apps.is_installed('teamshifts')` (see `control/forms/organizer_forms/team_form.py:76`). It also avoids wrapping the ORM query itself in the `try`, which currently swallows any `ImportError` raised from deeper inside query execution.

app/eventyay/base/models/room.py
```diff
@@ -141,6 +145,14 @@ def validate_is_unscheduled_change(room) -> None:
     """Raise ValidationError if the room cannot be marked as unscheduled."""
     if room.pk and room_has_linked_submissions(room):
         raise ValidationError({'is_unscheduled': UNSCHEDULED_LINKED_SUBMISSIONS_MESSAGE})
+    if room.pk:
```
Aqil-Ahmad: This check runs on **every** save while `is_unscheduled` is true, not only on the transition to unscheduled — `Room.clean()` calls it whenever `self.is_unscheduled` is truthy, and `RoomForm` includes `is_unscheduled`.

Scenario: a room is marked unscheduled while its linked `ShiftLocation` has no shifts (allowed). An organizer later creates shifts at that location in TeamShifts. From then on, editing anything about the room — renaming it, changing capacity or description — fails validation with "Reassign or delete those shifts before marking this room as unscheduled", even though `is_unscheduled` is unchanged. The only way to save is to untick unscheduled.

Comparing against the stored value (e.g. skip when the DB row is already `is_unscheduled=True`) would restrict this to the actual state change.

app/eventyay/orga/templates/orga/schedule/room/delete.html
```diff
@@ -66,6 +66,33 @@ <h3 class="h5 mb-3">
       </div>
     </form>
   {% else %}
-    {% include "common/includes/action_confirm.html" with action_object_name=generic_title action_back_url=list_url %}
+    {% if has_linked_shifts %}
```
Aqil-Ahmad, line 69: This dialog lets the organizer delete a room that has linked shifts ("Delete anyway"), while `validate_is_unscheduled_change` in `base/models/room.py` hard-blocks marking the *same* room unscheduled with a message telling the user to reassign or delete the shifts first.

Since `_location_is_available()` in the paired teamshifts PR treats `room.deleted` and `room.is_unscheduled` identically, both paths produce exactly the same orphaned-shift state. So the validator does not actually prevent orphaning — an organizer who hits the block can just delete the room instead — it only makes the two routes behave inconsistently for no user-visible reason.

Worth picking one policy: either allow both with this warning, or block both.

app/eventyay/webapp/schedule-editor/src/App.vue
MukundC25: fixed

app/eventyay/orga/views/schedule.py
MukundC25: replaced

app/eventyay/orga/templates/orga/schedule/room/delete.html
MukundC25: Both delete and unschedule are now allowed for rooms with linked shifts. Delete shows a warning on the confirmation page. Unschedule has no hard block.

Aqil-Ahmad (approved):

app/eventyay/base/services/room.py
```diff
@@ -5,9 +5,11 @@
 from channels.db import database_sync_to_async
 from channels.layers import get_channel_layer
 from django.core.cache import cache
+from django.core.exceptions import ValidationError
```
Saksham-Sirohi: ValidationError is unused, and gettext_lazy on line 12 is unused as well. This file's other edits are quote changes in stream and room helpers that this fix does not need. Please remove both imports and revert the formatting-only edits.

app/eventyay/orga/views/schedule.py
```diff
@@ -540,6 +542,23 @@ def get_generic_title(self, instance=None):
 
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data(**kwargs)
+
+        if self.action == 'list':
+            if apps.is_installed('teamshifts'):
+                from teamshifts.models import ShiftLocation
+                if hasattr(ShiftLocation, 'linked_room'):
```
Saksham-Sirohi: hasattr(ShiftLocation, 'linked_room') is true once the model code is imported, including before migration 0023 adds the column. This filter then errors and the Talks room list fails. The delete branch at line 594 has the same problem via the shift_location relation. Catch that database error, or skip the lookup until the column exists. Please also add a test that the delete page shows the shift count when the linked location has shifts.

Saksham-Sirohi (changes requested):
The recovery panel is the right fix for shifts whose room is no longer in the schedule payload, and it belongs with eventyay-teamshifts#163. A few changes are still required.

1. room.py adds ValidationError and gettext and does not use them. It also rewrites quotes through unrelated room helpers. Please drop both imports and leave that file's formatting alone.

2. The room list and delete page query ShiftLocation.linked_room as soon as the model attribute exists. That query errors if teamshifts code is deployed before migration 0023. Guard it so Talks room pages still load, and add a test that a room linked to a location with shifts shows the deletion warning and the shift count.

PR #5751, opened 2026-09-19
fix(schedule): make schedule import version timestamp timezone-aware

### LINKED ISSUE #5749: fix(schedule): exported schedule "version" uses naive server-local timestamp instead of event/UTC timezone
**Describe the bug**
In `app/eventyay/storage/schedule_to_json.py`, `convert()` builds the schedule payload but
sets the top-level `version` field with a **timezone-naive, server-local** timestamp:

    result = {"version": dt.datetime.now().isoformat()}

In the same function, every talk/session timestamp is converted using the event's
`timezone` (`to_iso, timezone=timezone`). So `version` is in the server's local timezone
while all talk times are in the event's timezone — inconsistent whenever they differ, and
the naive ISO string (no offset) can't be parsed reliably.

**Expected behavior**
`version` should be a **timezone-aware** timestamp (UTC via `django.utils.timezone.now()`,
or derived from the event `timezone` passed to `convert()`), consistent with the schedule
data in the same payload.

**Screenshots**
N/A (backend data bug).

**Additional context**
- File: `app/eventyay/storage/schedule_to_json.py`, function `convert()`
- Component: schedule / storage
- Proposed fix: replace `dt.datetime.now()` with `django.utils.timezone.now()` (or the
  event timezone) and add a regression test asserting `version` is timezone-aware.

techvoyager-varun:
Fixes #5749

The schedule import `version` field used a naive, server-local timestamp
(`dt.datetime.now()`) while all talk times use the event timezone. This makes
`version` timezone-aware (event tz, falling back to UTC) and consistent with
the payload. Adds regression tests.

## Summary by Sourcery

Ensure schedule import version timestamps use an event-aware timezone instead of server-local time.

Bug Fixes:
- Make schedule import version timestamps timezone-aware and aligned with the event timezone, falling back to UTC when no timezone is provided.

Enhancements:
- Support timezone names and resolved timezone objects consistently when serializing schedule timestamps.

Tests:
- Add regression coverage for timezone-aware version timestamps, UTC fallback, and event timezone offsets.

## Summary by CodeRabbit

* **Bug Fixes**
  * Schedule export version timestamps now include timezone information.
  * Version timestamps consistently reflect the selected event timezone or UTC when none is provided.

* **Tests**
  * Added coverage to verify timezone-aware timestamps for default and event-specific timezone settings.


techvoyager-varun:
@ArnavBallinCode anything I have to add in this more?

PR #5772, opened 2026-09-20
feat(orga): add "Until the first schedule is published" featured visibility

### LINKED ISSUE #5757: Add “Until the first schedule is published” option for featured sessions
Featured sessions can be useful as a teaser for event content before the full schedule is available.

Currently, **Talk settings → General → Featured → Show featured sessions** supports options such as:

```text
Never
Once the first schedule version is published
Always
```

What is missing is the opposite behavior:

```text
Until the first schedule is published
```

This would allow organisers to showcase selected sessions before the schedule is released and automatically stop showing the Featured sessions section once the first public schedule is published.

Image

## Current behavior

Organisers can configure featured sessions to:

* never appear
* appear once the first schedule version has been published
* always appear

There is currently no option to use featured sessions specifically as a **pre-schedule teaser**.

## Expected behavior

Add a new option:

```text
Until the first schedule is published
```

The full selector should become:

```text
Show featured sessions

Never
Until the first schedule is published
Once the first schedule version is published
Always
```

### Never

Featured sessions are never shown publicly.

### Until the first schedule is published

Featured sessions are shown:

```text
before first public schedule publication → visible
after first public schedule publication  → hidden
```

The transition should happen automatically when the first schedule version becomes publicly available.

### Once the first schedule version is published

Keep the existing behavior:

```text
before schedule publication → hidden
after schedule publication  → visible
```

### Always

Featured sessions are always shown, provided sessions have actually been marked as featured.

## Important distinction

The new option should depend on whether the event has **ever published its first schedule**, not simply whether a schedule is currently visible.

For example, if the first schedule was published and later temporarily unpublished or modified, the platform should not unexpectedly restore the pre-schedule teaser unless this matches the existing schedule publication model.

Use the same definition of **first schedule published** that is already used by:

```text
Once the first schedule version is published
```

so both options are exact opposites.

## Public event behavior

When:

```text
Show featured sessions = Until the first schedule is published
```

the public Featured sessions page/block should be accessible before schedule publication.

Once the first schedule is published:

* remove/hide the Featured sessions teaser from normal event navigation where applicable
* do not show the featured sessions block on the event page
* preserve the underlying `featured` flags on sessions

Do **not** automatically unmark sessions as featured.

If the organiser later changes the setting to **Always**, the same featured sessions should become visible again.

## Admin UI

Update:

```text
Talk settings
→ General
→ Featured
→ Show featured sessions
```

Add:

```text
Until the first schedule is published
```

Place it between:

```text
Never
```

and:

```text
Once the first schedule version is published
```

## Help text

Update the field description to explain the available behavior.

Suggested text:

> Controls when featured sessions are shown publicly. They can be used as a teaser before the schedule is published, shown only after schedule publication, always shown, or disabled completely.

## Required changes

* Add the new **Until the first schedule is published** setting value.
* Extend the backend/configuration model accordingly.
* Update public visibility logic for featured sessions.
* Use the existing first-schedule-publication state.
* Preserve existing featured session selections.
* Update translations.
* Update tests for all four modes.

## Acceptance criteria

* **Show featured sessions** contains:

  * Never
  * Until the first schedule is published
  * Once the first schedule version is published
  * Always
* With **Until the first schedule is published**, featured sessions are visible before the first schedule publication.
* They automatically stop being shown after the first schedule is published.
* Existing **Once the first schedule version is published** behavior continues to work.
* **Always** continues to show featured sessions before and after schedule publication.
* **Never** continues to hide featured sessions.
* Changing visibility does not remove the `featured` state from sessions.
* Existing events keep their currently configured behavior.
* Tests cover all four visibility modes.

arpan7sarkar:
## What

Adds a fourth option to **Talk settings → General → Featured → Show featured sessions**:

```
Never
Until the first schedule is published   <- new
Once the first schedule version is published
Always
```

`until_schedule` shows featured sessions as a pre-schedule teaser and hides them
automatically once the first schedule version is published.

Featured settings with Show featured sessions and Show featured speakers set to Until the first schedule is published, with the updated help text

The control itself is unchanged: same `ChoiceField`, same widget, same template. The only
UI change is the extra option and the reworded help text on both featured fields.

<details>
<summary>All four options and their order (native select forced open to fit one screenshot)</summary>

Selector showing Never, Until the first schedule is published, Once the first schedule version is published, Always

</details>

## Why

Closes #5757. Featured sessions could be hidden *before* publication but never *after*,
so there was no way to use them as a teaser that retires itself.

## How

The new value is the mirror of `after_schedule`: both show featured content while no
schedule version has been published, and only the published state differs. So it resolves
in `_featured_public_visible`, the single gate the public featured page and the nav tab
already share, and no call site needs its own guard.

"First schedule published" reuses `Event.current_schedule`, the same definition
`after_schedule` uses, so the two options stay exact opposites. As the issue notes, that
means unpublishing every schedule restores the teaser; this matches the existing model
rather than introducing a separate "has ever published" tracker.

Two things worth flagging for review:

- The key is `until_schedule`, not `pre_schedule`. `pre_schedule` is a legacy value that
  both normalizers migrate to `after_schedule`, which would have been the exact opposite
  behaviour.
- `SHOW_FEATURED_VISIBILITY_CHOICES` is shared with `show_featured_speakers`, and that
  setting already falls back to the `show_featured` value. So the new value is handled for
  speakers too, rather than silently falling through to the `after_schedule` branch. If you
  would rather the speakers selector kept three options, say so and I will split the tuple.

No migration: `feature_flags` is a JSON field and existing events keep their stored value.
Nothing writes `is_featured`, so switching back to `Always` restores the same sessions.
Translations are left to Weblate, matching how this repo syncs `.po` files.

## Public behaviour

Same event and same viewport, with `Show featured sessions = Until the first schedule is
published`, before and after releasing the first schedule version.

| Before the first schedule | After the first schedule |
|---|---|
| Event page with the Featured nav tab and the featured speakers block visible | Same event page with the Featured nav tab and featured speakers block gone |
| `Featured` tab present, `/featured/` returns 200 | `Featured` tab gone, `/featured/` no longer public |

The three sessions keep `is_featured = True` throughout; only their visibility changes.

## Testing

Added coverage for all four modes across the orga form, the agenda view, the nav
templatetag and featured speakers.

Verified against a local stack, before and after publishing the first schedule
(`page` = featured page, `nav` = nav tab, `spk` = featured speakers):

```
mode             schedule     page   nav    spk
never            unpublished  False  False  False
until_schedule   unpublished  True   True   True
after_schedule   unpublished  True   True   True
always           unpublished  True   True   True

never            published    False  False  False
until_schedule   published    False  False  False
after_schedule   published    True   False  True
always           published    True   True   True
```

Only the two `until_schedule` rows differ from the same table on `dev`.

The six new tests fail on unmodified `dev` and pass here; the round-trip test fails there
with `assert 'never' == 'until_schedule'`, which is the normalizer trap above.

```
tests/stable/test_featured.py tests/stable/test_featured_speakers.py
15 passed

tests/talk/... featured, choices-order and round-trip selection
22 passed, 28 deselected
```

`tests/talk/agenda/views/test_agenda_featured.py` has `django_assert_max_num_queries`
failures in my environment, but they reproduce identically on unmodified `dev` for every
mode including `never` and `always` (12-13 queries against a budget of 8), so they are
pre-existing and this change adds no queries. `ruff check` and `ruff format --check` on the
touched files are byte-identical to `dev`.

## Summary by Sourcery

Add schedule-aware featured visibility that presents featured content as a pre-schedule teaser and hides it after the first schedule is published.

New Features:
- Add an “Until the first schedule is published” visibility option for featured sessions and speakers.
- Automatically hide featured content and its public navigation after the first schedule version is published while preserving featured flags.

Enhancements:
- Centralize featured visibility resolution for the new schedule-dependent mode and retain existing visibility behavior.
- Update featured visibility help text and support round-trip persistence of all four options.

Tests:
- Add coverage for the new visibility mode across featured pages, navigation, speakers, schedule publication, and organization settings.

## Summary by CodeRabbit

- **New Features**
  - Added an “Until first schedule” visibility option for featured sessions and speakers.
  - Featured content can appear as a preview before the first schedule is published, then automatically become hidden afterward.
  - Expanded visibility descriptions to clarify all available modes.

- **Tests**
  - Added coverage for this behavior across featured sessions, speakers, navigation, and event settings.


Rachit7168:
Closing this PR in favor of #5768. This PR contains a syntax error in  due to a duplicate and incomplete  decorator, making it invalid.

Rachit7168:
Closing this PR in favor of #5768. This PR contains a syntax error in `app/tests/talk/agenda/views/test_agenda_featured.py` due to a duplicate and incomplete `@pytest.mark.parametrize` decorator, making it invalid.

PR #5773, opened 2026-09-20
fix: Validation error on product variation save does not indicate which field is invalid

### LINKED ISSUE #5769: Validation error on product variation save does not indicate which field is invalid
### Description
When saving changes to a product's variations (e.g. under Products → Modify product → Variations), if validation fails, the system displays a generic error banner: "We could not save your changes. See below for details." However, no specific field is highlighted or flagged as the source of the error.

**Notes: the test is on Non-Admission product**

Image

### Steps to reproduce:
- Go to a product's Variations tab 
- Edit a variation, only fill out the Variation name and keep the rest of the field empty
- Attempt to save with invalid or incomplete data
- Observe the error banner at the top, but no indication of which field caused it

### Expected behavior:
The invalid field(s) should be visually highlighted (e.g. red border, inline error message) so organizers can immediately identify and correct the problem, consistent with how other forms in the system handle validation errors.

jagritixjha:
### Description
This PR resolves an issue where validation errors on the Product Variations form would fail silently for Non-Admission products, displaying a generic formset error banner without highlighting the problematic fields. It also fixes a styling bug that prevented the Variation Name field from displaying error states (red borders and inline messages).

### Root Cause
1. **Invisible Field Validation:** In `ProductVariationForm`, `admission_validity_mode` is defined as a required field by default. For Non-Admission products, this field (along with other admission-specific fields) is intentionally excluded from the DOM via template logic (`include_variations.html`). When the form is submitted, the absence of these fields triggers a validation error. Because the field is hidden, the error cannot be displayed inline, leaving the user with an ambiguous generic form failure.
2. **Missing Error State CSS:** The `value` field (Variation name) in the template explicitly declared `form_group_class=""`. This suppressed the `form-group` wrapper required by `django-bootstrap3` to append the `.has-error` class, thereby preventing visual highlights when the variation name was invalid.

### Changes Proposed
- **`app/eventyay/control/forms/product.py`**: Dynamically set `required=False` for all `ADMISSION_VALIDITY_FIELD_NAMES` during `ProductVariationForm` initialization. This ensures that missing admission fields do not trigger validation blockers for Non-Admission products. The underlying `clean_admission_validity_data` handler already falls back to `ADMISSION_VALIDITY_MODE_NONE` when passed empty values, preserving data integrity.
- **`app/eventyay/control/templates/pretixcontrol/item/include_variations.html`**: Removed the `form_group_class=""` override on the variation `value` field. This restores the standard Bootstrap 3 wrapper, allowing `.has-error` to be successfully applied to the field when validation fails.

### Steps to Reproduce / Test
1. Create a **Non-Admission** product, ensuring the "The product should exist in multiple variations" setting is enabled.
2. Navigate to the **Variations** tab and create a variation with valid data. Verify that the form now saves successfully.
3. Edit the variation, intentionally clear the "Variation Name" field to make it empty, and submit.
4. Verify that the Variation Name field is now correctly highlighted with a red border and an inline validation message.

https://github.com/user-attachments/assets/2e3d5c42-715d-40d2-b3af-62730e9b347c

### Related Issues
Fixes #5769

## Summary by Sourcery

Fix product variation validation so non-admission variations save correctly and invalid variation names display clear field-level errors.

Bug Fixes:
- Allow non-admission product variations to save without validation failures from hidden admission fields.
- Restore inline error styling for invalid variation names.
- Preserve admission variation values through hidden form fields when admission options are not displayed.

Chores:
- Update the ProductVariation admission validity field to permit blank values and add the corresponding database migrations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Admission validity fields are no longer incorrectly required when editing product variations.
  * Variation value fields now behave consistently in existing and newly added product variations.
  * Product variations retain their admission validity settings when admission validity is disabled.
  * New variations correctly use the inherited admission validity setting by default.
  * Editing or adding variations now preserves admission validity information more reliably across form submissions.
  * Blank admission validity settings are handled correctly when saved through the product variation form.


Aqil-Ahmad (changes requested):
please fix this:

The hidden input silently rewrites a valid mode

Image

Rachit7168 (changes requested):
There is no need for 2 migration file

app/eventyay/control/templates/pretixcontrol/item/include_variations.html
```diff
@@ -57,6 +57,12 @@ <h4 class="panel-title">
                         {% bootstrap_field variation_form.admission_valid_until layout="control" %}
                         {% bootstrap_field variation_form.admission_valid_from_offset_minutes layout="control" %}
                         {% bootstrap_field variation_form.admission_valid_until_offset_minutes layout="control" %}
+                    {% else %}
+                        <input type="hidden" name="{{ variation_form.admission_validity_mode.html_name }}" value="{{ variation_form.instance.admission_validity_mode|default_if_none:'inherit' }}">
+                        <input type="hidden" name="{{ variation_form.admission_valid_from.html_name }}" value="{{ variation_form.admission_valid_from.value|default_if_none:'' }}">
```
Aqil-Ahmad: The new hidden inputs for admission_valid_from / admission_valid_until don't work, and the failure mode is data loss

Aqil-Ahmad (changes requested):

Rachit7168 (changes requested):
When validation fails for the number (I enter a string). It shows a error message below the field as well "Enter a number." Change it to something meaningfull message and if validation fails for case in which I did not add product name it just shows plain border. Make it consistent and show message for every validation there is.

https://github.com/user-attachments/assets/2c0e5611-0cdd-4e17-948b-f5ef381b1fb1

jagritixjha:
https://github.com/user-attachments/assets/9b11ddb9-24a6-4ef0-9a92-48b0b1f17707

app/eventyay/control/templates/pretixcontrol/item/include_variations.html
```diff
@@ -21,7 +21,7 @@
                     <h4 class="panel-title">
                         <div class="row">
                             <div class="col-md-9">
-                                {% bootstrap_field variation_form.value layout='inline' form_group_class="" %}
+                                {% bootstrap_field variation_form.value layout='control' show_label=False %}
```
Rachit7168: Using layout='control' forces Eventyay's ControlFieldRenderer (which sets layout='horizontal'). This wraps the input inside an additional <div class="col-md-9">. Because this template already places this field inside a <div class="col-md-9"> (to leave room for the col-md-3 up/down/delete buttons), the double-nesting shrinks the input box to 75% of 75% of the panel width, leaving an awkward gap on the right. Using layout='vertical' gives you the error rendering you want without injecting nested grid columns.

app/eventyay/control/templates/pretixcontrol/item/include_variations.html
```diff
@@ -74,7 +74,7 @@ <h4 class="panel-title">
                     <h4 class="panel-title">
                         <div class="row">
                             <div class="col-md-9">
-                                {% bootstrap_field formset.empty_form.value layout='inline' form_group_class="" %}
+                                {% bootstrap_field formset.empty_form.value layout='control' show_label=False %}
```
Rachit7168: Same as above, you need to apply this to the empty_form template so new variations added dynamically don't suffer from the same squished layout.

Rachit7168 (changes requested):
Working is fine. Please address

Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/cab731ed-8671-423e-9f15-16edb2ca5da2

ArnavBallinCode (approved):
pr will be merged post https://github.com/fossasia/eventyay/pull/5844

please update migrations once https://github.com/fossasia/eventyay/pull/5844 is emrged

Aqil-Ahmad (approved):

Rachit7168:
@jagritixjha , Please update the migration file

PR #5784, opened 2026-09-20
docs: add -y flag to apt install dependency command

Sunjay-492:
Description
This pull request adds the `-y` (automatic yes) flag to the Ubuntu/Debian dependency installation command using `xargs`. 

Why is this change necessary?
Without the `-y` flag, running `xargs -a deb-packages.txt sudo apt install` prompts the user interactively in the terminal to confirm the installation (`Y/n`). 
* If a beginner runs this command without realizing it requires manual input, the terminal session can stall.
* If this setup documentation is used in automated installation scripts or continuous integration setups, the interactive prompt causes the build workflow to fail entirely.

Adding the `-y` flag ensures a completely smooth, non-interactive setup experience for developers setting up the `eventyay` local environment.

Track Tracker
Codeheat participant contribution

## Summary by Sourcery

Documentation:
- Update dependency installation instructions to run Debian/Ubuntu package installation non-interactively.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Debian/Ubuntu dependency installation command to proceed without confirmation prompts during local setup.


Saksham-Sirohi (changes requested):
Please explain the need for this. Any installation by default should require permission before moving forward

PR #5834, opened 2026-09-21
fix(oauth): disable OAuth apps instead of deleting them

### LINKED ISSUE #5833: "Disable application" deletes the OAuth app, or crashes with a 500
## Describe the bug

When a user tries to disable one of their own OAuth applications via **Account → OAuth applications → Disable**, one of two things happens:

1. **If the app has ever made an API call:** Clicking **Disable** returns an **HTTP 500 error** with `ProtectedError`. The app cannot be disabled.

2. **If the app has never been used:** The app is **hard-deleted** instead of disabled, along with all its access tokens and refresh tokens. A redirect back to the list makes it look like it worked.

The confirmation page and button both say "disable … permanently", but neither path actually disables the app (sets `active = False`).

## Expected behavior

Clicking **Disable** should:
- Keep the row in the database
- Set `active = False`
- Hide the app from the user's list (query filters `active=True`)
- Allow the row to remain because `LogEntry.oauth_application` has `on_delete=PROTECT`

The intent is clear from the code: `OAuthApplication.active` exists, `is_usable()` checks it, and every queryset filters `active=True`.

## Screenshots

**1. Your applications list:**

Image

**2. Disable confirmation (says "disable permanently"):**
Image

**3. After clicking Disable on an app that's been used — 500 ProtectedError:**
Image

**4. With the fix applied — app is disabled, hidden from list:**

Image

## Additional context

**Root cause:** Since Django 4.0, `DeleteView.post()` calls `form_valid()`, not `delete()`. The view's `delete()` override never runs:

```python
class OAuthApplicationDeleteView(ApplicationDelete):
    def delete(self, request, *args, **kwargs):
        self.object.active = False      # ← never reached
        self.object.save()
        return HttpResponseRedirect(self.success_url)
```

Django 6.0.7 code path:
```python
def form_valid(self, form):
    success_url = self.get_success_url()
    self.object.delete()                # ← hard delete, no `active=False`
    return HttpResponseRedirect(success_url)
```

**Same bug in five admin views:** `BBBServerDelete`, `JanusServerDelete`, `JitsiServerDelete`, `TurnServerDelete`, `LoungeMeshServerDelete` in `control/views/admin_views.py` all have unreachable `delete()` overrides that write audit log entries. The servers are still hard-deleted with no audit record.

**The fix:**
- Route `form_valid()` to the existing soft-disable in OAuth view
- Change the five admin views to use `CompatDeleteView` (already in the codebase for this purpose)

**Verified:** Confirmed on upstream `dev` (commit `062d72f`). Tests: `2 failed, 1 passed` before fix → `3 passed` after fix.

sanketkokate712:
Fixes #5833

## Description

On **Account → OAuth applications → Disable**, the app is supposed to be disabled (`active = False`). Instead:

- if the app has made an API call, the request fails with a **500 `ProtectedError`**, because `LogEntry.oauth_application` is `on_delete=PROTECT`
- otherwise the app is **hard-deleted**, along with its grants, access tokens and refresh tokens

The cause: since Django 4.0, `DeleteView.post()` calls `form_valid()`, not `delete()`. So the soft-disable in `OAuthApplicationDeleteView.delete()` never runs, and Django's default `form_valid()` calls `self.object.delete()`.

The five video server delete views in the admin (`BBBServerDelete`, `JanusServerDelete`, `JitsiServerDelete`, `TurnServerDelete`, `LoungeMeshServerDelete`) have the same problem. Their `delete()` overrides write the `*server.deleted` audit log entry, but those overrides never run either, so servers are deleted with no audit record.

## Changes

- `eventyay_common/views/account/oauth.py`: `OAuthApplicationDeleteView.form_valid()` now calls the existing soft-disable in `delete()`, which saves only `active`
- `control/views/admin_views.py`: the five server delete views now use `CompatDeleteView`, which is already in the codebase for this case (`control/views/pages.py` uses it too). Its `form_valid()` calls `delete()`, so the audit log entry is written again.

## Screenshots

**1. The user's application list**

Your applications

**2. The confirmation page asks to *disable* the app**

Disable confirmation

**3. Before: clicking Disable on an app that has been used returns a 500**

500 ProtectedError before the fix

**4. After: the app is disabled and no longer shown in the list. The row stays in the database with `active = False`.**

App disabled after the fix

## Tests

- `tests/eventyay_common/test_oauth_app_disable.py`: disabling keeps the row and sets `active = False`; disabling an app referenced by a `LogEntry` no longer errors; a disabled app is hidden from the list
- `tests/stable/test_video_server_delete.py`: for each of the five server types, deleting from the admin removes the server and writes the `*server.deleted` log entry

```
pytest tests/eventyay_common/test_oauth_app_disable.py tests/stable/test_video_server_delete.py
8 passed
```

## Summary by Sourcery

Fix OAuth application disabling and video server deletion so their intended soft-disable and audit logging behavior is applied consistently.

Bug Fixes:
- Preserve OAuth application records while disabling them, preventing protected-delete errors and ensuring disabled applications are hidden from users.
- Restore audit logging when administrators delete video servers across all five supported server types.

Tests:
- Add coverage for OAuth application disabling, protected log references, and list visibility.
- Add coverage for deletion and audit logging of all supported video server types.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deletion handling for video server configurations in the administration interface.
  - OAuth applications are now deactivated without removing their records, preserving related audit history.
  - Disabled OAuth applications no longer appear in users’ application lists.

- **Tests**
  - Added coverage for OAuth application deactivation, audit-log preservation, and list filtering.
  - Added coverage for deleting all supported video server types and recording the corresponding audit entries.


sanketkokate712:
fixed in [c78f8a6](https://github.com/fossasia/eventyay/pull/5834/commits/c78f8a62c9576a909522b5090a2837c749513b02)

sanketkokate712:
Fixed in[b5a12b8](https://github.com/fossasia/eventyay/pull/5834/commits/b5a12b8b7ae8c298e3cd8958512e957f22d2c633)

PR #5838, opened 2026-09-21
fix(dashboard): apply currency filtering before limiting top events

### LINKED ISSUE #5837: Fix Top events by ticket sales showing no sales for a currency outside the overall top ten
On the organizer dashboard, **Top events by ticket sales** reports "No ticket sales for this currency" for a currency that has sales, whenever that currency's events rank below the overall top ten.
The **Gross revenue** card lists and totals the same currency, so the two panels contradict each other.

The dashboard is reached at:

```text
/common/organizer/<organizer>/
```

## Current behavior

* An organizer has ten USD events with 3 paid orders each and one INR event with 1 paid order, all with confirmed payments.
* **Gross revenue** shows `₹1,500.00 · $1,200.00` and offers **INR** and **USD** in its currency dropdown.
* **Top events by ticket sales** lists only the ten USD events, and has no currency dropdown of its own.
* Selecting **INR** on the **Gross revenue** card changes the card to `₹1,500.00` and the URL to `?revenue_currency=INR`.
* **Top events by ticket sales** then hides every row and shows "No ticket sales for this currency."
* The INR event never appears in the table under any setting.

## Expected behavior

Selecting a currency shows that currency's top events, up to the same limit of ten rows.
**All currencies** keeps showing the overall top ten.
The table's own currency dropdown lists every currency that has ticket sales.

## Probable cause

`OrganizerAnalyticsView._get_tickets_data()` in `eventyay/eventyay_common/views/organizer_analytics.py` ranks all events together and cuts the result before any currency is known:

```text
.order_by('-total_orders')[:10]
```

`applyRevenueFilter()` in `organizer-dashboard-filters.js` filters on the client, so it can only hide or show rows that were already rendered.
An event outside the overall top ten never reaches the browser.

The table's currency options are built from those same ten rows in `get_context_data()`, while the card's options come from all confirmed payments in `organizer_dashboard.py`, which is how the card can offer a currency the table cannot show.

## Proposed solution

Rank the top ten per currency on the server, with a deterministic tie-breaker, and mark which of those rows form the overall top ten for the **All currencies** view.

## Screencast
https://github.com/user-attachments/assets/1f2aa68f-1725-482e-95df-109fe3ca65e2
## Required changes

* Return the top ten events per currency from `_get_tickets_data()` instead of the overall top ten, keeping the query bounded and the existing event permission filtering.
* Break ties in order counts deterministically.
* Show the overall top ten for **All currencies** and only the selected currency's rows otherwise.
* List every currency present in those rows in the table's currency dropdown.

## Acceptance criteria

* With ten USD events ranking above one INR event, selecting **INR** lists the INR event instead of "No ticket sales for this currency."
* **All currencies** lists the same overall top ten as before.
* No currency shows more than ten rows.
* Events with equal order counts always appear in the same order.
* The table's currency dropdown offers **INR** in that setup.
* An organizer with a single currency sees the table unchanged.

arpan7sarkar:
## What

**Top events by ticket sales** now shows the selected currency's top events, instead of "No ticket sales for this currency" when that currency ranks below the overall top ten.
**All currencies** still shows the overall top ten.

## Why

Fixes #5837.

`_get_tickets_data()` cut the ranking to the overall top ten before any currency was chosen, and the currency filter runs on the client, so it could only hide rows that were already rendered.

Measured on a local organizer with ten USD events (3 paid orders each) and one INR event (1 paid order):

```text
                          before                                  after
All currencies            10 USD rows, no table currency filter   same 10 USD rows, table filter offers INR and USD
INR (card or table)       0 rows, "No ticket sales ..."          the INR event
USD                       10 rows                                 10 rows
```

## How

- The top events query ranks with `ROW_NUMBER()` partitioned by currency and keeps at most `top_events_limit` (10) rows per currency, so the result stays bounded and still respects `can_view_orders`.
  Ties are broken by event id, where the old `-total_orders` order was not deterministic.
- Every event in the overall top ten is also in its own currency's top ten, so the overall top ten is the first rows of the merged ranking.
  `get_context_data()` flags them, and the other rows render `hidden` for **All currencies**.
  The flag is set after the cache read, so payloads cached before a deploy still render correctly.
- `syncSelects()` falls back to **All currencies** when the other dropdown has no option for the chosen currency.
  The table can now offer a currency with orders but no confirmed payments, which the card does not list, and this used to leave the card dropdown blank.

The server does not apply `?revenue_currency` to the rendered rows, and the script reads that value from the URL rather than from the validated dropdowns, so an unknown code such as `?revenue_currency=XYZ` still shows an empty table.
Both behave the same on `dev` and are left for a separate change.

## Before

https://github.com/user-attachments/assets/1f2aa68f-1725-482e-95df-109fe3ca65e2

## After
https://github.com/user-attachments/assets/08ad053a-9147-47c8-9e67-db16f6864036

## Testing

Two tests added to `app/tests/eventyay_common/test_organizer_analytics.py`.
Both fail on `dev` and pass here: one checks that an INR event outside the overall top ten is listed, that the overall top ten is unchanged with a deterministic order, and that the table offers both currencies; the other checks that each currency is capped at ten rows.

`pytest tests/eventyay_common/test_organizer_analytics.py`: 8 passed.
`pytest tests/eventyay_common/`: 7 failures in clone, private testmode, video team permissions and 2FA tests, present unchanged on `dev`.

Checked in the browser with JavaScript on and off, switching currencies from both the card and the table, reloading with `?revenue_currency=INR`, and on a single-currency organizer.

## Summary by Sourcery

Fix top-event currency filtering so each currency retains its own ranked results while All currencies continues to show the overall top ten.

Bug Fixes:
- Apply currency filtering to the full per-currency top-events ranking so currencies outside the overall top ten remain available for selection.

Enhancements:
- Preserve the overall top ten as the default view while retaining up to ten ranked events per currency, with deterministic ordering and fallback handling for unavailable currency selections.

Tests:
- Add analytics coverage for including currencies outside the overall top ten and enforcing per-currency result limits.


Rachit7168 (approved):
LGTM , Tested 

https://github.com/user-attachments/assets/c0a1b3a7-5658-4d02-9b58-62595721c5f9

ArnavBallinCode (approved):

arpan7sarkar:
Thanks @Rachit7168 :)

PR #5853, opened 2026-09-22
fix(quickstart): honor unchecked "Ticket downloads" on Save and continue

### LINKED ISSUE #5852: Quickstart "Ticket downloads" checkbox is ignored on "Save and continue"
**Describe the bug**
In the Ticketing Quickstart wizard, unchecking Ticket downloads (step 4) and clicking Save and continue does not save the change. The `ticket_download` event setting remains enabled, because the save handler only writes the True case.

Steps to reproduce
1. Create a fresh event and open `/control/event/<organizer>/<event>/quickstart/`.
2. In step 4 ("Ticket features"), uncheck "Ticket downloads".
3. Complete the required steps and click Save and continue.
4. Go to the event's Tickets tab → "Allow users to download tickets" is still enabled.

**Expected behavior**
Unchecking the box should disable `ticket_download`, the same way Save draft already does.

**Screenshots**
Screencast of the bug

https://github.com/user-attachments/assets/d4948962-667f-4904-9b12-ea72c2edfae3

Screencast when clicking Save draft

https://github.com/user-attachments/assets/086cb05f-2e1f-41c3-aeba-7ebd1951c910

**Additional context**
Add an else branch writing `self.request.event.settings.ticket_download = False` in the continue path, leaving `ticketoutput_pdf__enabled` untouched.

munazzaghazali7-hash:
## Summary
Fixes #5852 
Unchecking Ticket downloads in the Ticketing Quickstart and clicking Save and continue silently kept the setting enabled: the save path only wrote the `True` case, so `ticket_download` always fell back to its `True `default (`base/configurations/default_setting.py`). Save draft already handled both states — this makes Save and continue consistent with it.

## Root cause
`QuickSetupView.form_valid` (`control/views/event.py:1728`) wrote the setting only inside if `form.cleaned_data['ticket_download']:`, with no else. Unchecked checkboxes submit nothing → `cleaned_data['ticket_download']` is False → branch skipped → nothing written.

## Reproduction (verified locally)
Fresh event, quickstart step 4, baseline: settings.get(`'ticket_download'`) → True, `Event_SettingsStore row value='True'`
Uncheck Ticket downloads → Save and continue → success message
Re-check: still True, row still 'True' — the "off" was never written

## Fix
Add the else branch writing `ticket_download = False` after the existing if block. `ticketoutput_pdf__enabled` is intentionally left enabled-on-only: it's the PDF output format, defaults to True, and is meaningless while downloads are off.

## Screencast

https://github.com/user-attachments/assets/6ad9c9c4-03af-4f7c-8211-9203c7d4874f

## Testing
- [x] Uncheck + Save and continue → settings.get('ticket_download') is False, stored row 'False'
- [x] Check + Save and continue → still True, passbook plugin still enabled when available
- [x] Save draft path unchanged (both states, as before)

## Summary by Sourcery

Bug Fixes:
- Ensure the Ticketing Quickstart's Save and continue action persists Ticket downloads as disabled when the option is unchecked.

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabling ticket downloads during event setup now correctly prevents attendees from downloading tickets.
  * Enabling ticket downloads continues to support the existing PDF and passbook ticket options.


Rachit7168 (approved):
LGTM, Tested 

https://github.com/user-attachments/assets/e70ab0c3-2c7c-4e95-adfb-31dfa1a0afef

